PR #10, opened 2025-05-23
Event overview page scaffold

trevnels-cs:
- Created a Container component to constrain main content width consistently.

- Scaffolded out event overview page with tabs:
image

Tabs are kept in an object, so we can use keys to reference them (i.e. in query parameter) - the tabs component only allows the use of an index, which won't be reliable if we're showing/hiding tabs conditionally.

- Adjusted our stylesheet to match the Tailwind palette used by flowbite.

- Added `react-icons`, which is what flowbite-react uses for icons. FA icon components are imported and passed as props to flowbite components. Still trying to figure out how to set aria-hidden on these - will probably need an HoC wrapper.


trevnels-cs:
image

Updated to use radix and decoupled the tab bar width from tab content. Still working on updating the navigation logic but it's nearly ready.

frontend/src/index.css
```diff
@@ -1,25 +1,16 @@
-@import 'tailwindcss';
+@import url('https://fonts.googleapis.com/css2?family=Inter:ital,opsz,wght@0,14..32,100..900;1,14..32,100..900&display=swap');
```
cjones-cs: We shouldn't access an api for a font that's only a placeholder. Using the system font should be fine until we have a styling guide from the graphic designer

frontend/src/index.css
```diff
@@ -1,25 +1,16 @@
...
   -moz-osx-font-smoothing: grayscale;
 }
 
-@media (prefers-color-scheme: light) {
-  :root {
-    color: #213547;
-    background-color: #ffffff;
-  }
+.radix-themes {
+  --default-font-family: 'Inter', sans-serif;
```
cjones-cs: love this

frontend/src/main.tsx
```diff
@@ -11,7 +11,12 @@ import './index.css';
 createRoot(document.getElementById('root')!).render(
   <StrictMode>
     <BrowserRouter>
-      <Theme>
+      <Theme
+        appearance="dark" // TODO: use system preference & allow user to toggle
+        panelBackground="solid" // disable blur effect on surfaces for performance
```
cjones-cs, line 16: Can you explain the issue with blue in more detail?

frontend/src/routes/events/Overview.tsx
```diff
@@ -1,3 +1,103 @@
...
+            </Tabs.Trigger>
+            <Tabs.Trigger value="team">
+              <FaUserGroup className="mr-1" />
+              Team
+            </Tabs.Trigger>
+          </Tabs.List>
+        </Container>
+
+        <Tabs.Content value="challenges">
+          { /* TODO: extract to a separate component */ }
```
cjones-cs: I would go ahead and make these into separate components so you have the placeholder tabs to fill out later

frontend/src/main.tsx
trevnels-cs: I assume you meant blur here - by default radix applies a translucency/blur effect to cards and other surfaces, which seems to cause some lag in my browser. I assume this will be even more noticeable on slower machines.

Since we're only putting cards on top of solid colors, the blur effect isn't noticeable anyway, so I think it makes sense to turn off.

cjones-cs: lol, yeah, I meant blur. Okay cool

cjones-cs (approved):

PR #13, opened 2025-06-03
build workflow to s3 for dev

zoecough:


.github/workflows/build.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Setup AWS CLI
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+
+      - name: Sync dist to S3 bucket
+        run: |
+          aws s3 sync ./frontend/dist s3://assets.development.ctf-ng.com --delete
```
fpayer, line 39: Does this work as the domain name?

zoecough: it's the bucket name, i don't know the planned domain name so i just made a guess since it'll change once we move to CISA/OTA resources because the buckets will need to be remade

fpayer: Ah, so it's just a name for within S3. Not Internet-routable, right?

zoecough: yup

fpayer (approved):

PR #17, opened 2025-06-05
Build for other branches

zoecough:
info on semver-release action here:
https://github.com/marketplace/actions/semver-release

once this change is merged into staging, further merges to staging will require a label (major, minor, or patch) to increment the version number


zoecough:
(s3 buckets were remade with these new names in the AWS web console as well)

fpayer (approved):

PR #22, opened 2025-06-07
Initial implementation of backend models, API endpoints, and controllers. (MINOR)

### LINKED ISSUE #21: Initial implementation of backend models, API endpoints, and controllers.
Foundational backend models, controllers, and API.

CarterPerez-dev:
### This PR implements the initial backend system for the multi world and team features. It includes the models, controllers, API endpoints,  tests, and some extra documentation for the new backend developer. (MINOR)

**Resolves: #21**

#### Adds a team management system with "Worlds" (events) where users can join different teams across multiple worlds using invite codes. The backend uses  SQLAlchemy models (World, Team, User, TeamMember), business logic controllers, and REST API endpoints. The /plugin/tests/ folder includes unit tests for config, integration tests for models/controllers/database constraints, API endpoint tests

***Added a .python-version file to fix the build failures I was seeing with newer, incompatible Python versions. (3.13)***
***Added .pre-commit-config.yaml for basic python linting/formatting before commiting. (now ruff)***

Edit: Fixed ruff linting and did some DRY refactoirng while I was at it.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,20 @@
+# .pre-commit-config.yaml
```
RayofLightz: Add this to gitignore

backend/ctfd/plugin/controllers/database_controller.py
```diff
@@ -0,0 +1,260 @@
+# /plugin/controllers/database_controller.py
```
RayofLightz: Heads up we are not piling all of the controllers into one folder. I believe we had established making a dir for the domain ex `team` and then having models, controllers, and route directories for as sub directories of it.

backend/ctfd/plugin/tests/api/test_api.py
```diff
@@ -0,0 +1,183 @@
+"""
+/plugin/tests/test_api.py
+API Tests
+"""
+
+import pytest
+from plugin.models import Team, TeamMember
+from plugin.models.enums import TeamRole
+
+pytestmark = pytest.mark.db
```
RayofLightz: Is there a reason you have this file committed 3 times?

RayofLightz (commented):
Not the main reviewer but just something I noticed looking at the code.

backend/ctfd/plugin/controllers/database_controller.py
```diff
@@ -0,0 +1,260 @@
...
+                members_in_world = TeamMember.query.filter_by(world_id=world.id).count()
+
+                world_stats.append(
+                    {
+                        "id": world.id,
+                        "name": world.name,
+                        "teams": teams_in_world,
+                        "total_members": members_in_world,
+                    }
+                )
```
RayofLightz: This should be a union query done in the db

backend/ctfd/plugin/controllers/database_controller.py
```diff
@@ -0,0 +1,260 @@
...
+
+            all_teams = Team.query.all()
+            empty_teams = []
+            for team in all_teams:
+                if team.member_count == 0:
+                    empty_teams.append({"id": team.id, "name": team.name, "world_id": team.world_id})
+
```
RayofLightz: If possible just run a query for member_count == 0 and get the count from that instead of pulling and iterating over every team.

backend/ctfd/plugin/controllers/database_controller.py
RayofLightz: ^Join not union

backend/ctfd/plugin/controllers/database_controller.py
```diff
@@ -0,0 +1,260 @@
...
+    @staticmethod
+    def reset_all_plugin_data() -> dict[str, Any]:
+        """Deletes all plugin data from the database.
+
+        Returns:
+            dict: Success status and deletion counts or error info.
+        """
+        try:
+            from ..models.TeamMember import TeamMember
```
fpayer: Is there a reason why this import is being done within the function and not at the top of the file?

backend/ctfd/plugin/controllers/user_controller.py
```diff
@@ -0,0 +1,255 @@
...
+                    extra={"context": {"user_id": user_id, "world_id": world_id}},
+                )
+                return {
+                    "success": False,
+                    "error": f"World with ID {world_id} does not exist",
+                }
+
+            membership = TeamMember.query.filter_by(user_id=user_id, world_id=world_id).first()
+
+            if membership:
```
fpayer: Use a early-returns to reduce code complexity. i.e.

Replace:
```
if membership:
  ...
else:
  ...
```

With:
```
if not membership:
  return ...
...
```

This will remove 1 level of indentation from the "main" execution flow.

backend/ctfd/plugin/controllers/user_controller.py
```diff
@@ -0,0 +1,255 @@
...
+            user_id (int): The user ID.
+            world_id (int): The world ID to check eligibility for.
+
+        Returns:
+            dict: Success status, eligibility boolean, and reason if not eligible.
+        """
+        try:
+            from ..models.TeamMember import TeamMember
+
+            existing_membership = TeamMember.query.filter_by(user_id=user_id, world_id=world_id).first()
```
fpayer: Also cannot join a team if the team has already reached the max number of users

fpayer (changes requested):
General feedback:

1. 
Instead of wrapping all of the controllers in the `try/except IntegrityError`, have the route logic generically wrap all controllers with a try/except (specifically for the IntegrityError)

This will ensure consistent behavior and reduce complexity in the controllers

2. 
Imports should ideally happen at the top of the file and not within individual functions

backend/ctfd/plugin/controllers/database_controller.py
CarterPerez-dev: Ok that makes sense, I'll restructure, thanks!

backend/ctfd/plugin/controllers/database_controller.py
CarterPerez-dev: I believe that was a circular dependency issue. The domain restructuring will help minimize these. I'll move imports to the top where possible and clearly document any that must remain lazy imports.

backend/ctfd/plugin/controllers/user_controller.py
CarterPerez-dev: Agreed, that's much cleaner.

backend/ctfd/plugin/admin/routes/admin.py
```diff
@@ -0,0 +1,369 @@
...
+                    }
+                },
+            )
+            return error_response("Unable to fetch system statistics", "health", 500)
+
+        health_report = {
+            "status": "healthy",
+            "timestamp": datetime.utcnow().isoformat(),
+            "data_counts": counts,
+            "worlds_count": len(detailed.get("worlds", [])),
```
cjones-cs: Count from db model accessor function

backend/ctfd/plugin/admin/controllers/database_controller.py
```diff
@@ -0,0 +1,202 @@
...
+from ...world.models.World import World
+from ...team.models.Team import Team
+from ...team.models.TeamMember import TeamMember
+from ...user.models.User import User
+from sqlalchemy import func
+
+logger = get_logger(__name__)
+
+
+class DatabaseController:
```
RayofLightz: No need to have as a class. Just import the functions directly.

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+                "Team creation failed - world not found",
+                extra={"context": {"world_id": world_id, "team_name": name}},
+            )
+            return {
+                "success": False,
+                "error": f"World with ID {world_id} does not exist",
+            }
+
+        if limit is None:
+            limit = world.default_team_size
```
cjones-cs: If event.teamSize == 1, then it's an individual game. That also means if there is a Team record, then it only has a single player (the captain)

cjones-cs: use max team size and remove default team size

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+
+        existing_team = Team.query.filter_by(name=name, world_id=world_id).first()
```
cjones-cs: is name coming from the team? We may want to use the team id instead of name. Name isn't as stable as id would be. Or did I misunderstand this and this is checking to make sure names aren't dups?

backend/ctfd/plugin/admin/controllers/database_controller.py
```diff
@@ -0,0 +1,202 @@
...
+        world_stats = []
+        for world_id, name, teams, total_members in world_stats_query:
+            world_stats.append(
+                {
+                    "id": world_id,
+                    "name": name,
+                    "teams": teams,
+                    "total_members": total_members,
+                }
+            )
```
RayofLightz: You should be able to get these with a `select` and not need to itter the results.

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+
+        return {
+            "success": True,
+            "team": team,
+            "invite_code": invite_code,
+            "message": f"Team '{name}' created successfully in {world.name}",
+        }
+
+    @staticmethod
+    def join_team(user_id: int, team_id: int, world_id: int) -> dict[str, Any]:
```
cjones-cs: user id and invitecode id. Can get the team and event ids from the invite code

cjones-cs: Make sure invite code is valid. They are forcefully invalidated when a captain forcefully removes a player from their team.

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+        if existing_membership:
+            existing_team = Team.query.get(existing_membership.team_id)
+            logger.warning(
+                "Team join failed - user already in team",
+                extra={
+                    "context": {
+                        "user_id": user_id,
+                        "world_id": world_id,
+                        "existing_team_id": existing_membership.team_id,
+                        "existing_team_name": existing_team.name if existing_team else "Unknown",
```
cjones-cs: don't join on names, primarily on ids only. idt you need world id either bc you have the team id

backend/ctfd/plugin/utils/decorators.py
```diff
@@ -0,0 +1,65 @@
...
+from .api_responses import error_response
+from .logger import get_logger
+
+
+def authed_user_required(f):
+    """Decorator that ensures user is authenticated and attaches user to g.user"""
+
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        current_user = get_current_user()
```
trevnels-cs, line 19: Not sure if `ctfd.utils.user` will be usable for this if we're using our own user system - may need to set up something temporary to mock a user on authed endpoints until we have our own OAuth set up.

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+
+        return {
+            "success": True,
+            "message": f"User successfully joined team '{team.name}'",
+            "team": team,
+            "membership": membership,
+        }
+
+    @staticmethod
+    def leave_team(user_id: int, world_id: int) -> dict[str, Any]:
```
cjones-cs: Is there a function that allows them to transfer their captainship to another player. A player can't leave the team if they are the captain. They must select a new player as captain before leaving.

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+        teams = Team.query.filter_by(world_id=world_id).all()
+
+        team_list = []
+        for team in teams:
+            team_list.append(
+                {
+                    "id": team.id,
+                    "name": team.name,
+                    "member_count": team.member_count,
+                    "limit": team.limit,
```
cjones-cs: limit is max_size? maybe change it for clarity?

backend/ctfd/plugin/team/controllers/team_controller.py
cjones-cs: I see there's a join team by invite code below, so this function may be duplicative

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+from ..models.Team import Team
+from ..models.TeamMember import TeamMember
+from ..models.enums import TeamRole
+
+logger = get_logger(__name__)
+
+
+class TeamController:
+    @staticmethod
+    def create_team(
```
cjones-cs: need to check if world.locked, then can't create teams or add members. If team.locked, then can't add/remove members if a regular player/captain. Can remove if admin.

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+        """
+        team = Team.query.get(team_id)
+        if not team:
+            logger.warning(
+                "Team disband failed - team not found",
+                extra={"context": {"team_id": team_id, "actor_id": actor_id}},
+            )
+            return {"success": False, "error": "Team not found."}
+
+        is_captain = TeamMember.query.filter_by(team_id=team_id, user_id=actor_id, role=TeamRole.CAPTAIN).first()
```
cjones-cs: If there's only one captain per team, do we need .first()?

fpayer: The query returns an array. .first makes sure to return a single object instead

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+                "success": False,
+                "error": "Captains cannot remove themselves. Use the 'Leave Team' or 'Disband Team' feature.",
+            }
+
+        db.session.delete(membership_to_remove)
+        db.session.commit()
+        return {"success": True, "message": "Team member removed successfully."}
+
+    @staticmethod
+    def _generate_invite_code(length: int = config.INVITE_CODE_LENGTH) -> str:
```
cjones-cs: When removing a member, the invite code should change as well. This prevents malicious people from re-adding themselves if they have the old code. All codes should have some kind of valid/not valid flag to check as well.

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+                    "new_captain_id": new_captain_id,
+                    "actor_id": actor_id,
+                    "is_admin": is_admin,
+                }
+            },
+        )
+
+        return {
+            "success": True,
+            "message": f"User {new_captain_id} is now captain of '{team.name}'",
```
cjones-cs: This message combines ids and names, but if it's intended to be user readable, should just use names

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+            "team_id": team.id,
+            "captain_id": new_captain_id,
+        }
+
+    @staticmethod
+    def remove_captain(team_id: int) -> dict[str, Any]:
```
cjones-cs: I think this function wont be used. All teams must have a captain at all times otherwise frontend functionality would be borked

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+        db.session.commit()
+
+        return {
+            "success": True,
+            "message": f"Captain removed from '{team.name}'",
+            "team": team,
+        }
+
+    @staticmethod
+    def get_team_captain(team_id: int) -> dict[str, Any]:
```
cjones-cs: Not sure if we would need this function either, but if we did, then we may need the captains name as well.

backend/ctfd/plugin/team/controllers/team_controller.py
```diff
@@ -0,0 +1,785 @@
...
+            "limit": team.limit,
+            "is_full": team.is_full,
+            "invite_code": team.invite_code,
+            "ranked": team.ranked,
+        }
+
+        return {
+            "success": True,
+            "team": team_data,
+            "member_ids": member_ids,
```
cjones-cs: Since memberships below would have member_ids, we don't need an extra field for this data

backend/ctfd/plugin/team/models/Team.py
```diff
@@ -0,0 +1,40 @@
...
+from sqlalchemy import select, func
+from ... import config
+
+
+class Team(db.Model):
+    __tablename__ = "ng_teams"
+
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(config.TEAM_NAME_MAX_LENGTH), nullable=False)
+    limit = db.Column(db.Integer, nullable=False, default=config.DEFAULT_TEAM_SIZE)
```
cjones-cs: same as above. Change limit to "Max size" or something more descriptive

backend/ctfd/plugin/user/controllers/user_controller.py
```diff
@@ -0,0 +1,214 @@
...
+            member_counts = dict(
+                db.session.query(TeamMember.team_id, func.count(TeamMember.id))
+                .filter(TeamMember.team_id.in_(team_ids))
+                .group_by(TeamMember.team_id)
+                .all()
+            )
+
+            for team in teams_data:
+                team["team_member_count"] = member_counts.get(team["team_id"], 0)
+
```
cjones-cs: I'm a little confused by this section. It looks like you are doing 2 queries, one to get the team/world info for the singular player, then another to count the number of players on each of the teams the player is involved in. Is it possible to combine the groupBy into the top query so it's just a singular query?

As far as converting it from the query result to a dict, if we have to do this often, it may be worth writing a little one liner utility. A quick google search brings up a few ideas on this that could be explored.

cjones-cs: Does team.memberCount track to this same info? I just saw that on the model

backend/ctfd/plugin/world/controllers/world_controller.py
```diff
@@ -0,0 +1,253 @@
...
+        for world_id, name, description, team_count, total_members in world_stats:
+            worlds_data.append(
+                {
+                    "id": world_id,
+                    "name": name,
+                    "description": description,
+                    "team_count": team_count,
+                    "total_members": total_members,
+                }
+            )
```
cjones-cs: As mentioned above, a utility to easily convert row data to [] would be useful

backend/ctfd/plugin/world/controllers/world_controller.py
```diff
@@ -0,0 +1,253 @@
...
+                    "team_count": len(teams),
+                    "total_members": total_members,
+                }
+            },
+        )
+
+        return {"success": True, "world": world_data, "teams": teams_data}
+
+    @staticmethod
+    def update_world(world_id: int, name: str | None = None, description: str | None = None) -> dict[str, Any]:
```
cjones-cs: Do we need to include other information about the world?

backend/ctfd/plugin/world/routes/worlds.py
```diff
@@ -0,0 +1,293 @@
...
+from ...utils.decorators import json_body_required, handle_integrity_error
+from ...utils.logger import get_logger
+from ...utils import get_current_user_id
+from ...utils.validators import validate_world_creation, validate_world_update
+
+worlds_namespace = Namespace("worlds", description="world management operations")
+logger = get_logger(__name__)
+
+
+@worlds_namespace.route("")
```
cjones-cs: Is this supposed to be empty string?

backend/ctfd/plugin/team/controllers/team_controller.py
CarterPerez-dev: ok that makes sense, thanks.

backend/ctfd/plugin/admin/controllers/database_controller.py
CarterPerez-dev: Good call. I can definitely make that more efficient.

backend/ctfd/plugin/utils/decorators.py
CarterPerez-dev: Maybe a better approach is to create our own internal helper that would just be a simple wrapper around CTFd.utils.user.get_current_user(). This would give us a single, clean 'seam' , so when we later introduce our own OAuth, we only have to change the implementation of that one function, and the rest of the code doesn't need to be touched. What do you think?

backend/ctfd/plugin/team/controllers/team_controller.py
CarterPerez-dev: Ok thanks for clarifying this, I'll  update leave_team to explicitly check the user's role and block the action if they are the captain, prompting them to the transfer instead.

backend/ctfd/plugin/team/controllers/team_controller.py
CarterPerez-dev: Good point, I'll update limit and look for any other possible ambiguous names to fix as well.

backend/ctfd/plugin/team/controllers/team_controller.py
CarterPerez-dev: I'll add the locked fields, and quick question: when a team is locked, should we also block name changes and captain transfers, or just adding/removing members?

backend/ctfd/plugin/team/controllers/team_controller.py
CarterPerez-dev: True, I'll do a quick sanity check on the rest of the code to make sure there aren't any other obvious loopholes like that. as well. Thanks!

backend/ctfd/plugin/team/controllers/team_controller.py
CarterPerez-dev: Yea true, wouldn't really make sense seeing the ID from a user perspective, good catch, thanks.

backend/ctfd/plugin/user/controllers/user_controller.py
CarterPerez-dev: I wrote this part of the controller before I implemented the member_count hybrid property on the Team model, and I clearly forgot to go back and refactor this to use it, thanks!

backend/ctfd/plugin/world/controllers/world_controller.py
CarterPerez-dev: For now, my intention was to keep this endpoint focused on the most basic mutable fields (name and description). But good question, are you thinking we should be able to update fields like size and locked here as well?

backend/ctfd/plugin/world/routes/worlds.py
CarterPerez-dev: Yea, that's the right syntax, It maps to /worlds.

zoecough:
when these changes get merged into development, be sure to add an appropriate semver label (`(MAJOR)`, `(MINOR)`, or `(PATCH)`) to the squashed commit generated by the squash and merge operation, it'll get picked up when this eventually goes to staging

backend/ctfd/plugin/team/controllers/_generate_invite_code.py
```diff
@@ -0,0 +1,33 @@
...
+import secrets
+import string
+import uuid
+
+from ... import config
+from ...config import INVITE_CODE_GENERATION_ATTEMPTS
+from ..models.Team import Team
+
+
+def _generate_invite_code(length: int = config.INVITE_CODE_LENGTH) -> str:
```
fpayer, line 16: Is there a reason why this function is prefixed with `_`?

backend/ctfd/plugin/team/models/Team.py
```diff
@@ -0,0 +1,88 @@
...
+    def __repr__(self):
+        return f"<Team {self.name}>"
+
+    # Avoids a separate query for every team's member count
+    @hybrid_property
+    def member_count(self):
+        return len(self.members)
+
+    @member_count.expression
+    def member_count(cls):
```
fpayer, line 36: Avoid using duplicate function name `member_count`

fpayer (commented):
I have some additional suggestions for more code organization, but I think that would be best for a followup PR.

backend/ctfd/plugin/team/controllers/_generate_invite_code.py
CarterPerez-dev: Yea, that was intentional, my thinking was to treat it as an 'internal' helper for the module, since it's only ever called by other functions here and isn't part of the API. That said, glad you pointed that out it because I see I also put it in the __all__, which kinda defeats the purpose. I'll fix that and add a comment.

backend/ctfd/plugin/team/models/Team.py
CarterPerez-dev: It's one of those weird Alchemy things. You have to name it the same thing twice or it breaks. I'll add a comment there so it doesn't confuse anyone in the future.

backend/ctfd/plugin/team/controllers/_generate_invite_code.py
CarterPerez-dev: *`__all__`

fpayer (approved):

backend/ctfd/plugin/team/models/Team.py
```diff
@@ -0,0 +1,89 @@
...
+    invite_code = db.Column(db.String(config.INVITE_CODE_MAX_LENGTH), nullable=False, unique=True)
+    event_id = db.Column(db.Integer, db.ForeignKey("ng_events.id"), nullable=False, index=True)
+    locked = db.Column(db.Boolean, default=False, nullable=False)
+
+    __table_args__ = (db.Index("ix_ng_teams_event_name", "event_id", "name"),)
+
+    members = db.relationship("TeamMember", back_populates="team", cascade="all, delete-orphan")
+
+    def __repr__(self):
+        return f"<Team {self.name}>"
```
RayofLightz, line 27: Is there a reason for overriding the string to this? I would think we would want the id as well since that is the primary key (Apply this advice to the other models)

backend/ctfd/plugin/utils/logger.py
```diff
@@ -0,0 +1,41 @@
+"""
+/backend/ctfd/plugin/utils/logger.py
+Configures a JSON logger for machine readable app logging.
```
RayofLightz, line 3: In the future we will want this to be disabled on prod as its currently very noisy.

backend/ctfd/plugin/user/controllers/get_user_teams_in_event.py
```diff
@@ -0,0 +1,82 @@
...
+        logger.warning(
+            "Get user teams in event failed - event not found",
+            extra={"context": {"user_id": user_id, "event_id": event_id}},
+        )
+        return {
+            "success": False,
+            "error": f"Event with ID {event_id} does not exist",
+        }
+
+    team_member = TeamMember.query.filter_by(user_id=user_id, event_id=event_id).first()
```
RayofLightz, line 38: We might want to compound index this.

RayofLightz (commented):
Mostly some small things. I don't think they need addressed now but are worth thinking about

backend/ctfd/plugin/team/controllers/team_controller.py
cjones-cs: Yes, No changes to teams can be made by regular/captain players at all, including the name

backend/ctfd/plugin/world/controllers/world_controller.py
cjones-cs: Yes, If I'm looking for a function to update the world, I would expect to be able to update any field listed on the model.

PR #26, opened 2025-06-09
Build and semver changes

zoecough:


zoecough:
need new PR so the version number doesn't get incremented by this

PR #29, opened 2025-06-10
Fixed some missed renames from placeholder -> ctf-ng

fpayer:


zoecough (approved):
looks good to me, i tried a git grep for placeholder and the only other instances i saw looked unrelated to the old project name

PR #30, opened 2025-06-12
Create Docker Compose Challenge Parser

johnmarsden-edu:
## Overview
- Defined Python types for all chosen supported features from Docker Compose and for all chosen supported features from the challenge specification
- Implemented parser from yaml to ComposeFile that includes ChallengeInfo
- Added tests for the parser that test that it correctly parses the chosen format and provides examples of how to write the compose challenge files
- Created CLI for local validation

## Details
- Implemented a yaml stream rewriter that moves the anchor from the default value to the template
- Utilized cattrs and attrs to implement a structuring process that follows from the yaml parsing to allow for users to work with the data in a more convenient structured typed format that will be amenable to intellisense. 
- Defined specialized wrapper type Template that can be used by the team to detect which environment variables are templated and thus generated and which are straight strings

## Required Work
- [x] Build out CLI for local validation
- [x] Build out full example challenge compose files that can be used both as examples and test inputs
- [x] Clean up error reporting for ease of use with the parser such that you can easily understand what the errors are with your file.
- [x] Add further documentation to code to express the motivations behind certain choices


tools/chall-parser/parser/challenge_info.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+# # This is the container block for all of the information that you can specify
+# # for use in challenge development
+# x-challenge:
+# 	# Name of challenge
+# 	name: Text
+# 	# The description presented to players
+# 	description: Text
+# 	# The name of a https://fontawesome.com/icons representative of the challenge
```
trevnels-cs: On the frontend side we are moving towards Tabler icons instead of fontawesome, so this comment should be adjusted to reflect that. The new link is https://react-icons.github.io/react-icons/icons/tb/.

tools/chall-parser/README.md
```diff
@@ -0,0 +1,388 @@
+# Challenge Parser
+
+A Docker Compose parser with CTF challenge extensions that supports templated variables and challenge metadata.
+
+## Usage
+
+Collecting workspace informationBased on the CLI code, here's a comprehensive usage explanation for the README:
```
fpayer: Nit - `informationBased` -> `information based`

tools/chall-parser/main.py
fpayer, line 1: Unnecessary file?

fpayer (commented):
For organizational purposes, I'd recommend keeping the py file that define the structures in a separate folder from the actual parsers: eg. `parser/compose.py`

tools/chall-parser/parser/cli.py
```diff
@@ -0,0 +1,425 @@
...
+        
+        console.print(f"[bold green]✓[/bold green] Successfully validated: {file_path}")
+        console.print()
+        
+        if show_summary:
+            if output_format == "table":
+                display_challenge_summary(compose.challenge)
+                display_services_summary(compose)
+            elif output_format == "json":
+                import json
```
RayofLightz: Import json at the top level?

tools/chall-parser/parser/cli.py
```diff
@@ -0,0 +1,425 @@
...
+    """
+    Quick validation check (exit code only).
+    
+    Performs validation and returns appropriate exit codes:
+    - 0: Valid
+    - 1: Invalid/Error
+    """
+    try:
+        if stdin:
+            import sys
```
RayofLightz: Sys is already imported at the top level

backend/ctfd/plugin/requirements.txt
```diff
@@ -3,4 +3,4 @@ flask-restx==1.1.0
 pre-commit==4.2.0
 
 # You'll need to pip install from the root directory of the repository for this to work
-./lib/parser
\ No newline at end of file
+../lib/parser
```
RayofLightz: Have you verified that this works when you start up ctfd with docker compose. Even if it is just requiring the parser in the init and not committing it?

RayofLightz (commented):
Just one question. We want to merge this into dev in the next day or two so I can get started working on the initial stuff for actually importing the yaml into a db object.

backend/ctfd/plugin/requirements.txt
johnmarsden-edu: I have started digging into this and no, it does not work. I'm looking at other potential solutions as it currently exists as a package and I'm not sure how nicely that will play with symlinking it directly into the plugin folder. There's also not currently a great spot to put it without adding another volume to the container which I'm not sure where that would belong.

PR #33, opened 2025-06-13
Pytest runner

zoecough:


zoecough:
restarting, i shouldnt need to modify the requirements or makefile since this is all just for the gh runner. barking up the wrong tree

PR #34, opened 2025-06-13
added installs to run before make

zoecough:


zoecough:
should they be split up? it looks like Carter put together options in the makefile for unit/integration/api tests as well as domain-specific tests but i think the tests under tests/fixtures get left out in either of those configurations BUT it would likely make the test results more readable

fpayer:
I'd double check with Carter - we are doing a slight re-org with the folder structure so that test cases will be located with their relevant domain rather than being in a standalone "tests" folder. So maybe it'll get addressed with that change.

fpayer (approved):

zoecough:
will merge for now and update based on the re-org changes to split things up for readability

PR #35, opened 2025-06-13
Dashboard UI scaffolding

trevnels-cs:
Looks like this:
image

For now, colored boxes are placeholders for card graphics. The specific text, icons, and data presented are also subject to change.


frontend/src/components/dashboard/PastEvents.tsx
```diff
@@ -0,0 +1,38 @@
+import { Grid, Heading } from '@radix-ui/themes';
+import type { accentColors } from '@radix-ui/themes/props';
+import EventCard from 'components/event/EventCard';
```
cjones-cs: Move this to routes/dashboard

frontend/src/components/dashboard/UpcomingEvents.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <>
+        {heading}
+        <Callout.Root color="jade" variant="surface">
+          <Callout.Icon>
+            <TbInfoCircle />
+          </Callout.Icon>
+          <Callout.Text>
+            You are not registered for any upcoming events.
+            {' '}
+            <RadixLink asChild><Link to="./events">Register for one</Link></RadixLink>
```
cjones-cs: We should use more explicit language. "Register for an upcoming event"

frontend/src/components/event/EventCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+  color = 'lime',
+}: {
+  id: string;
+  name: string;
+  description: string;
+  color?: typeof accentColors[number];
+}) {
+  return (
+    <Card asChild>
+      <Link to={`/${ROUTEPREFIX}/events/${id}`}>
```
cjones-cs: You updated to remove ROUTEPREFIX on your other PR, so make sure to fix this

frontend/src/components/event/EventCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+              {/* Placeholder for event card graphic */}
+              <Box className="h-full w-full" style={{ backgroundColor: `var(--${color}-8)` }} />
+            </AspectRatio>
+          </Inset>
+          <Flex direction="column" gap="2" className="flex-grow" justify="between">
+            <Box>
+              <Heading size="4">{name}</Heading>
+              <Text size="2" color="gray">{description}</Text>
+            </Box>
+            <Box>
```
cjones-cs: I don't think we should include divs that aren't going to have explicit rendered content. I would either remove it or specify content so it doesn't end up being dead code we missed removing.

frontend/src/components/HeaderContainer.tsx
```diff
@@ -1,19 +1,21 @@
 import { Container } from '@radix-ui/themes';
 
-export default function HeaderContainer({ children }: {
-  children: React.ReactNode;
+export default function HeaderContainer({ children = undefined }: {
+  children?: React.ReactNode
```
cjones-cs, line 4: Should this be optional? Seems like if it were undefined, you wouldn't use this at all

trevnels-cs: Since the dashboard (and potentially other routes) will populate the children of this component dynamically (which may be an empty list), I'm allowing this to be optional so that empty lists can be handled once inside this component and an explicit check won't be needed everywhere this component is being populated from a `map`.

cjones-cs:
Don't subdivide src/components by routes. It only goes in src/components if it is globally reusable.

frontend/src/components/HeaderContainer.tsx
```diff
@@ -1,19 +1,21 @@
...
       px="4"
-      py="8"
+      py="9"
       mx="-4"
       mt="-4"
       mb="4"
       className="bg-[var(--gray-2)] bg-radial from-[var(--gray-3)] from-[1px] to-transparent to-0% bg-size-[4px_4px] bg-repeat-space"
     >
       {children}
     </Container>
```
cjones-cs: I think this container is supposed to be indented.

frontend/src/routes/admin/dashboard/AdminOverviewCard.tsx
```diff
@@ -1,4 +1,4 @@
-import Statistic from 'components/admin/Statistic';
+import Statistic from '@/components/Statistic';
```
cjones-cs: Should just be 'components/Statistic' since we have an alias directly to the components folder. This happens several times in different files

frontend/src/routes/dashboard/EventCard.tsx
```diff
@@ -0,0 +1,43 @@
+import {
+  AspectRatio,
+  Box,
+  Card, Flex, Heading, Inset, Text,
+} from '@radix-ui/themes';
```
cjones-cs, line 3: Either all on one line or all on separate lines

frontend/src/routes/dashboard/PastEvents.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <>
+      <Heading size="6">Your Past Events</Heading>
+      <Grid
+        columns={{
+          initial: '1', xs: '1', sm: '2', lg: '3',
+        }}
+        gap="4"
+      >
+        {events.map((event) => (
```
cjones-cs, line 26: We may consider using lodash here. I installed it on the teamManagement branch, but of course that hasn't been merged in yet.

frontend/src/routes/dashboard/UpcomingEvents.tsx
```diff
@@ -0,0 +1,61 @@
...
+} from '@radix-ui/themes';
+import type { accentColors } from '@radix-ui/themes/props';
+import { TbInfoCircle } from 'react-icons/tb';
+import { Link } from 'react-router';
+import EventCard from './EventCard';
+
+export default function UpcomingEvents({
+  events,
+}: {
+    events: { id: string; name: string; description: string; color?: typeof accentColors[number] }[]
```
cjones-cs, line 12: Since we use this definition for events in several places, do we think there's a way to generalize the type, so we can just import it instead of redefining it multiple times?

frontend/src/routes/dashboard/UpcomingEvents.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <>
+        {heading}
+        <Callout.Root color="jade" variant="surface">
+          <Callout.Icon>
+            <TbInfoCircle />
+          </Callout.Icon>
+          <Callout.Text>
+            You are not registered for any upcoming events.
+            {' '}
+            <RadixLink asChild><Link to="/events">Register for an upcoming event</Link></RadixLink>
```
cjones-cs, line 27: Should we be placing a link inside of another link? The documentation mentioned that the asChild prop basically combines the two components, but you still have to make sure the accessibility is functional.

frontend/src/routes/dashboard/UpcomingEvents.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <>
+      {heading}
+      <Grid
+        columns={{
+          initial: '1', xs: '1', sm: '2', lg: '3',
+        }}
+        gap="4"
+      >
+        {events.map((event) => (
```
cjones-cs, line 49: lodash?

frontend/src/routes/events/Overview.tsx
```diff
@@ -1,26 +1,29 @@
...
-  FaRankingStar, FaTrophy, FaUserGroup,
-} from 'react-icons/fa6';
 import { useSearchParams } from 'react-router';
+import HeaderContainer from 'components/HeaderContainer';
+import { TbCube, TbStar, TbUsersGroup } from 'react-icons/tb';
+
+import EventHeader from 'components/EventHeader';
+import LeaderboardTab from 'routes/events/OverviewTabs/LeaderboardTab';
+import TeamTab from 'routes/events/OverviewTabs/TeamTab';
+import ChallengesTab from './OverviewTabs/ChallengesTab';
```
cjones-cs, line 11: We have a mix of relative and absolute paths here that end up in the same folder. We should make this consistent. You can use the relative path when it's only one degree of separation like this.

frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
```diff
@@ -22,28 +20,30 @@ export default function ChallengeCard({
     completed?: boolean;
     icon?: IconType;
```
cjones-cs: Can you double check this indention? Looks like its too much, but that could just be the github diff looking weird

frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
```diff
@@ -22,28 +20,30 @@ export default function ChallengeCard({
...
         <Flex direction="row" align="center" gap="1">
           {Icon ? (
-            <Heading size="4" color="lime">
+            <Heading size="4" color={color}>
               <Icon />
             </Heading>
           ) : null}
-          <Heading size="4" color="lime">{name}</Heading>
+          <Heading size="4" color={color}>{name}</Heading>
           <Box flexGrow="1" />
```
cjones-cs, line 35: Confused why there is a box here without any content. If you needed separation of space, wouldn't it have been better to use margin or padding?

frontend/src/routes/dashboard/UpcomingEvents.tsx
trevnels-cs: These types will be standardized as the SWR routes are set up. See `types.ts` on the team management branch for an example. For now these are all just placeholders since these components have some extra props that the final ones won't - for example, color here is only used for placeholder graphics and won't be present on the actual model.

frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
trevnels-cs: Since the width of these cards is dynamic, normal margin/padding won't work here. The way this works is that items before the box are aligned to the left of the card, items after the box are aligned to the right of the card, and the empty box grows to fill the remaining space in between. Another way to do this is with `justify-content: space-between`, but that would require additional wrapper divs.

frontend/src/routes/dashboard/UpcomingEvents.tsx
trevnels-cs: The `asChild` means that this will render as a single `<a>` tag that functions as a react-router `Link` and has the styling of a radix `Link`.

cjones-cs (approved):

PR #36, opened 2025-06-14
Team Management - JOINT

cjones-cs:
Here is the scaffold for the team management. We are turning this into a joint branch to integrate swr for fetching from the backend. Pushing to the backend will be separate.


frontend/src/components/Modal.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <Dialog.Content aria-describedby={undefined}>
+        <Flex direction="row">
+          <Dialog.Title
+            className="pb-2"
+          >
+            {title}
+          </Dialog.Title>
+          <Dialog.Close>
+            <button type="button" aria-label="Close" className="absolute left-auto right-4 top-4">
+              <FaXmark />
```
trevnels-cs: This should be replaced with `<TbX />` for consistency

frontend/src/components/Modal.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const [open, setOpen] = useState<boolean>(false);
+
+  return (
+    <Dialog.Root open={open} onOpenChange={setOpen}>
+      <Dialog.Trigger>
+        <Button>{buttonText}</Button>
+      </Dialog.Trigger>
+
+      <Dialog.Content aria-describedby={undefined}>
+        <Flex direction="row">
```
trevnels-cs, line 28: Consider using a gap on this flex container instead of margins/padding on child elements

frontend/src/routes/events/OverviewTabs/TeamTab/AddMemberModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+  inviteCode: string,
+}
+
+export default function AddMemberModal({ inviteCode }: AddMemberModalProps) {
+  const [open, setOpen] = useState<boolean>(false);
+  const inviteURL = `${window.location.origin}/teamSetup/invite/${inviteCode}`;
+
+  return (
+    <Box maxWidth="200px">
+      <Dialog.Root open={open} onOpenChange={setOpen}>
```
trevnels-cs, line 21: Is there a way we could use the `Modal` component here? Seems like this is very similar apart from not having cancel/submit buttons.

frontend/src/routes/events/OverviewTabs/TeamTab/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const inviteCode = 'httptempcode';
+  const membersList = [
+    { id: '1', name: 'cj', role: 'captain' },
+    { id: '2', name: 'md', role: 'captain' },
+  ];
+
+  const transferCaptain = true;
+
+  return (
+    <>
```
trevnels-cs, line 21: I would advise wrapping this in a `<Container size="4">` - while limiting the width of the table like this does result in extra whitespace on the page, it also helps with readability (requiring less eye movement to scan the table contents)

frontend/src/routes/events/OverviewTabs/TeamTab/AddMemberModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <Dialog.Content aria-describedby={undefined}>
+          <Flex direction="row">
+            <Dialog.Title
+              className="pb-2"
+            >
+              Invite Members by shaing this link.
+            </Dialog.Title>
+            <Dialog.Close>
+              <button type="button" aria-label="Close" className="absolute left-auto right-4 top-4">
+                <FaXmark />
```
trevnels-cs: This should also be replaced with `<TbX />` for consistency

RayofLightz (approved):
LGTM

PR #37, opened 2025-06-16
Admin page scaffolding

trevnels-cs:
Initial UI for admin subroutes. Most are designed around a two-column layout, with a datagrid of items on the left, and information about the currently selected item on the right. The selected item will be parameterized in the URL to allow for direct linking. Note that many specifics are currently missing, such as filtering, pagination, and individual form controls.

Pages are currently very reliant on hardcoded placeholder data and there is a lot of duplication. Once our strategy for pagination, sorting, filtering, etc. has been determined, I will look at abstracting the common grid functionality to be reusable if it is realistic to do so.

I also adjusted react-router to use a global basename, so we only need to specify the route prefix once in `main.ts` - all defined routes and links within the router will use this prefix, so we don't need to concatenate `ROUTEPREFIX` everywhere.


frontend/src/components/admin/AdminEventCard.tsx
```diff
@@ -0,0 +1,26 @@
+import { Card, Flex, Heading } from '@radix-ui/themes';
+import Statistic from 'components/admin/Statistic';
+import { TbLink } from 'react-icons/tb';
+import { Link } from 'react-router';
+
+/**
+ * Admin dashboard card for information about an active event.
+ */
```
cjones-cs: You should not subdivide the components folder based on route location. Move the component to the routes/admin folder or make it more generic so it can stay in components

frontend/src/components/admin/Statistic.tsx
```diff
@@ -0,0 +1,40 @@
+import {
+  Badge, Box, Flex, Heading, Text,
+  Tooltip,
```
cjones-cs: I would place these on separate lines or all on one line.

frontend/src/components/admin/Statistic.tsx
```diff
@@ -0,0 +1,40 @@
...
+  label,
+  value,
+  description = undefined,
+  delta = undefined,
+}: {
+    label: string;
+    value: string | number;
+    description?: string;
+    delta?: string;
+}) {
```
cjones-cs, line 22: I think we should determine a standard for defining types. I've been using interface and putting it above. While inline is completely valid, if we have more complex types, it would get cluttered and mix styles.

frontend/src/components/admin/Statistic.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * Displays an emphasized statistic with an optional label,
+ * description (accessible via hover), and delta.
+ *
+ * Combine multiple in a Flex or Grid component for more complex layouts.
+ */
+export default function Statistic({
+  label,
+  value,
+  description = undefined,
+  delta = undefined,
```
cjones-cs, line 16: Do you need to initialize these as undefined? I think that's probably not needed.

frontend/src/routes/admin/containers/index.tsx
```diff
@@ -0,0 +1,99 @@
+import {
+  Callout, Flex, Heading,
+} from '@radix-ui/themes';
+import {
+  TbInfoCircle,
+} from 'react-icons/tb';
+import { AgGridReact } from 'ag-grid-react';
+import type { ColDef } from 'ag-grid-community';
+import { radixTheme } from '../../../grid';
```
cjones-cs: Should probably not use relative paths for this far up the tree. If this is going to be reused, then adding an alias to it may be appropriate. Otherwise, it should exist in a closer area to it's usage.

frontend/src/routes/admin/containers/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+            <TbInfoCircle />
+          </Callout.Icon>
+          <Callout.Text>
+            List of containers that are part of the challenge
+            with statuses, resource usage, and actions.
+          </Callout.Text>
+        </Callout.Root>
+
+        <Heading>Network</Heading>
+        <Callout.Root variant="surface" color="jade">
```
cjones-cs, line 66: Where did jade come from? Weren't we using lime for green coloring?

frontend/src/routes/admin/dashboard/index.tsx
```diff
@@ -1,3 +1,102 @@
...
+                  <Table.ColumnHeaderCell>Containers</Table.ColumnHeaderCell>
+                  <Table.ColumnHeaderCell>Uptime</Table.ColumnHeaderCell>
+                  <Table.ColumnHeaderCell>CPU</Table.ColumnHeaderCell>
+                  <Table.ColumnHeaderCell>Memory</Table.ColumnHeaderCell>
+                  <Table.ColumnHeaderCell>Disk</Table.ColumnHeaderCell>
+                </Table.Row>
+              </Table.Header>
+              <Table.Body>
+                <Table.Row>
+                  <Table.Cell>instance-1</Table.Cell>
```
cjones-cs: I know we probably don't have real data for this yet, but it would probably be good to set up the lodash map for this. Can also be done later when we have real data

frontend/src/components/admin/Statistic.tsx
trevnels-cs: This is an eslint rule - optional props must provide a default value. They're already implied to be undefined but eslint requires that it be made explicit.

frontend/src/routes/admin/containers/index.tsx
trevnels-cs: Typically I've seen \@ defined as an alias to reference the root src directory, so this would be `@/grid`, but I saw that we already had some other aliases configured so I wasn't sure if I should mess with that. Currently there's no way to reference things in the top-level `src` without using relative paths, so I can add the extra alias if that sounds good.

frontend/src/routes/admin/containers/index.tsx
trevnels-cs: I've been using it for these callout boxes as an info color as opposed to the base accent. Typically callouts get semantic colors for info, warning, error, etc. and not the normal accent.

frontend/src/components/admin/Statistic.tsx
cjones-cs: That seems like a dumb, unneccessary rule. This is fine, but we may discuss disabling this rule

cjones-cs (approved):

PR #38, opened 2025-06-18
Ctfd middleware

droge91:
Added decorator based middleware for use with flask routes and moved some decorators from utils/decorators.py.  Also added unit tests for the new decorators.


backend/ctfd/plugin/tests/api/team/test_team_api.py
```diff
@@ -23,7 +23,8 @@ def test_teams_endpoint_requires_authentication(client, event):
 
 def test_teams_endpoint_with_authentication(logged_in_client, event):
     """Check that teams endpoint works for an authenticated user."""
-    response = logged_in_client.get(f"/plugin/api/teams?event_id={event.id}")
+    response = logged_in_client.get(f"/plugin/api/teams?id={event.id}")
+    print(response)
```
fpayer: Remove unnecessary print

fpayer (changes requested):
Fix the ruff error with the unused imports, then this is good to merge in. 

Also, make sure to tag this PR with "minor" for the CI/CD

backend/ctfd/plugin/tests/api/team/test_team_api.py
droge91: removed all unused imports + the extra print. some tests have issues which is causing the test runner to fail but carter's pr should fix that when it gets merged in right after.

PR #40, opened 2025-06-20
Added RichTextEditor component

fpayer:


frontend/src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,61 @@
...
+    placeholder : {
+      text : 'Type here...',
+    },
+  };
+
+  const { } = useEditor((root) => {
+    const crepe = new Crepe({ root, features, featureConfigs, defaultValue : initialValue || ''});
+
+    crepe.editor.use(automd);
+    crepe.editor.config(nord);
```
trevnels-cs, line 49: We should eventually add custom styling that uses Radix CSS variables to match the rest of the site (similar to how ag-grid is handled), though that's not a super high priority.

fpayer: Roger that. 

I'll leave those changes to you you/Caitlin. Caitlin had just asked me last week to add this component so that it could be used in the support ticket interface.

PR #43, opened 2025-06-25
Fixes some serialization bugs and made improvements for the frontend. (MINOR)

CarterPerez-dev:
Standardized the public team list endpoint to GET /events/<event_id>/teams.
Removed the old, ambiguous GET /teams?event_id= endpoint.
Secured the public team list by removing invite codes from the response.
Added New Admin Endpoints:
Added a new admin-only GET /teams to fetch all teams for the admin grid.
Added a new admin-only GET /users/<user_id> to get details for a specific user.
Improved performance of GET /events/<event_id> by no longer returning the full team list.
Fixed a TypeError by correctly serializing datetime objects in API responses.
Fixed an AttributeError in the event creation logic by correctly parsing datetime strings.


backend/ng/event/routes/events.py
```diff
@@ -109,20 +104,19 @@ def post(self):
...
-        name = data.get("name")
-        description = data.get("description")
-        max_team_size = data.get("max_team_size")
-        start_time = data.get("start_time")
-        end_time = data.get("end_time")
-        locked = data.get("locked", False)
+        start_time_str = data.get("start_time")
+        end_time_str = data.get("end_time")
+
+        start_time_obj = datetime.fromisoformat(start_time_str.replace("Z", "+00:00")) if start_time_str else None
```
fpayer: I'm concerned about running a `.replace` on the user input here. It's an indication that the input isn't validated to be in the format you want it in. Does there need to be a change to the validator?

backend/ng/user/routes/users.py
```diff
@@ -51,6 +52,37 @@ def get(self):
...
+            404: "Not Found - User does not exist",
+        },
+    )
+    def get(self, user_id):
+        """Get detailed information for a specific user."""
+        result = get_user_info(user_id)
+        if not result.get("success"):
+            return controller_response(result, error_field="user"), 404
+
+        logger.info(
```
fpayer: Unnecessary logging

fpayer (changes requested):
General feedback - I think you're likely logging more than necessary, which can add some bloat to the code and make it harder to read. We don't have specific guidelines for logging, but I would say that logging read-only activity is not necessary in the majority of situations. 

Also want to get Trevor's sign-off before merging.

backend/ng/admin/controllers/get_detailed_stats.py
```diff
@@ -26,23 +25,34 @@ def get_detailed_stats() -> dict[str, Any]:
...
         .all()
     )
 
-    event_stats = rows_to_dicts(event_stats_query)
+    event_stats_raw = rows_to_dicts(event_stats_query)
+
+    event_stats_serialized = []
+
+    for stat in event_stats_raw:
+        if stat.get("start_time"):
```
fpayer: Should have a `serialize` function in the model

backend/ng/event/controllers/get_event_info.py
```diff
@@ -29,20 +23,30 @@ def get_event_info(event_id: int) -> dict[str, Any]:
...
         return {"success": False, "error": "Event not found."}
 
-    result = event.get_event_details_with_teams()
-    event_data = result["event"]
-    teams_data = result["teams"]
+    total_members = TeamMember.count_by_event(event.id)
+    team_count = Team.count_by_event(event.id)
+
+    event_data = {
+        "id": event.id,
```
fpayer: Should have a serialize function in the model

backend/ng/team/routes/teams.py
```diff
@@ -49,44 +48,31 @@
...
-        """
-        event = kwargs["event"]
-        result = list_teams_in_event(event.id)
+    def get(self):
+        """Get all teams in the system for the admin grid."""
+        result = list_all_teams()
 
         if result["success"]:
             logger.info(
-                "Teams list retrieved",
```
fpayer, line 79: Unnecessary log

backend/ng/core/utils/api_responses.py
```diff
@@ -58,3 +58,16 @@ def controller_response(
     else:
         error_msg = result.get("error", "Unknown error occurred")
         return error_response(error_msg, error_field, 400)
+
+
+def serialize_item_datetimes(item: dict) -> dict:
```
trevnels-cs, line 63: I would recommend making this handle all datetime objects instead of specific keys and including this logic in `serialize_model_for_api` instead of individual routes, since there's some other routes that return datetimes as well that aren't calling this function. (`GET /team/<team_id>`'s team members field, for example.) Also, as we add more functionality, having to add special cases here and call this function inside each route that needs them is not as maintainable long term.

Currently, `serialize_model_for_api` works on actual model objects via `__dict__`, we would just need to have it handle arrays and normal objects as well in a similar way.

CarterPerez-dev:
I closed this PR since it feels redundant considering all these changes will be bundled with the new support functionality which I'll be submitting tmrw anyway.

PR #45, opened 2025-06-26
Vite build

fpayer:
Set up configuration to support a "prod" deployment using built vite assets.

**NOTE**: This PR will require devs to re-run `install.sh`

Run `yarn start` as normal to start the python webapp in dev mode with the vite HMR server.

Run `yarn run prod-start` to build the vite assets to be served with nginx and start the python webapp in prod mod.


frontend/index.html
```diff
@@ -1,3 +1,4 @@
+<!-- This is only used for the local dev vite server. It does not get used by the main CTF-ng application -->
```
fpayer: @trevnels-cs with my new changes, do we still need this `index.html` file or can we delete it?

frontend/.eslintrc.cjs
```diff
@@ -1,53 +1,56 @@
...
       },
     ],
-    'react/jsx-filename-extension': ['error', { extensions: ['.js', '.jsx', '.ts', '.tsx'] }],
-    'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
-    'react/require-default-props': ['error', {
-      functions: 'defaultArguments', // Default props must now be defined in arguments - .defaultProps was removed in React 19
-    }],
+    'react/jsx-filename-extension' : [ 'error', { extensions : [ '.js', '.jsx', '.ts', '.tsx' ] } ],
+    'import/no-extraneous-dependencies' : [ 'error', { devDependencies : true } ],
+    'react/require-default-props' : [ 'error', {
```
trevnels-cs: While we're making eslint changes, I believe Caitlin and I agreed to disable the `require-default-props` rule. Any thoughts on that?

frontend/index.html
trevnels-cs: I just tested and Vite cannot build if it's removed. Even if we're not directly using it, it's still treated as the main entrypoint to the build for module resolution.

See https://vite.dev/guide/#index-html-and-project-root

frontend/.eslintrc.cjs
fpayer: No strong feelings on this. I'm good to remove. Want me to make that change in this update?

frontend/index.html
fpayer: Ok, will leave it in then.

PR #47, opened 2025-06-27
Service import

RayofLightz:
Imports a yaml file into the DB


backend/ng/core/routes/__init__.py
```diff
@@ -11,6 +11,8 @@
 from ...event.routes.events import events_namespace
 from ...user.routes.users import users_namespace
 from ...admin.routes.admin import admin_namespace
+from ...challenge.routes.routes import challenge_namespace
+print(challenge_namespace)
```
fpayer: Remove debug print

backend/ng/challenge/controllers/import_yaml.py
```diff
@@ -0,0 +1,36 @@
...
+from ..models.Hint import Hint
+from ..models.Tag import Tag
+from ..models.Question import Question
+from ..models.Challenge import Challenge
+
+def import_yaml(yaml):
+    try:
+        parsed_yaml = asdict(parse_compose_string(yaml), filter=lambda y, x: x is not None)
+
+        challenge_fields = {
```
trevnels-cs, line 14: Does `summary` need to be included here?

backend/ng/core/routes/__init__.py
trevnels-cs: Can this be removed?

backend/ng/challenge/models/ContainerBlueprint.py
```diff
@@ -0,0 +1,35 @@
+from CTFd.models import db
+
+class ContainerBlueprint(db.Model):
+    __tablename__ = 'ng_container_blueprint'
+
+    id = db.Column(db.Integer, primary_key=True)
+    image = db.Column(db.String(255))
+    hostname = db.Column(db.String(255))
+    stdin_open = db.Column(db.Boolean, nullable=True)
```
fpayer, line 9: Should this be `nullable=False` to enforce a True/False choice instead of an implicit null=False?

backend/ng/challenge/models/Tag.py
```diff
@@ -0,0 +1,25 @@
+from CTFd.models import db
+
+class Tag(db.Model):
+    __tablename__ = 'ng_challengetag'
+    id = db.Column(db.Integer, primary_key=True)
+    challenge_id = db.Column(db.Integer, db.ForeignKey('ng_challenge.id'), nullable=False, index=True)
+    tag = db.Column(db.String(255), nullable=False, index=True)
+
+    challenge = db.relationship('Challenge', back_populates='tags')
```
fpayer, line 9: Does this support having the same tag be used on multiple challenges?

fpayer (changes requested):

backend/ng/challenge/models/Tag.py
RayofLightz: Yes it should since tag name is not specified as being unique.

backend/ng/challenge/models/ContainerBlueprint.py
RayofLightz: I will move it to implicit to match the rest of the code base.

fpayer (approved):

backend/ng/challenge/models/Question.py
```diff
@@ -0,0 +1,28 @@
+from CTFd.models import db
+
+class Question(db.Model):
+    __tablename__ = 'ng_challengequestion'
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(255), nullable=False)
+    question = db.Column(db.String(255), nullable=False)
```
fpayer: Would recommend renaming this from "question" to "text". That would avoid confusing usage (e.g. accessing `question.question`)

backend/ng/challenge/models/Hint.py
```diff
@@ -0,0 +1,27 @@
+from CTFd.models import db
+
+class Hint(db.Model):
+    __tablename__ = 'ng_challenge_hint'
+    id = db.Column(db.Integer, primary_key=True)
+    challenge_id = db.Column(db.Integer, db.ForeignKey('ng_challenge.id'), nullable=False, index=True)
+    preview = db.Column(db.String(255), nullable=False)
+    hint = db.Column(db.String(255), nullable=False)
```
fpayer: Suggest renaming this from "hint" to "text"

backend/ng/challenge/models/Tag.py
```diff
@@ -0,0 +1,25 @@
+from CTFd.models import db
+
+class Tag(db.Model):
+    __tablename__ = 'ng_challengetag'
+    id = db.Column(db.Integer, primary_key=True)
+    challenge_id = db.Column(db.Integer, db.ForeignKey('ng_challenge.id'), nullable=False, index=True)
+    tag = db.Column(db.String(255), nullable=False, index=True)
```
fpayer: Suggest renaming this from "tag" to "name" or something else

RayofLightz:
Bumping @johnmarsden-edu on the name change stuff as it should be consistent between the yaml file and the db

fpayer:
@johnmarsden-edu for context:

This issue with naming is that while this makes sense as yaml:

```
questions:
  - name: flag
    question: What is the flag?
```

The resulting object once the yaml is imported would look like this:

```
{ questions : [ 
  { name : "flag", question : "What is the flag?"
] }
```

Which would subsequently lead to code that looks like

```
question = questions[0]
return question.question
```

The `question.question` is kinda weird. 

=====

What are your thoughts on these renames?

```
questions
  question -> text
hints:
  hint -> text
```

johnmarsden-edu:
Yeah, that's fine. I'll just need to adjust the tests.

johnmarsden-edu:
After looking at it for a little while I decided that a better term to use there would be body. Especially because I allow for multiple different types of hints and it feels more descriptive of what actually goes in there.

RayofLightz:
Uses body now plus change tag to name

fpayer (approved):

PR #49, opened 2025-06-30
Hotfix for gitignored env files

fpayer:


zoecough (approved):
tested locally, everything seems good

PR #55, opened 2025-07-02
Created a setup script to configure ctfd

fpayer:
Run using `yarn setup-ctfd`


package.json
```diff
@@ -18,19 +18,20 @@
...
     "reload": "docker exec -it ng-ctfd supervisorctl restart ctfd",
     "restart": "docker exec -it ng-ctfd supervisorctl restart ctfd",
-    "restart-nginx": "docker exec -it ng-nginx service nginx restart",
-    "logs-nginx": "docker compose logs -f nginx",
-    "logs-vite": "docker compose logs -f vite",
-    "logs-ctfd": "docker compose logs -f ctfd",
+    "logs-nginx": "docker logs --tail=1000 ng-nginx -f",
+    "logs-ctfd": "docker logs --tail=1000 ng-ctfd -f",
+    "logs-vite": "docker logs --tail=1000 ng-vite -f",
+    "errors-ctfd": "docker exec -it ng-ctfd cat /var/log/supervisor/ctfd.error.log",
```
trevnels-cs, line 34: Also consider setting `redirect_stderr=true` in the supervisord progam configuration to send errors to stdout. This way, errors and normal output will both be part of the docker log stream and we won't need a separate script for each

fpayer: Interesting idea. I'm gonna for people's preferences on this. Cause I think having a separate stream for `stderr` can be helpful since there will be less noise and there doesn't seem to be a native way to have both a combined stderr+stdout stream and a separate stderr stream.

PR #57, opened 2025-07-03
Add Test Case Support to Answers and Rename Question and Hint to Body

johnmarsden-edu:


fpayer:
Yup this is exactly what I was imagining with the testcases. Will review later today.

lib/parser/src/cyber_skyline/chall_parser/compose/challenge_info.py
```diff
@@ -31,19 +32,24 @@ class Question:
...
+class TextBody:
     """A text-based hint for players."""
     type: Literal['text']
     content: str  # The actual hint text content
 
 # @define
-# class ImageHint:
+# class ImageBody:
 #     type: Literal['image']
 #     source: str
```
RayofLightz, line 55: How likely are we to actually need images in hints? I feel like this would be a weird thing to define in a yaml spec.

RayofLightz (commented):
I know Franz is reviewing as well but only one thing on my end

fpayer:
@johnmarsden-edu can you make that update to the hints and then this is good to merge

fpayer (approved):

lib/parser/src/cyber_skyline/chall_parser/compose/challenge_info.py
fpayer: I'm leaning towards no to have images in hints. Seems like extra scope.

johnmarsden-edu: That's why this is commented out. It still lives there just as an example of what it might look like to add different types of hints but it isn't in use.

PR #58, opened 2025-07-03
Frontend SWR setup

trevnels-cs:
Current SWR setup cherry-picked from admin panel branch. This should move to development so other feature branches have a consistent data fetching base to work from.

Note that a few hooks and types are currently missing/incomplete pending further backend progress, but they are in a good enough state to start consuming from components and can be refined as the backend stabilizes.

### DIFF
--- a/frontend/src/constants.ts
+++ b/frontend/src/constants.ts
@@ -1,2 +1,2 @@
 export const ROUTEPREFIX: string = '/hello';
-export const APIPREFIX: string = '/plugin/api';
+export const APIPREFIX: string = '/ng';
--- a/frontend/src/constants.ts
+++ b/frontend/src/constants.ts
@@ -1,2 +1,2 @@
-// eslint-disable-next-line import/prefer-default-export
 export const ROUTEPREFIX: string = '/hello';
+export const APIPREFIX: string = '/plugin/api';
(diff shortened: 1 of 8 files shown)


frontend/src/constants.ts
```diff
@@ -1,2 +1,2 @@
-// eslint-disable-next-line import/prefer-default-export
 export const ROUTEPREFIX: string = '/hello';
+export const APIPREFIX: string = '/ng';
```
fpayer, line 2: Should the backend get updated to reflect this?

trevnels-cs: This is updating to reflect the backend - routes moved from /plugin/api to /ng a while ago

frontend/src/fetchers.ts
```diff
@@ -0,0 +1,79 @@
...
+    // If we got a stack trace, append it to the error message.
+    if (stackTrace) {
+      message += `\n\n${stackTrace}`;
+    }
+
+    throw new Error(message);
+  }
+
+  // At this point, we should have a success/failure JSON response from the API.
+  if (!parsedData.success) {
```
fpayer, line 39: Not considering how the backend is currently implemented, would it be better to check the HTTP status code instead of using `parsedData.success`?

fpayer (approved):

PR #59, opened 2025-07-07
Feature/user support

cjones-cs:


frontend/src/Routes.tsx
```diff
@@ -27,37 +31,44 @@ import AdminSettings from 'routes/admin/settings';
...
-      children: [
-        { index: true, element: <Overview /> },
-        { path: 'registration', element: <Registration /> },
-        { path: 'challenge/:idChallenge', element: <Challenge /> },
+      path : '/support',
+      children : [
+        { index : true, element : <Support /> },
+        { path : 'createTicket', element : <CreateTicket /> },
+        { path : ':idTicket', element : <TicketDetail /> },
       ],
```
RayofLightz, line 55: In the future is there a good way to split out the route objects into separate files? I have a feeling that this could get really long.

frontend/src/components/StatusBadge.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const CLOSED = 'closed';
+const IN_PROGRESS = 'inprogress';
+
+function StatusBadge(status: string) {
+  switch (status) {
+    case CLOSED: return <Badge color="jade">Closed</Badge>;
+    case IN_PROGRESS: return <Badge color="blue">In Progress</Badge>;
+    default: return <Badge color="orange">Open</Badge>;
+  }
```
RayofLightz, line 12: I am surprised eslint is not upset about this I thought switches are disallowed with airbnb, Unless TS switches are treated differently.

frontend/src/routes/support/Detail.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { TbArrowLeft } from 'react-icons/tb';
+import { useNavigate } from 'react-router';
+import { StatusBadge } from 'components/StatusBadge';
+// import Editor from 'components/Editor'; //milkdown
+import { map } from 'lodash';
+/* import ReactMarkdown from 'react-markdown';
+import remarkGfm from 'remark-gfm';
+import remarkBreaks from 'remark-breaks';
+import rehypeRaw from 'rehype-raw';
+import rehypeSanitize from 'rehype-sanitize'; */
```
RayofLightz, line 13: Delete commented out code?

frontend/src/routes/support/Detail.tsx
```diff
@@ -0,0 +1,98 @@
...
+    { user : 'user2', timestamp : '2025-11-1', text : 'blah' },
+    { user : 'user3', timestamp : '2025-11-1', text : 'blah' },
+    { user : 'user4', timestamp : '2025-11-1', text : 'blah' },
+  ];
+
+  const navigate = useNavigate();
+
+  function resolveTicket() {
+    // do server call to resolve ticket
+    console.log('Resolved Ticket');
```
RayofLightz: Guessing this is here since there is no server test data?

frontend/src/routes/support/Detail.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <Flex direction="row" gap="4">
+      <Flex gap="3" direction="column" className="w-4/5">
+        <Box maxWidth="200px">
+          <Button
+            variant="ghost"
+            color="lime"
+            onClick={() => { navigate('/support'); }}
+          >
+            <TbArrowLeft />
+            {' '}
```
RayofLightz: Are these blank strings just for line breaks? Why not use a `<br>`

RayofLightz (approved):
Honestly most of these are questions and less of hard issues that I have. I don't know if you wrote it but if while you are at it you can fix the linter issue with the `alert` call in `frontend/src/components/Navbar.jsx:68` that would be excellent

frontend/src/routes/support/Detail.tsx
cjones-cs: I would normally agree with you, but I left it there for now because it was sent to me in a comment and I didn't want to lose it when I implement the milkdown editor. I'm not currently sure when I will have time to do that; probably in a few more weeks.

frontend/src/routes/support/Detail.tsx
cjones-cs: Yes

frontend/src/routes/support/Detail.tsx
cjones-cs: It's not a new line, only a space:  [<- Support]

cjones-cs: Removing

frontend/src/components/StatusBadge.tsx
cjones-cs: It didn't complain about it except to demand I include a default case. The default case here is set to Open, because that makes sure the ticket is resolved instead of overlooked.

frontend/src/Routes.tsx
cjones-cs: We should keep the route definitions in one place since we are running a single page app

PR #60, opened 2025-07-07
Role based Permissions Middleware and Admin Routes

droge91:
Added models Role, UserRole, Permission, and RolePermission
added middleware to check permissions of Users based on user role or role in team
added routes for admins to configure permissions associated with roles and what roles users have.
added to default_ctfd_setup to create an initial admin user and setup basic permissions and roles


fpayer (changes requested):
Can you make the different Roles and Permissions enums instead of raw strings? That'll help us avoid any bugs caused by typos in the string.

backend/ng/Makefile
```diff
@@ -11,7 +11,7 @@ endif
 
 # All tests
 test-all:
-	$(PYTHON_EXEC) -m pytest .
+	$(PYTHON_EXEC) -m pytest -m "not middleware" -v
```
fpayer, line 14: Any reason for this?

backend/ng/core/middleware/permission_middleware.py
```diff
@@ -0,0 +1,67 @@
...
+
+    """
+
+    @wraps(f)
+    def wrapped(*args, **kwargs):
+        user = get_current_user()
+        if not user:
+            return error_response("User not authenticated", "unauthorized", 401)
+        
+        team_id = request.view_args.get('team_id') or request.args.get('team_id')
```
fpayer: You should get the team directly from the kwargs since the team loader middleware should come first.

backend/ng/core/middleware/permission_middleware.py
```diff
@@ -0,0 +1,67 @@
...
+        if not team_id:
+            return error_response("Team ID is required", "bad_request", 400)
+        response = get_team_management_permissions(team_id,user.id)
+        if "error" in response:
+            return error_response(response["error"], "team_management", 400)
+
+        if kwargs.get('permissions') is None:
```
fpayer, line 57: You need to check if the wrapped function has 'permissions' in its kwargs or an error will get thrown. See https://github.com/CyberSkyline/ctf-ng/blob/development/backend/ng/core/middleware/loaders/_util.py#L67 for reference

backend/ng/core/tests/system/middleware_test_routes.py
```diff
@@ -0,0 +1,87 @@
...
+
+@middleware_test_routes.route("/admin_decorator_test", methods=["GET"])
+@admin_endpoint(
+    json_required=False,
+)
+def admin_decorator_test(**kwargs):
+    """
+    Endpoint to test the admin decorator.
+    """
+    print("gets here")
```
fpayer, line 45: delete

backend/ng/Makefile
droge91: The middleware tests have to be run isolated from the normal tests, It was like this before the refactor I just changed it back since there are middleware tests again

backend/ng/permissions/routes/admin_routes.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+@permissions_admin_namespace.route("/<int:user_id>/roles")
+class UserRoles(Resource):
+    """
+    Resource to manage roles for a specific user.
+    """
+    @api_endpoint()
+    @permissions_admin_namespace.doc("get_user_roles")
+    def get(self, user_id):
```
fpayer: Use the `load_user` middleware. It'll throw an error if the user is DNE so you don't need to handle that in the controller.

backend/ng/permissions/routes/admin_routes.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+        Get all roles for a specific user by ID.
+        """
+
+        response = get_user_roles(user_id)
```
fpayer: You can call `UserRole.get_user_roles` directly so you can delete the controller and remove redundant code.

backend/ng/permissions/routes/admin_routes.py
```diff
@@ -0,0 +1,142 @@
...
+        roles = response.get("roles", [])
+        return {
```
fpayer: Just return the roles array and skip the serialize. The middleware will add in the `success` = True and run `serialize` on the array for you. See: https://github.com/CyberSkyline/ctf-ng/blob/development/backend/ng/core/utils/api.py#L12

backend/ng/permissions/routes/admin_routes.py
```diff
@@ -0,0 +1,142 @@
...
+        if not data or "roles" not in data:
+            return error_response("No role names provided", "roles", 400)
+
+        User.query.get_or_404(user_id)
+
+        response = update_user_roles(user_id, data)
+        if "error" in response:
+            return error_response(response["error"], "user", 400)
+
+        return {
```
fpayer: No need for the message, just return the user

backend/ng/permissions/controllers/update_role.py
fpayer, line 1: Move function to the model itself

backend/ng/permissions/routes/admin_routes.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Resource to manage permissions for a specific role.
+    """
+    @admin_endpoint()
+    @permissions_admin_namespace.doc(
+        description="Get all details for a specific role by ID",
+        responses={404: "Role not found", 200: "Success"},
+        params={"role_id": "Role ID to retrieve details for"},
+    )
+    def get(self, role_id):
```
fpayer, line 30: Create a load_role middleware instead of having to perform the lookup in the route/controller

backend/ng/permissions/routes/admin_routes.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+        role_details = get_role_details(role_id)
+        if not role_details.get("success"):
+            return error_response(role_details.get("error", "Role not found"), "role", 404)
+        role = role_details.get("role")
+        users = role_details.get("users", [])
+        return {
+            "success": True,
+            "role": role.serialize(),
+            "users": [user.serialize() for user in users],
```
fpayer: Just return the role. No need for the users list

backend/ng/permissions/routes/admin_routes.py
```diff
@@ -0,0 +1,142 @@
...
+
+        },
+        params={"role_id": "Role ID to update"},
+        body={
+            "name": "Name of the role",
+            "description": "Description of the role",
+            "permissions": "List of permission names to assign to the role"
+        }
+    )
+    def patch(self, role_id):
```
fpayer: Use the `validated_data` kwarg instead of `g.json_data`. `g.json_data` won't be sanitized through the validator.

backend/ng/permissions/routes/admin_routes.py
```diff
@@ -0,0 +1,142 @@
...
+        role = Role.query.get_or_404(role_id)
+        if not role:
+            return error_response("Role not found", "role", 404)
+
+        response = update_role(role_id, data)
+        if "error" in response:
+            return error_response(response["error"], "role", 400)
+        return {
+            "success": True,
+            "role": response.get("role", {}).serialize(),
```
fpayer: Just return role

fpayer (changes requested):
Looking good. Most of my feedback is to switch to some of the new ways of doing things post my refactor:

- If all a controller does is call a function on a model, you can avoid writing a separate file for the controller and just call that function directly within the route
- Also, if the functionality of a controller is likely to be used again elsewhere (e.g. get_xyz, or update_xyz). It's better to put that function in a model so it can be re-used more easily. Controllers are not supposed to be imported for anything other than to attach to be attached to a route. 
- Avoid using Flask's `g` variable. You should be able to get everything you need via middleware + the kwargs now  
- Use the middleware loaders instead of performing the db lookup on an id within the route/controller. These will also handle the 404 for you if it DNE.
- You don't need to return "success" : True / False in your routes. If it succeeds, just return normally. If it fails, throw an uncaught exception. The middleware wrapper will handle the rest. 
- You also don't need to call .serialize anymore. The wrapper handles that, too

backend/ng/core/middleware/permission_middleware.py
droge91: So i'm not sure what error you are referring to here, as this middleware works without issues on the testcases.
this logic in the loader confuses me as it seems unnecessary. the wrapped route will take any kwarg as long as **kwargs is an argument.

droge91: We can require loaded models to be arguments in the routes if we want so its easier to see whats being gotten but it doesn't appear to be functionally necessary

backend/ng/permissions/routes/admin_routes.py
droge91: validated_data appears to be attached to g and not a kwarg? is that intended?

droge91: looks like that issue is fixed in trevors PR

backend/ng/core/middleware/permission_middleware.py
fpayer: I was running into issues where is the injected kwarg was missing from the function signature of the wrapped function, an error would get thrown, which is why I added that inspect code in my linked middleware. If that's not an issue here, then it's all good.

fpayer (approved):

PR #62, opened 2025-07-08
support tests

CarterPerez-dev:
Created support tests.

### DIFF
--- a/backend/ng/support/tests/test_support_models.py
+++ b/backend/ng/support/tests/test_support_models.py
@@ -1,0 +1,317 @@
+"""
+Model tests for support domain
+"""
+
+import pytest
+from datetime import datetime, timedelta
+from ...core.exceptions import ValidationError
+from ...core.utils import utc_now
+from ..models.Ticket import Ticket
+from ..models.TicketMessage import TicketMessage
+from ..models.TicketTag import TicketTag
+
+
+class TestTicketModel:
+    """Tests for the Ticket model."""
+
+    def test_repr(self, ticket):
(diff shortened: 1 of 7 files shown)


backend/ng/support/tests/test_support_models.py
fpayer, line 1: For all of the tests that are making an update to the database (insert/update/delete), the test should query the database to confirm that the change was committed to the database. Otherwise, it's possible the change could just be made on the variable, but not saved to the database.

fpayer (changes requested):

fpayer:
Also, there should be one "test class" for each function being tested to make it more clear which what tests correspond go what functions. 

`test_support_models.py` should be split into a file for each model:
- `test_support_ticket_model.py`
- `test_support_ticket_message_model.py`
- `test_support_ticket_tag_model.py`

backend/ng/support/tests/test_support_models.py
fpayer: Actually, if you re-query the database with the same session, you will get the updated values even if `commit` is not called. You should instead use a mock to check if commit was called or not. 

e.g.
```
    def test_should_respect_commit_flag(self, event, team_factory, db_session):
        team = team_factory(event=event)

        with patch.object(db_session, 'commit') as mock_commit:
            team.update_invite_code(commit=False)
            mock_commit.assert_not_called()

        with patch.object(db_session, 'commit') as mock_commit:
            team.update_invite_code(commit=True)
            mock_commit.assert_called_once()
```

fpayer:
Closing, migrated to #66

PR #63, opened 2025-07-08
Adding extra routes and controllers for EventRegistration

droge91:
Adding additional routes to EventRegistration for retrieving event_registration objects and updating them from the frontend. as well as an endpoint to retrieve all events currently available for registration.


fpayer:
Copied over functionality into a separate PR

PR #65, opened 2025-07-09
Remove version req in ng

RayofLightz:
Remove attrs version req in ng


fpayer (approved):

fpayer:
Github actions still looks angry?

fpayer:
Does John's other PR make this one redundant?

RayofLightz:
Yeah John's makes this redundant. We had to move the parser back to an older version of attrs because pip's solution to the issue of two deps requiring different versions of the same package is to just not deal with it.

PR #67, opened 2025-07-10
Todo support

CarterPerez-dev:


CarterPerez-dev:
Sorry for the ruff format if it makes it harder to review, didn't think 74 files needed formatting, to clarify though, do we want ruff format, or are we just using ruff to check?

fpayer:
Yeah, can you roll those back? I just meant `ruff check`

PR #68, opened 2025-07-10
Added script to populate sample data in the database

fpayer:


.yarn-scripts/populate_data.sh
```diff
@@ -0,0 +1,15 @@
...
+
+DIR=$(dirname $BASH_SOURCE)
+source $DIR/utils.sh
+
+if ! docker ps --format '{{.Names}}' | grep -q '^ng-ctfd$'; then
+  echo "The 'ng-ctfd' container is not running. Please run 'yarn start' to start the container"
+  exit 1
+fi
+
+prompt_user "Would you like to reset and populate the database with sample data? " && {
```
RayofLightz: Add a note about how this will completely remove the current db. Reset seems obvious, but you might want to have it in big bold letters lol

backend/pyproject.toml
```diff
@@ -0,0 +1,83 @@
...
+]
+
+[tool.ruff.lint.per-file-ignores]
+"**/test_*.py" = ["F401", "F811"]  # Allow unused imports in tests
+"**/conftest.py" = ["F401", "F811"]  # Allow unused imports in pytest config
+"**/scripts/*.py" = ["T201"]  # Allow print statements in scripts
+
+[tool.ruff.format]
+# Use double quotes for strings
+quote-style = "double"
```
RayofLightz, line 70: I am fine if we want to use double quote, but if we want to follow pep8 for style we should be using single quotes irrc

backend/ng/scripts/default_ctfd_setup.py
```diff
@@ -1,21 +1,51 @@
...
 from tests.helpers import setup_ctfd
-import os
 
 if "SCRIPT" not in os.environ:
-    raise EnvironmentError('This should only be run from a script. DO NOT run this manually.')
+    raise OSError("This should only be run from a script. DO NOT run this manually.")
+
+DEFAULT_ADMIN_EMAIL = "admin@examplectf.com"
+DEFAULT_ADMIN_PASSWORD = "ctfng_password"
+
```
RayofLightz, line 17: Print out the default creds for the user.

RayofLightz (approved):
It all looks good to me. Just some minor stuff to make it more clear to the end user whats going on in the scripts.

.yarn-scripts/populate_data.sh
fpayer: Adding in yellow text highlighting

backend/ng/scripts/default_ctfd_setup.py
fpayer: Will do

backend/pyproject.toml
fpayer: I don't have a preference either. I was just noticing that the majority of the codebase is using doublequotes and it doesnt look like the ruff linter is actually warning on that.

PR #70, opened 2025-07-10
Backend fixes for admin work

trevnels-cs:
Applies the following changes to backend code:
- Use `kwargs` consistently for passing data into route handlers. `g.` is now only used for giving validators access to the incoming JSON data. Some things were previously not working due to this inconsistency.
- Sets up endpoint for getting teams for the current user, and equivalent admin endpoints for getting events/teams of any user
- Adds a route for admins to get the TeamMembers on a given team
- Fixes double `/users` in path for all users endpoint
- Validators now enforce that incoming ISO timestamps are in UTC
- ISO timestamps are explicitly specified as UTC in model `serialize` methods (`Z` is appended onto the naive datetimes)

Additionally, frontend data fetching code has been updated to account for recent backend changes. The frontend will also now automatically parse ISO dates into JS Date objects as part of SWR logic, so that components don't need to do that conversion themselves.

Admin frontend still needs a bit more cleanup before it's ready to merge, but it is now working with these changes applied - don't want these fixes to be blocked by that.

Backend tests that involve timestamp validation will likely need adjustment for the new logic, as the output of  `utc_now` -> `isoformat()` does _not_ explicitly specify UTC. We'll either want to append `Z` or attach tzinfo to the produced datetime.

### DIFF
--- a/backend/ng/core/middleware/auth.py
+++ b/backend/ng/core/middleware/auth.py
@@ -10,7 +10,7 @@ from CTFd.utils.user import get_current_user, is_admin as is_admin_ctfd
 from CTFd.utils.decorators import admins_only, authed_only
 from ..exceptions import PermissionError, ValidationError
 from .error_handler import handle_exceptions
-
+from ...user.models.User import User
 
 def api_endpoint(auth_required=True, admin_required=False, json_required=False, validation_func=None):
     """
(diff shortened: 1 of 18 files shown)


fpayer (approved):

fpayer:
Looks good, no notes

PR #75, opened 2025-07-15
Challenge model updates

fpayer:
Checkpoint before I make more changes. Want to make sure these changes get merged now to avoid more merge conflicts in the future. 

- Updated table names to be more standardized using pluralized names
- Created validation test cases for challenge-related models
- Moved yaml upload route


backend/ng/challenge/models/ChallengeTag.py
```diff
@@ -1,25 +1,68 @@
...
+
+    @classmethod
+    def create_tag(cls, challenge_id: int, name: str, commit=True):
         try:
-            tag = cls(**kwargs)
+            validated_data = cls.validate({"name": name, "challenge_id": challenge_id})
+
+            tag = cls(**validated_data)
             db.session.add(tag)
+            db.session.flush()
```
RayofLightz, line 54: Allegedly  sql alchemy  autoflushes by default so you shouldn't need to manually flush. Unless ctfd changed this.

backend/ng/challenge/controllers/import_challenge_from_yaml.py
```diff
@@ -0,0 +1,55 @@
...
+        print(e)
+        raise ValidationError(f"Invalid YAML format: {e}") from e
+
+    try:
+        challenge_yaml = yaml["challenge"]
+        challenge = Challenge.create_challenge(
+            name=challenge_yaml["name"],
+            icon=challenge_yaml["icon"],
+            description=challenge_yaml["description"],
+            summary=challenge_yaml["summary"],
```
RayofLightz: For Stuff like icon and summary you might want to wrap it a `get` function call. As these are optional fields and without them you will throw a key error.

backend/ng/challenge/models/ChallengeTag.py
```diff
@@ -1,25 +1,68 @@
...
-    name = db.Column(db.String(255), nullable=False, index=True)
+    challenge_id = db.Column(db.Integer, db.ForeignKey("ng_challenges.id"), nullable=False, index=True)
+    name = db.Column(db.String(MAX_CHALLENGE_TAG_NAME_LENGTH), nullable=False, index=True)
 
-    challenge = db.relationship('Challenge', back_populates='tags')
+    challenge = db.relationship("Challenge", back_populates="tags")
 
     def __repr__(self):
-        return f'<NgChallengeTag {self.id}>'
+        return f"<NgChallengeTag {self.id}, name={self.name}>"
```
RayofLightz, line 19: Do we at all have a standard for how we are doing the __repr__?

backend/ng/challenge/models/ContainerBlueprint.py
```diff
@@ -1,32 +1,119 @@
...
+            friendly_name="Container Hostname",
+        )
+        validator.validate_positive_integer(
+            data,
+            "challenge_id",
+            required=True,
+            friendly_name="Challenge ID",
+        )
+
+        is_valid, errors, parsed_data = validator.is_valid()
```
RayofLightz: I am not saying we should change this but this is a really weird way of returning data

backend/ng/challenge/models/ContainerBlueprint.py
```diff
@@ -1,32 +1,119 @@
...
+from ...core.exceptions import ValidationError
+from ...core.validation import BaseValidator
+
+MAX_CONTAINER_BLUEPRINT_IMAGE_LENGTH = 1024
+MAX_CONTAINER_BLUEPRINT_HOSTNAME_LENGTH = 256
+MAX_CONTAINER_BLUEPRINT_MEM_LIMIT_LENGTH = 256
+MAX_CONTAINER_BLUEPRINT_MEMSWAP_LIMIT_LENGTH = 256
+MAX_CONTAINER_BLUEPRINT_USER_LENGTH = 256
+
+DEFAULT_CONTAINER_BLUEPRINT_CPUS = 1.0
```
RayofLightz: This is actually way to high of a default. Reminds me I need to talk to John about removing the ability for challenge devs to change cpu limits

backend/ng/challenge/tests/test_challenge_tag_model.py
```diff
@@ -0,0 +1,256 @@
...
+        with pytest.raises(ValidationError):
+            ChallengeTag.create_tag(
+                challenge_id=-1,  # Invalid challenge_id
+                name="test",
+                commit=False,
+            )
+
+        # Verify no tag was created
+        tags = ChallengeTag.query.filter_by(name="test").all()
+        assert len(tags) == 0
```
RayofLightz: Will this actually test this? With commit=False the query wouldn't work anyway?

backend/ng/challenge/tests/test_challenge_tag_model.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def test_challenge_tag_repr_should_return_correct_format(self, challenge):
+        """Test that __repr__ returns the expected format."""
+        tag = ChallengeTag.create_tag(challenge_id=challenge.id, name="forensics", commit=True)
+
+        repr_str = repr(tag)
+        assert f"<NgChallengeTag {tag.id}, name=forensics>" == repr_str
+
```
RayofLightz: Testing the repr format seems strange. Like this is kinda just cluttering up the test file

backend/ng/event/tests/test_event_validation.py
```diff
@@ -17,8 +19,8 @@ def test_event_time_constraint_validation(self):
         """Test that start_time < end_time constraint is understood."""
 
         now = utc_now()
-        future_start = (now + timedelta(hours=1)).isoformat()
-        future_end = (now + timedelta(hours=2)).isoformat()
+        future_start = (now + timedelta(hours=1)).isoformat() + "Z"
+        future_end = (now + timedelta(hours=2)).isoformat() + "Z"
```
RayofLightz: I thought we had a lib for date formatting because of the dumb issue lol

backend/ng/team/models/Team.py
```diff
@@ -26,15 +29,18 @@ class SerializedTeam(TypedDict):
     member_count: int
     ranked: bool
     locked: bool
+    seed: str
```
RayofLightz: We might want this optional so players can't see the seed.

backend/ng/team/models/Team.py
```diff
@@ -124,20 +139,32 @@ def self_validate(self) -> None:
...
+        return "".join(random.choices(HEX_CHARS, k=SEED_LENGTH))
 
     @classmethod
-    def create_team(cls, name : str, event_id : int, invite_code : str | None = None, ranked : bool = True, commit : bool = True):
+    def create_team(
+        cls,
+        name: str,
+        event_id: int,
+        invite_code: str | None = None,
+        seed: str | None = None,
```
RayofLightz, line 160: What are we using seed for? I don't see it used anywhere other than being stored at creation.

RayofLightz (changes requested):
TBH I only really see two things being actual issues. The dict indexing in create_challenge and the default cpu. Most of the other things are more or less questions.

backend/ng/challenge/models/ChallengeTag.py
fpayer: Nope - open to ideas though.

backend/ng/challenge/models/ChallengeTag.py
fpayer: I was getting None `id` prior to adding in the flush. I'm not sure if CTFd is doing something or if the autoflush runs on a timer, but the explicit flush was necessary to ensure that the `id` was generated.

backend/ng/team/models/Team.py
fpayer: This is just an initial step before I start implementing the answer validation functionality. I'll be doing that after this PR is merged in.

backend/ng/challenge/tests/test_challenge_tag_model.py
fpayer: Removed

backend/ng/challenge/tests/test_challenge_tag_model.py
fpayer: I've updated to `commit=True`.

However, SQLalchemy is weird. Even if you don't commit, the uncommitted data will still be available in-memory and will be returned in queries made from the same session. It won't be visible for queries from other sessions though.

backend/ng/challenge/models/ContainerBlueprint.py
fpayer: What's a better default?

Also, why remove that ability from challenge devs? What if there is an image that is more CPU-intensive?

backend/ng/challenge/models/ContainerBlueprint.py
fpayer: This was the flow that Carter set up. I also think it's a bit weird with the tuple return, but it's less verbose than unpacking a dict. Open to other solutions.

backend/ng/challenge/controllers/import_challenge_from_yaml.py
fpayer: Done

backend/ng/challenge/models/ContainerBlueprint.py
fpayer: Actually, I'm just going to remove the default and leave it None since the field is nullable

backend/ng/event/tests/test_event_validation.py
fpayer: No, the solution was that the front-end would always send UTC ISO strings. I can make a util for it though.

backend/ng/challenge/models/ChallengeTag.py
RayofLightz: Gotcha

backend/ng/challenge/models/ChallengeTag.py
RayofLightz: My only thing is minimum id

backend/ng/challenge/models/ContainerBlueprint.py
RayofLightz: Errors should be a throw not a return. Is valid is redundant imo because it either returns your data or errors

backend/ng/challenge/models/ContainerBlueprint.py
RayofLightz: The reason for it is that it is relative to the count of cpu's. We don't want a container that is going to have like potentially hundreds of instances and each one is allowed to potentially fight for a whole cpu.

backend/ng/challenge/tests/test_challenge_tag_model.py
RayofLightz: Thats really really weird lol

RayofLightz (approved):
Looks good to me

backend/ng/challenge/models/ContainerBlueprint.py
fpayer: Let me check if there's anywhere in the code that actually has any real branching logic based on if there is validation errors.

Most of my usage of validator has been `if not is_valid`, then throw. If all of the callers are just throwing an error if the validation fails, then it would be cleaner to just throw inside the validator.

PR #77, opened 2025-07-16
Ctfd ext team management changes

droge91:
- Added new admin and user routes for team management
- Simplified permissions checks to one middleware and included permission checks for all team management routes
- Added additional logic to existing team management functionality


backend/ng/event/controllers/user/join_event_controller.py
```diff
@@ -31,5 +33,6 @@ def join_event_controller(event: Event, user : User, invite_code : str = "", tea
         db.session.commit()
         return team
     except Exception as e:
+        print(e)
```
fpayer: remove debug

backend/ng/event/controllers/user/join_event_controller.py
```diff
@@ -11,6 +11,8 @@ def join_event_controller(event: Event, user : User, invite_code : str = "", tea
     """
     if (not invite_code) and (not team_name):
         raise ValidationError("Either invite_code or team_name must be provided")
+
+    Event.check_eligibility(event, user)
```
fpayer: The eligibility check should happen in the route.

backend/ng/event/routes/user_routes.py
```diff
@@ -31,39 +42,51 @@ def get(self):
 class EventDetail(Resource):
     @user_endpoint()
     @load_event(source=LoaderType.PARAM)
-    def get(self, event_id, event):
+    def get(self, event_id, event, **kwargs):
         """Get event details"""
+        if not event.public:
```
fpayer: The `if not event.public` should probably use a permission middleware since the logic is reused in multiple places. 

And also admins should be able to bypass the "public" restriction.

backend/ng/event/routes/user_routes.py
```diff
@@ -31,39 +42,51 @@ def get(self):
...
+        
         return success_response(True)
 
 
 @events_user_namespace.route("/<int:event_id>/me/register")
 class EventRegistration(Resource):
     @user_endpoint(json_required=True)
     @load_event(source=LoaderType.PARAM)
-    def post(self, event_id: str, current_user: User, json_data, event):
+    def post(self, event_id, current_user, json_data, event):
```
fpayer: Any reason for removing the typing?

backend/ng/event/routes/user_routes.py
```diff
@@ -94,5 +117,68 @@ class EventTeamMembers(Resource):
...
+@events_user_namespace.route("/<int:event_id>/me/team/kick")
+class EventTeamKick(Resource):
+    @user_endpoint(json_required=True)
+    @load_event(source=LoaderType.PARAM)
+    @load_user(source=LoaderType.BODY)
+    @load_team_by_user_and_event()
+    @get_permissions
+    def post(self, event_id, **kwargs):
+        """Kick a user from the user's team in the event"""
+        json_data = kwargs.get("json_data", {})
```
fpayer: The "get"s are not necessary since the middleware will throw an error before this function if the values are empty / invalid. Same applies to the below routes

backend/ng/event/routes/user_routes.py
```diff
@@ -94,5 +117,68 @@ class EventTeamMembers(Resource):
...
+        user_id = json_data.get("user_id")
+        team = kwargs.get("team")
+        permissions = kwargs.get("permissions", [])
+        if user_id == kwargs.get("current_user").id:
+            return error_response("You cannot kick yourself from the team.", "validation", 400)
+        if PermissionEnum.CAN_EDIT_TEAM not in permissions:
+            return error_response("You do not have permission to kick team members", "forbidden", 403)
+
+
+        team.remove_member_and_regenerate_code(user_id)
```
fpayer: After removing the `TeamMember` entry, the user's demographic entry must also be removed or they will not be able to re-register.

backend/ng/event/routes/user_routes.py
```diff
@@ -94,5 +117,68 @@ class EventTeamMembers(Resource):
...
+    @user_endpoint()
+    @load_event(source=LoaderType.PARAM)
+    @load_team_by_user_and_event()
+    def get(self, event_id, current_user,**kwargs):
+        """Leave the user's team in the event"""
+        team = kwargs.get("team")
+        team_member = TeamMember.find_by_user_and_team(current_user.id, team.id)
+        if team_member.role == TeamRole.CAPTAIN:
+            return error_response("You cannot leave the team as a captain. Please promote another member first.", "forbidden", 403)
+        team_member.remove_team_member(commit=True)
```
fpayer: After removing the `TeamMember` entry, the user's demographic entry must also be removed or they will not be able to re-register.

backend/ng/team/models/Team.py
```diff
@@ -453,23 +464,26 @@ def remove_member_and_regenerate_code(self, user_id: int, commit=True) -> None:
 
         try:
             member = TeamMember.find_by_user_and_event(user_id=user_id, event_id=self.event_id)
```
fpayer: Just realized there's a security vuln in the logic here. Instead of calling `find_by_user_and_event`, there should be a direct query for the team member because we also need to check that `member.team == self.id`. Otherwise, it would be possible for any team captain to kick another person off someone else's team.

backend/ng/team/models/Team.py
```diff
@@ -453,23 +464,26 @@ def remove_member_and_regenerate_code(self, user_id: int, commit=True) -> None:
...
+            if not member:
+                raise ValidationError(f"User {user_id} is not a member of team {self.id}.")
             if member:
                 member.remove_team_member(commit=False)
                 self.update_invite_code(commit=False)
                 if commit:
                     db.session.commit()
+                return True
```
fpayer: Shouldn't have a return val

backend/ng/team/routes/admin_routes.py
```diff
@@ -25,16 +26,69 @@ def get(self, **kwargs):
...
         """Get a team"""
         return success_response(team)
 
+    @admin_endpoint(json_required=True, validation_func=Team.validate)
+    @load_team(source=LoaderType.PARAM)
+    def patch(self, team_id, **kwargs):
+        """Update a team"""
+        json_data = kwargs.get("validated_data", None)
+        if not json_data:
+            return success_response({"message": "Invalid data"}, status_code=400)
```
fpayer: `json_required=True` in the `admin_endpoint` decorator means that this check is already performed for you.

backend/ng/team/routes/admin_routes.py
```diff
@@ -25,16 +26,69 @@ def get(self, **kwargs):
...
 @teams_admin_namespace.route("/<int:team_id>/members")
 class TeamMembers(Resource):
     @admin_endpoint()
     @load_team(source=LoaderType.PARAM)
-    def get(self, team_id, team):
+    def get(self, team_id, team, **kwargs):
         """Get all members of a team"""
-        return success_response(team.members)
\ No newline at end of file
+        return success_response(team.members)
```
fpayer, line 61: Some inconsistency here. The user version of this route calls `team.get_full_team_details` while this admin version returns `team.members` directly. I'm not sure which is correct, but it should be the same for both.

backend/ng/team/routes/admin_routes.py
```diff
@@ -25,16 +26,69 @@ def get(self, **kwargs):
...
+class TeamKick(Resource):
+    @admin_endpoint(json_required=True, validation_func=User.validate)
+    @load_team(source=LoaderType.PARAM)
+    def post(self, team_id, **kwargs):
+        """Kick a user from a team"""
+        user_id = kwargs.get("validated_data", {}).get("user_id")
+        team = kwargs.get("team")
+
+        success = team.remove_member_and_regenerate_code(user_id)
+        if not success:
```
fpayer: No need for this check. If there's an error in the function, the error will bubble up.

backend/ng/team/routes/admin_routes.py
```diff
@@ -25,16 +26,69 @@ def get(self, **kwargs):
...
+        return success_response(team.members)
+
+
+@teams_admin_namespace.route("/<int:team_id>/kick")
+class TeamKick(Resource):
+    @admin_endpoint(json_required=True, validation_func=User.validate)
+    @load_team(source=LoaderType.PARAM)
+    def post(self, team_id, **kwargs):
+        """Kick a user from a team"""
+        user_id = kwargs.get("validated_data", {}).get("user_id")
```
fpayer: You can use the `load_user` decorator with `LoaderType.BODY` to look up the `user_id` from the JSON body. That will also throw an error for you if the `user_id` is invalid.

backend/ng/team/routes/admin_routes.py
```diff
@@ -25,16 +26,69 @@ def get(self, **kwargs):
...
+            return error_response({"message": "Kick failed"}, status_code=400)
+        return success_response()
+
+@teams_admin_namespace.route("/<int:team_id>/promote")
+class TeamPromote(Resource):
+    @admin_endpoint(json_required=True)
+    @load_team(source=LoaderType.PARAM)
+    def post(self, team_id, **kwargs):
+        """Promote a user to team leader"""
+        user_id = kwargs.get("validated_data", {}).get("user_id")
```
fpayer: Same here with `load_user`

backend/ng/user/models/User.py
```diff
@@ -132,7 +132,7 @@ def find_by_id(cls, user_id: int):
         return cls.query.get(user_id)
 
     @classmethod
-    def find_or_create_by_ctfd_id(cls, ctfd_user_id: int, commit: bool = True) -> User:
+    def find_or_create_by_ctfd_id(cls, ctfd_user_id: int, commit: bool = True):
```
fpayer: Any reason for removing the typing?

fpayer (changes requested):
Is there a reason why you do `kwargs.get` instead of defining the kwarg in the function signature? If it's for null-safety, you don't have to worry about that if the middleware is set up properly because the middleware will error if data/ids are missing or invalid. 

The logic for the "is a team member's name in the team name" is repeated. Should make that logic into its own function to avoid issues with the logic getting out of sync between locations. It should also be a little more comprehensive with checking the name. 

With the current logic, if someone's name is "Alex John Smith" and they try to set their team name to "Alex Smith", that would be allowed. On the flip side, if someone's name is "Alex Ng" and they want a team name of "pwning", that would be unnecessarily blocked.

I'm thinking of maybe doing a split on " " in the name and then turn each word into a regex where you check for a non-word boundary on either side. That would allow a team name of "pwning" since the "ng" is part of another word, but a team name of "pwni ng" would not be allowed. That might also need to require a min length of 2 for each word or someone with the name "A Smith" wouldn't be able to have a team name of "a hacker". I'm open to other possible solutions here - I don't think we're going to find anything perfect.

backend/ng/event/routes/user_routes.py
droge91: no functional one, just seemed weird that those two params were the only ones typed out of everything.

backend/ng/event/routes/user_routes.py
droge91: yeah. I had considered modifying the existing middleware for this, but making a new one is probably the simplest way to to handle it.  Do we need admin bypass for this? since they already have their own admin endpoints without this logic present at all.

fpayer: The use-case for it would be if you are in "god mode" and the admin is logged in as a user

backend/ng/event/routes/user_routes.py
```diff
@@ -34,47 +52,67 @@ def get(self):
...
+
         """Check event eligibility"""
-        # TODO - Actually implement eligibility check
-        # event.registration_open = True
-        # current time is inbetween registration start and end dates (if set)
-        # user is not already in the event (i.e. does not have a demographic or is on a team)
 
+        try:
+            Event.check_eligibility(event, current_user)
+        except ValidationError as e:
```
fpayer: Is there a reason to catch this exception instead of bubbling up?

backend/ng/event/models/Event.py
```diff
@@ -335,3 +337,28 @@ def delete_all(cls) -> None:
         except Exception:
             db.session.rollback()
             raise
+
+    @classmethod
+    def check_eligibility(cls, event, user):
```
fpayer: I think this would make more sense as an instance method instead of a class method.

backend/ng/team/routes/admin_routes.py
```diff
@@ -25,16 +27,60 @@ def get(self, **kwargs):
...
+
+        new_name = validated_data.get("name", team.name)
+        if Team.team_name_contains_member_name(name=new_name, member_names=[m.user.ctfd_user.name for m in team.members]):
+            return error_response(
+                "Team name cannot include a member's name.",
+                "validation",
+                400,
+            )
+
+        team = team.update_name(team_id, new_name)
```
fpayer: Is this correct usage? I don't think `update_name` takes in a `team_id`

backend/ng/core/middleware/loaders/load_team_by_user_and_event.py
```diff
@@ -6,18 +6,18 @@
 from ....user.models.User import User
 from ....event.models.Event import Event
 
-def load_team_by_user_and_event(current_user = User, event = Event, output_key="team") -> Callable:
+def load_team_by_user_and_event(output_key="team") -> Callable:
```
fpayer, line 9: What is the purpose of moving out of the function signature?

backend/ng/event/routes/user_routes.py
```diff
@@ -97,15 +135,97 @@ class EventTeamMembers(Resource):
...
+        """Kick a user from the user's team in the event"""
+        user_id = json_data.get("user_id")
+        if user_id == current_user.id:
+            return error_response("You cannot kick yourself from the team.", "validation", 400)
+        if PermissionEnum.CAN_EDIT_TEAM not in permissions:
+            return error_response("You do not have permission to kick team members", "forbidden", 403)
+
+
+        team.remove_member_and_regenerate_code(user_id)
+        demographic = Demographic.find_by_user_and_event(user_id, event_id)
```
fpayer: Should wrap this in a try/catch and perform a rollback if either `remove_member_and_regnerate_code` or the `demographic.delete` fails

backend/ng/event/routes/user_routes.py
```diff
@@ -97,15 +135,97 @@ class EventTeamMembers(Resource):
...
+class EventTeamLeave(Resource):
+    @user_endpoint()
+    @load_event(source=LoaderType.PARAM)
+    @load_team_by_user_and_event()
+    def get(self, event_id, team, current_user,**kwargs):
+        """Leave the user's team in the event"""
+        team_member = TeamMember.find_by_user_and_team(current_user.id, team.id)
+        if team_member.role == TeamRole.CAPTAIN:
+            return error_response("You cannot leave the team as a captain. Please promote another member first.", "forbidden", 403)
+        team_member.remove_team_member(commit=True)
```
fpayer: Same here with the rollback

fpayer (changes requested):

backend/ng/event/models/Event.py
droge91: I had it as an instance method originally, but that causes an issue because it needs to be called in the EventRegistration route before the team object is created.

backend/ng/core/middleware/loaders/load_team_by_user_and_event.py
droge91: it doesn't function properly with them there, you never want the current_user or event to be the models themselves.

backend/ng/event/routes/user_routes.py
droge91: will add that logic with the godmode/impersonation implementation then

backend/ng/core/middleware/loaders/load_team_by_user_and_event.py
fpayer: What do you mean? They are the models themselves

fpayer: That's the distinction between `event` and `event_id`. `event` would be the actual Event model

droge91: we want event to be an Event instance not the class Event.
maybe this was meant to be current_user : User, event : Event as an enforcement of types?

fpayer: ohh, sorry. yes, you are correct. it is an instance of the model. It should be `current_user: User` and `event: Event`. 

If you put those in the function signature instead of doing the `kwargs.get`, wouldn't that throw an error if they are None? Then you could skip the `if not current_user or not event` check

droge91: because the values for current_user and event are injected kwargs from previous middlewares, it'll throw an error if you include them in the function signature without a default value, since we aren't giving it those values in the load_team_by_user_and_event() call itself. Them defaulting to the model class was causing that error not to happen but breaking its functionality in other ways.


I guess we can have them default to None if we really want them in the signature?

fpayer: Nah, all good then. I was just getting confused by kwargs

fpayer (approved):

PR #79, opened 2025-07-18
Challenge endpoints

fpayer:
- YAML upload now works. Also moved the route to /admin/events/
- Created new endpoints for getting challenge-related info
- Updated the populate data script to now import the default challenge test yaml


trevnels-cs (approved):

trevnels-cs (commented):
One thing to consider, do we want an admin equivalent for `/event/challenges` to list challenges on the admin side? I assume the user-facing endpoint will eventually be locked down until they've started the event, so admins need some way around that.

backend/ng/challenge/models/Hint.py
```diff
@@ -75,3 +89,14 @@ def create_hint(cls, challenge_id: int, body: str, preview: str = "", deduction:
...
+    def as_attr(self) -> HintAttr:
+        """
+        Convert the Hint model to a HintAttr object.
+        :return: A HintAttr object representing the hint.
+        """
+        return HintAttr(
+            body=self.body,  # type: ignore
+            preview=self.preview,  # type: ignore
+            deduction=self.deduction,  # type: ignore
+        )
```
RayofLightz, line 102: I am assuming this is whay we are going to send to the end user. Maybe make a typed dict just because I don't know how well Attr's serialize

backend/ng/event/routes/admin_routes.py
```diff
@@ -45,3 +42,20 @@ def put(self, event_id, event, validated_data):
...
+            200: "Success",
+            400: "Bad request",
+        },
+    )
+    @admin_endpoint(json_required=True)
+    @load_event(source=LoaderType.PARAM)
+    def post(self, event_id, event, json_data):
+        """Create a challenge for an event"""
+        challenge = import_challenge_from_yaml(event, json_data)
+        return success_response(challenge)
```
RayofLightz, line 61: minor nit. The route on this seems kinda ambiguous. You are just posting directly to challenges. Its unclear to the callee what the route is doing from reading it. I would include import in the url so the callee on the front end at least sees some kind of verb as to what the route is going to do.

RayofLightz (commented):
Minor things.

backend/ng/event/routes/admin_routes.py
trevnels-cs: I think a POST to challenges makes sense to represent this, as this operation is what creates a new challenge in the system.

backend/ng/challenge/models/Hint.py
fpayer: This isn't being sent to the end user - that is still `.serialize`. `.as_attr` is to convert the db models back into John's parser classes so that I can use the parser's existing functions to render the template values.

backend/ng/event/routes/admin_routes.py
fpayer: I'm just gonna leave as-is for now and we can always come back to change it. I just don't want making this decision to slow us down.

PR #81, opened 2025-07-18
Ng score

CarterPerez-dev:
# Scoring Implementation

## Clarifications:
- Copied over support tests and cleaned up models from previous cleanup, was too much behind of developement so was easier just to copy them. 
- Copied over new Team model and route file from Daniels PR because I needed to edit the update team method anyway and wanted to prevent more complex merge conflicts in when his is merged.
- Looks like the tool.ruff.lint pyproject config possibly needs some edits becasue ruff --fix (non breakable) removed all white space and updated older python imports based on python 3.11+, could be purposeful, but just an fwi.


backend/ng/core/utils/validator.py
```diff
@@ -324,3 +327,10 @@ def validate_time_window(
                 f"{end_field.replace('_', ' ').title()} must be after {start_field.replace('_', ' ')}."
             )
             return
+
+    def validate_optional_timestamp(self, data: dict[str, Any], field: str = "timestamp", allow_past: bool = True) -> None:
```
fpayer: Why not just use `validate_datetime` with `required = False`?

backend/ng/scoring/models/ManualPointAward.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        Validate manual point award data
+        """
+        validator = BaseValidator()
+
+        if "admin_id" not in data:
+            validator.errors["admin_id"] = "Admin ID is required"
+        else:
+            admin = Users.query.get(data["admin_id"])
+            if not admin:
```
fpayer: Would be better to have a `validate_admin_id` function on the validator itself. Since we'd likely need to re-use that functionality.

backend/ng/scoring/models/ManualPointAward.py
```diff
@@ -0,0 +1,237 @@
...
+            validator.errors["points"] = "Points value is required"
+        else:
+            try:
+                points = int(data["points"])
+                if points == 0:
+                    validator.errors["points"] = "Points cannot be zero"
+                else:
+                    validator._add_parsed_data("points", points)
+            except (ValueError, TypeError):
+                validator.errors["points"] = "Points must be a valid integer"
```
fpayer: We should really have a generic `validate_integer` function with validator

backend/ng/scoring/models/ManualPointAward.py
```diff
@@ -0,0 +1,237 @@
...
+        validator.validate_string(
+            data, "reason", max_length=config.MANUAL_AWARD_REASON_MAX_LENGTH, required=True, friendly_name="Reason"
+        )
+
+        validator.validate_optional_timestamp(data)
+
+        return validator.validate()
+
+    @classmethod
+    def validate_api_award(cls, data: dict[str, Any]) -> dict[str, Any]:
```
fpayer: Why is `validate_api_award` necessary? Why not just use the regular `validate`?

backend/ng/scoring/controllers/admin_actions/award_manual_points.py
```diff
@@ -0,0 +1,40 @@
...
+    award = ManualPointAward.create_award(
+        admin_id=admin_id,
+        team_id=team_id,
+        points=points,
+        reason=reason,
+        event_id=event_id,
+    )
+
+    return {
+        "award": award,
```
fpayer: No need to return anything other than just the `award` object

backend/ng/scoring/controllers/admin_actions/award_manual_points.py
```diff
@@ -0,0 +1,40 @@
...
+    event_id: int,
+    team_id: int,
+    points: int,
+    reason: str,
+    admin_id: int,
+) -> dict[str, Any]:
+    """
+    Award manual points to a team (can be positive or negative)
+    """
+    score = Score.find_by_team_and_event(team_id=team_id, event_id=event_id)
```
fpayer: Don't need this lookup/check for `score`. That gets handled by the `create_award` function

backend/ng/scoring/controllers/admin_actions/get_score_history.py
```diff
@@ -0,0 +1,33 @@
...
+
+from typing import Any
+
+from ...models import ScoreEvent
+
+
+def get_score_history(
+    event_id: int,
+    team_id: int | None = None,
+    limit: int = 50,
```
fpayer: This should defined as a constant and be set as the default within the `ScoreEvent.find_filtered_events` function

backend/ng/scoring/controllers/admin_actions/get_score_history.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    events = ScoreEvent.find_filtered_events(
+        event_id=event_id,
+        team_id=team_id,
+        limit=limit,
+        eager_load_source=True,
+    )
+
+    return {
+        "total_events": len(events),
```
fpayer: Just return the array of `events`. The rest is unncessary.

backend/ng/scoring/controllers/admin_actions/recalculate_score.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def recalculate_score(
+    event_id: int,
+    team_id: int,
+) -> dict[str, Any]:
+    """
+    Recalculate a team's score from all ScoreEvents
+    """
+    score = Score.find_by_team_and_event(team_id=team_id, event_id=event_id)
```
fpayer: Should perform this lookup in the middleware instead of in the controller. Might need to creat a new middleware for this.

backend/ng/scoring/controllers/admin_actions/recalculate_score.py
```diff
@@ -0,0 +1,33 @@
...
+    if not score:
+        raise NotFoundError(f"Team {team_id} has no score in event {event_id}")
+
+    old_points = score.points
+
+    score.recalculate()
+
+    return {
```
fpayer: Just return the updated score object. The rest is unnecessary.

backend/ng/scoring/controllers/all_actions/get_leaderboard.py
```diff
@@ -0,0 +1,15 @@
...
+from typing import Any
+
+from ...models import Score
+
+
+def get_leaderboard(event_id: int, limit: int | None = 100) -> list[dict[str, Any]]:
+    """
+    Retrieves the cached leaderboard for a given event
+    """
+    leaderboard_data = Score.get_leaderboard(event_id=event_id, limit=limit)
```
fpayer, line 15: Default limit should be a constant and default within `Score.get_leaderboard`

backend/ng/scoring/controllers/all_actions/get_team_score.py
```diff
@@ -0,0 +1,28 @@
...
+Retrieves the score details for a specific team.
+"""
+
+from typing import Any
+from ....core.exceptions import NotFoundError
+
+from ...models import ScoreEvent, Score
+
+
+def get_team_score(event_id: int, team_id: int, include_history: bool = False) -> dict[str, Any]:
```
fpayer: Remove `include_history` and `rank`. This should just return score.

backend/ng/scoring/controllers/all_actions/get_team_score.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Get the current score and optionally the scoring history for a team
+    """
+    score = Score.find_by_team_and_event(team_id=team_id, event_id=event_id)
```
fpayer: Move to middleware

backend/ng/scoring/controllers/user_actions/redeem_hint.py
```diff
@@ -0,0 +1,40 @@
...
+def redeem_hint(
+    event_id: int,
+    challenge_id: int,
+    hint_id: int,
+    current_user_id: int,
+) -> dict[str, Any]:
+    """
+    Redeem a hint for a challenge
+    """
+    team = Team.find_by_user_and_event(current_user_id, event_id)
```
fpayer: Move to middleware

backend/ng/scoring/controllers/user_actions/submit_answer.py
```diff
@@ -0,0 +1,40 @@
...
+    event_id: int,
+    challenge_id: int,
+    question_id: int,
+    submission: str,
+    current_user_id: int,
+) -> dict:
+    """
+    Submit an answer to a challenge question
+    """
+    team = Team.find_by_user_and_event(current_user_id, event_id)
```
fpayer: Move to middleware

backend/ng/scoring/controllers/user_actions/submit_answer.py
```diff
@@ -0,0 +1,40 @@
...
+        team_id=team.id,
+        event_id=event_id,
+        challenge_id=challenge_id,
+        question_id=question_id,
+        submission=submission,
+    )
+
+    score = Score.find_by_team_and_event(team.id, event_id)
+
+    return {
```
fpayer: Just return the attempt, nothing else

backend/ng/scoring/models/Attempt.py
```diff
@@ -0,0 +1,297 @@
...
+        validator.validate_model_id(data, "challenge_id", "Challenge", required=True)
+        validator.validate_model_id(data, "question_id", "Question", required=True)
+
+        validator.validate_string(
+            data, "submission", max_length=config.MAX_SUBMISSION_LENGTH, required=True, friendly_name="Submission"
+        )
+
+        validator.validate_optional_timestamp(data)
+
+        if "points" in data:
```
fpayer: Why not just use `validate_positive_integer`?

backend/ng/scoring/models/Attempt.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+        cls.validate_attempt_allowed(user_id, team_id, event_id, challenge_id, question_id)
+
+        if timestamp is None:
+            timestamp = utc_now()
+
+        # LAZY-IMPORT
+        from ...challenge.models.Question import Question
+
+        question = Question.query.get(question_id)
```
fpayer: The existance of question is alreadt checked already checked in `.validate`. Just move validate up so you don't need this duplicitive check

backend/ng/team/models/Team.py
```diff
@@ -145,6 +147,21 @@ def self_validate(self) -> None:
         except ValidationError as e:
             raise ValidationError(f"Team validation failed: {e.errors}") from e
 
+    @classmethod
+    def team_name_contains_member_name(cls,name, member_names) -> bool:
```
fpayer, line 151: Looks like this is a dup function from an incorrect rebase/merge

backend/ng/support/models/TicketTag.py
```diff
@@ -7,25 +7,80 @@
...
+
+        if "name" in data and "name" not in validator.errors:
+            existing = cls.query.filter_by(name=data["name"]).first()
+            if existing:
+                if not current_instance or existing.id != current_instance.id:
+                    validator.errors["name"] = (
+                        f"Tag name '{data['name']}' already exists"
+                    )
+
+        is_valid, errors, parsed_data = validator.is_valid()
```
fpayer: `validator.is_valid` no longer exists. Switch to `validator.validate`

backend/ng/support/models/TicketTag.py
```diff
@@ -59,37 +115,74 @@ def create(
         Returns:
             TicketTag: The created tag instance
         """
-        tag = cls(name=name, color=color, description=description)
+
+        data = {
+            "name": name,
+        }
+
+        if color is not None:
```
fpayer: No need for these `if color is not None` and `if description is not None`. This is already being checked by `.validate`

backend/ng/support/models/TicketMessage.py
```diff
@@ -5,57 +5,106 @@
...
+            friendly_name="Ticket ID",
+        )
+        validator.validate_positive_integer(
+            data,
+            "author_id",
+            required=True,
+            friendly_name="Author ID",
+        )
+
+        is_valid, errors, parsed_data = validator.is_valid()
```
fpayer: Replace `validator.is_valid` with the new `validator.validate`

backend/ng/support/models/TicketMessage.py
```diff
@@ -5,57 +5,106 @@
...
+        validator = BaseValidator()
+
+        validator.validate_string(
+            data,
+            "text",
+            config.TICKET_MESSAGE_MAX_LENGTH,
+            required=True,
+            friendly_name="Message text",
+        )
+        validator.validate_positive_integer(
```
fpayer: use `validator.validate_model_id`

backend/ng/support/models/TicketMessage.py
```diff
@@ -5,57 +5,106 @@
...
+            data,
+            "ticket_id",
+            required=True,
+            friendly_name="Ticket ID",
+        )
+        validator.validate_positive_integer(
```
fpayer: use `validator.validate_model_id`

backend/ng/support/models/TicketMessage.py
```diff
@@ -66,7 +115,32 @@ def create(cls, text: str, ticket_id: int, author_id: int, commit: bool = True)
...
+            "text": text,
+            "ticket_id": ticket_id,
+            "author_id": author_id,
+        }
+
+        validated_data = cls.validate(data)
+        # LAZY-IMPORT: Tagging all necessary lazy imports for easy searchability & visibility.
+        from .Ticket import Ticket
+
+        if not Ticket.find_by_id(validated_data["ticket_id"]):
```
fpayer: validate id in the `validate` function

backend/ng/support/models/TicketMessage.py
```diff
@@ -66,7 +115,32 @@ def create(cls, text: str, ticket_id: int, author_id: int, commit: bool = True)
...
+            raise NotFoundError(
+                f"Ticket with ID {validated_data['ticket_id']} not found"
+            )
+
+        if not Users.query.filter_by(id=validated_data["author_id"]).first():
```
fpayer: validate id in the `validate` function

backend/ng/support/models/Ticket.py
```diff
@@ -38,18 +53,95 @@ class Ticket(db.Model):
...
+        validator = BaseValidator()
+
+        validator.validate_string(
+            data,
+            "subject",
+            config.TICKET_SUBJECT_MAX_LENGTH,
+            required=True,
+            friendly_name="Ticket subject",
+        )
+        validator.validate_positive_integer(
```
fpayer: use `validate_model_id`

backend/ng/support/models/Ticket.py
```diff
@@ -38,18 +53,95 @@ class Ticket(db.Model):
...
+            data,
+            "author_id",
+            required=True,
+            friendly_name="Author ID",
+        )
+
+        validator.validate_positive_integer(
```
fpayer: use `validate_model_id`

backend/ng/support/models/Ticket.py
```diff
@@ -38,18 +53,95 @@ class Ticket(db.Model):
...
+            data,
+            "event_id",
+            required=False,
+            friendly_name="Event ID",
+        )
+        validator.validate_positive_integer(
```
fpayer: use `validate_model_id`

backend/ng/support/models/Ticket.py
```diff
@@ -38,18 +53,95 @@ class Ticket(db.Model):
...
+            data,
+            "team_id",
+            required=False,
+            friendly_name="Team ID",
+        )
+        validator.validate_positive_integer(
```
fpayer: use `validate_model_id`

backend/ng/support/models/Ticket.py
```diff
@@ -38,18 +53,95 @@ class Ticket(db.Model):
...
+            data,
+            "challenge_id",
+            required=False,
+            friendly_name="Challenge ID",
+        )
+        validator.validate_positive_integer(
```
fpayer: use `validate_model_id`

backend/ng/support/models/Ticket.py
```diff
@@ -38,18 +53,95 @@ class Ticket(db.Model):
...
+                    if not isinstance(tag_id, int) or tag_id <= 0:
+                        validator.errors[f"tag_ids[{idx}]"] = (
+                            "Each tag ID must be a positive integer"
+                        )
+                    else:
+                        valid_tag_ids.append(tag_id)
+                if not validator.errors:
+                    validator._add_parsed_data("tag_ids", valid_tag_ids)
+
+        is_valid, errors, parsed_data = validator.is_valid()
```
fpayer: Replace `validator.is_valid` with `validator.validate`

backend/ng/support/models/Ticket.py
```diff
@@ -38,18 +53,95 @@ class Ticket(db.Model):
...
     @hybrid_property
-    def status(self):
-        """Compute ticket status based on stored fields."""
+    def status(self) -> str:
+        """
+        Compute ticket status based on stored fields.
+        """
         if self.closed_timestamp is not None:
             return "closed"
         elif self.muted:
```
fpayer: `muted` is not a status. Status is just "open" or "closed"

backend/ng/support/models/Ticket.py
```diff
@@ -80,8 +174,12 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
             "subject": self.subject,
             "author_id": self.author_id,
             "status": self.status,
-            "opened_timestamp": self.opened_timestamp.isoformat() + "Z" if self.opened_timestamp else None,
-            "last_updated": self.last_updated.isoformat() + "Z" if self.last_updated else None,
+            "opened_timestamp": self.opened_timestamp.isoformat()
```
fpayer: `+ "Z"` required?

backend/ng/support/models/Ticket.py
```diff
@@ -80,8 +174,12 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
+            if self.opened_timestamp
+            else None,
+            "last_updated": self.last_updated.isoformat()
```
fpayer: `+ "Z"` required?

backend/ng/support/models/Ticket.py
```diff
@@ -91,12 +189,16 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
 
         if include_admin_fields:
             first_admin_response_timestamp = (
-                self.first_admin_response_timestamp.isoformat() + "Z" if self.first_admin_response_timestamp else None
+                self.first_admin_response_timestamp.isoformat()
```
fpayer: `+ "Z"` required?

backend/ng/support/models/Ticket.py
```diff
@@ -91,12 +189,16 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
+                if self.first_admin_response_timestamp
+                else None
             )
             data.update(
                 {
                     "assigned_to": self.assigned_to,
-                    "closed_timestamp": self.closed_timestamp.isoformat() + "Z" if self.closed_timestamp else None,
+                    "closed_timestamp": self.closed_timestamp.isoformat()
```
fpayer: `+ "Z"` required?

backend/ng/support/models/Ticket.py
```diff
@@ -105,120 +207,230 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
         Returns:
             Ticket: The created ticket instance
         """
+
+        data = {
+            "subject": subject,
+            "author_id": author_id,
+        }
+
+        if event_id is not None:
```
fpayer: Why not just pass these directly to `.validate`? There should be None checks in `.validate`

backend/ng/support/models/Ticket.py
```diff
@@ -105,120 +207,230 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
+        if team_id is not None:
+            data["team_id"] = team_id
+        if challenge_id is not None:
+            data["challenge_id"] = challenge_id
+        if tag_ids is not None:
+            data["tag_ids"] = tag_ids
+
+        validated_data = cls.validate(data)
+
+        if not Users.query.filter_by(id=validated_data["author_id"]).first():
```
fpayer: Validate the model ids in the `.validate` function

backend/ng/support/models/Ticket.py
```diff
@@ -105,120 +207,230 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
+            subject=validated_data["subject"],
+            author_id=validated_data["author_id"],
+            event_id=validated_data.get("event_id"),
+            team_id=validated_data.get("team_id"),
+            challenge_id=validated_data.get("challenge_id"),
         )
 
-        if tags:
-            ticket.tags.extend(tags)
+        if validated_data.get("tag_ids"):
```
fpayer, line 229: Validate the model ids in the .validate function

backend/ng/support/models/Ticket.py
```diff
@@ -105,120 +207,230 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
+            validator.validate_string(
+                update_data, "subject", config.TICKET_SUBJECT_MAX_LENGTH
+            )
+
+        if "event_id" in update_data and update_data["event_id"] is not None:
+            if not Event.find_by_id(update_data["event_id"]):
+                raise NotFoundError(
+                    f"Event with ID {update_data['event_id']} not found"
+                )
+            validator.validate_positive_integer(update_data, "event_id")
```
fpayer: Use `validate_model_id`

backend/ng/support/models/Ticket.py
```diff
@@ -105,120 +207,230 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
+
+        if "team_id" in update_data and update_data["team_id"] is not None:
+            if not Team.find_by_id(update_data["team_id"]):
+                raise NotFoundError(f"Team with ID {update_data['team_id']} not found")
+            validator.validate_positive_integer(update_data, "team_id")
```
fpayer: Use `validate_model_id`

backend/ng/support/models/Ticket.py
```diff
@@ -105,120 +207,230 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
+
+        if "assigned_to" in update_data and update_data["assigned_to"] is not None:
+            if not Users.query.get(update_data["assigned_to"]):
+                raise NotFoundError(
+                    f"User to assign with ID {update_data['assigned_to']} not found"
+                )
+            validator.validate_positive_integer(update_data, "assigned_to")
```
fpayer: Use `validate_model_id`

backend/ng/support/models/Ticket.py
```diff
@@ -105,120 +207,230 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
+
+        if "muted" in update_data:
+            validator.validate_boolean(update_data, "muted")
+
+        is_valid, errors, parsed_data = validator.is_valid()
```
fpayer: Use `validator.validate`

backend/ng/scoring/routes/user_routes.py
```diff
@@ -0,0 +1,102 @@
...
+    @scoring_user_namespace.doc(**SUBMIT_ANSWER_DOC)
+    @user_endpoint(json_required=True, validation_func=Attempt.validate_api_submission)
+    @load_event(LoaderType.PARAM)
+    def post(
+        self, event_id: int, challenge_id: int, question_id: int, event, current_user: User, validated_data, **kwargs
+    ):
+        """
+        Submit an answer to a question
+        """
+        result = submit_answer(
```
fpayer, line 96: Use the loaders to pass the model instances themselves instead of passing ids.

backend/ng/scoring/routes/user_routes.py
```diff
@@ -0,0 +1,102 @@
...
+@scoring_user_namespace.route("/<int:event_id>/challenges/<int:challenge_id>/hint/<int:hint_id>/redeem")
+class RedeemHint(Resource):
+    @scoring_user_namespace.doc(**REDEEM_HINT_DOC)
+    @user_endpoint(json_required=False)
+    @load_event(LoaderType.PARAM)
+    def post(self, event_id: int, challenge_id: int, hint_id: int, event, current_user: User, **kwargs):
+        """
+        Redeem a hint
+        """
+        result = redeem_hint(
```
fpayer, line 130: Use the loaders to pass the model instances themselves instead of passing ids.

backend/ng/scoring/models/ScoreEvent.py
```diff
@@ -0,0 +1,191 @@
...
+    def validate(cls, data: dict[str, Any]) -> dict[str, Any]:
+        """
+        Validate score event data
+        """
+        validator = BaseValidator()
+
+        validator.validate_model_id(data, "score_id", "Score", required=True)
+        validator.validate_model_id(data, "team_id", "Team", required=True)
+
+        if "points" not in data:
```
fpayer: Why not use `validate_positive_integer`?

backend/ng/scoring/models/ScoreEvent.py
```diff
@@ -0,0 +1,191 @@
...
+            timestamp=timestamp,
+        )
+
+        db.session.add(event)
+        db.session.flush()
+
+        # LAZY-IMPORT: Tagging all necessary lazy imports for easy searchability & visibility.
+        from .Score import Score
+
+        score = Score.query.get(score_id)
```
fpayer: `.validate` already checks if the score_id is valid.

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,264 @@
...
+    def validate(cls, data: dict[str, Any]) -> dict[str, Any]:
+        """
+        Validate score data.
+        """
+        validator = BaseValidator()
+
+        validator.validate_model_id(data, "team_id", "Team", required=True)
+        validator.validate_model_id(data, "event_id", "Event", required=True)
+
+        if "points" in data:
```
fpayer: Why not use `validate_positive_integer`?

backend/ng/team/models/Team.py
```diff
@@ -223,13 +245,19 @@ def update_invite_code(self, new_code: str | None = None, commit=True):
...
+    def update_name(self, new_name: str, commit: bool = True) -> None:
+        """
+        Updates the team's name and synchronizes the cached name
+        """
+        # LAZY-IMPORT
+        from ...scoring.models.Score import Score
+
+        # Delegate's cache update to the Score model
         self.name = new_name
+        Score.update_team_name(team_id=self.id, new_name=new_name, commit=False)
```
fpayer: Why is commit set to false here?

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,264 @@
...
+        query = cls.query.filter_by(event_id=event_id).order_by(cls.points.desc())
+
+        if limit:
+            query = query.limit(limit)
+
+        scores = query.all()
+        return [score.serialize() for score in scores]
+
+    @classmethod
+    def update_team_name(cls, team_id: int, new_name: str, commit: bool = True) -> None:
```
fpayer: I can't really think of a good reason why this function would be called by itself vs called via `team.update_name`. Might be better to just move this logic into that function directly unless there is a benefit to splitting this logic up into two separate functions.

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,264 @@
...
+        score = cls.find_by_team_and_event(team_id, event_id)
+        if not score:
+            return None
+
+        higher_scores = cls.query.filter(cls.event_id == event_id, cls.points > score.points).count()
+
+        return higher_scores + 1
+
+    @classmethod
+    def find_by_team_and_event(cls, team_id: int, event_id: int) -> Score | None:
```
fpayer: Could create the score object if the `team_id` and `event_id` are both valid, but a score does not exist. That would reduce complexity for what to do if the Score is None

backend/ng/scoring/models/HintRedemption.py
```diff
@@ -0,0 +1,257 @@
...
+        Validate hint redemption data.
+        """
+        validator = BaseValidator()
+
+        validator.validate_model_id(data, "hint_id", "Hint", required=True)
+        validator.validate_model_id(data, "user_id", "User", required=True)
+        validator.validate_model_id(data, "team_id", "Team", required=True)
+
+        if "points" not in data:
+            validator.errors["points"] = "Points value is required"
```
fpayer: Should have some sort of `validate_integer` function

backend/ng/scoring/models/HintRedemption.py
```diff
@@ -0,0 +1,257 @@
...
+                    validator._add_parsed_data("points", points)
+            except (ValueError, TypeError):
+                validator.errors["points"] = "Points must be a valid integer"
+
+        validator.validate_optional_timestamp(data)
+
+        return validator.validate()
+
+    @classmethod
+    def validate_redemption_allowed(
```
fpayer, line 86: At some point, most of this logic should be moved to use Daniel's permission check system.

fpayer (changes requested):
The scoring and support ticket-related changes should have been two separate PRs.

Looks like some code needs to be updated to reflect the new validator usage. 

Also, avoid returning extra stuff in your functions. The additional stuff increases complexity and the extra fields will likely be unused anyway.

backend/ng/scoring/models/ManualPointAward.py
CarterPerez-dev: I think its necessary because the decorator's validation_func only receives the JSON body. The full validate() method needs fields like admin_id and team_id, which are from the user's session and the URL, not the JSON. So, using the full validate in the decorator would cause a validation error for missing fields. Is there a better way to go about this?

backend/ng/team/models/Team.py
CarterPerez-dev: yea

backend/ng/support/models/TicketTag.py
CarterPerez-dev: Yea this was pre validation refactor, relates to how it should've been a seperate PR, could have done Score then moved onto updating these models as well, along with controllers/routes, rather than copy these over from the initial model refactor.

backend/ng/support/models/Ticket.py
CarterPerez-dev: Yes, I dont recall why I removed it...

backend/ng/scoring/models/Score.py
CarterPerez-dev: Yea you're right, I cant think of any reasons to keep them split.

backend/ng/scoring/models/HintRedemption.py
CarterPerez-dev: I was thinking the same, I'll put a TODO comment there.

backend/ng/scoring/models/ManualPointAward.py
fpayer: You could actually not specify a `validation_func` on the decorator. 

The controller is going to call `ManualPointAward.create_award`, which will call `ManualPointAward.validate` anyway. So the `validate_api_award` is kinda unnecessary since it doesn't do any validation that isn't already being done in the regular `validate` function

backend/ng/core/middleware/loaders/load_score.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Load score by team and event from the request context.
+    """
+    def decorator(f):
+        @wraps(f)
+        def decorated_function(*args, **kwargs):
+            check_output_exists(kwargs, output_key)
+            Score = get_model_class("Score")
+            
+            team = kwargs.get("team")
```
fpayer: Move `team` and `event` to the function signature

https://cyberskyline.slack.com/archives/C02AYEDCB9N/p1753209598808869

backend/ng/core/utils/validator.py
```diff
@@ -172,24 +177,97 @@ def validate_model_id(
...
+        data: dict[str, Any],
+        field: str = "admin_id",
+        required: bool = True,
+        friendly_name: str | None = None,
         value: Any = None,  # Injected by decorator
     ) -> None:
+        """
+        Validate that a user ID belongs to an admin user
+        """
+        if value is None:
```
fpayer: Should the lack of a user ID result in the validation check passing?

backend/ng/core/utils/validator.py
```diff
@@ -172,24 +177,97 @@ def validate_model_id(
...
+            required=required,
+            friendly_name=friendly_name
+        )
+
+    @validation_field
+    def validate_admin_id(
+        self,
+        data: dict[str, Any],
+        field: str = "admin_id",
+        required: bool = True,
```
fpayer: `required` is unused

backend/ng/core/utils/validator.py
```diff
@@ -172,24 +177,97 @@ def validate_model_id(
...
     ) -> None:
+        """
+        Validate that a user ID belongs to an admin user
+        """
+        if value is None:
+            return
+            
+        try:
+            user_id = int(value)
+            if user_id <= 0:
```
fpayer: The positive check is unnecessary - if the id is negative, then `Users.query.get` will not return anything and that will cause an error anyway.

backend/ng/core/utils/validator.py
```diff
@@ -172,24 +177,97 @@ def validate_model_id(
...
+
+    @validation_field
+    def validate_integer(
+        self,
+        data: dict[str, Any],
+        field: str,
+        min_value: int | None = None,
+        max_value: int | None = None,
+        allow_zero: bool = True,
+        required: bool = False,
```
fpayer: `required` is unused

backend/ng/scoring/models/Attempt.py
```diff
@@ -0,0 +1,291 @@
...
+
+        validator.validate_string(
+            data,
+            "submission",
+            max_length=config.MAX_SUBMISSION_LENGTH,
+            required=True,
+            friendly_name="Submission"
+        )
+
+        if "points" in data:
```
fpayer: Don't need this if statement since `required=False` is already being set

backend/ng/scoring/models/Attempt.py
```diff
@@ -0,0 +1,291 @@
...
+            validator.validate_integer(data, "points", required=False, min_value=0)
+
+        validator.validate_datetime(data, "timestamp", required=False)
+
+        validator.validate_boolean(data, "is_correct", required=False)
```
fpayer: Why is `required=False` if `is_correct` is not nullable?

backend/ng/scoring/models/Attempt.py
```diff
@@ -0,0 +1,291 @@
...
+
+        return validator.validate()
+
+    @classmethod
+    def validate_api_submission(cls, data: dict[str, Any]) -> dict[str, Any]:
```
fpayer: Still confused why this is needed since the `submission` field is going to be checked in the regular `validate` function anyway.

backend/ng/scoring/models/Attempt.py
```diff
@@ -0,0 +1,291 @@
...
+                "event_id": event_id,
+                "challenge_id": challenge_id,
+                "question_id": question_id,
+                "submission": submission,
+            }
+        )
+
+        cls.validate_attempt_allowed(user_id, team_id, event_id, challenge_id, question_id)
+
+        if timestamp is None:
```
fpayer: Setting the default timestamp needs to happen before `.validate` is called so that the timestamp can be included in the `.validate` call. As currently written, the `timestamp` is never validated if it is provided in the call to `create_attempt`

backend/ng/scoring/models/Attempt.py
```diff
@@ -0,0 +1,291 @@
...
+        if points != 0:
+            # LAZY-IMPORT
+            from .Score import Score
+            from .ScoreEvent import ScoreEvent
+
+            score = Score.find_by_team_and_event(team_id, event_id)
+            score_event = ScoreEvent.create_score_event(
+                score_id=score.id, team_id=team_id, points=points, timestamp=timestamp, commit=False
+            )
+            attempt.score_event_id = score_event.id
```
fpayer: Can we use backrefs to avoid needing to set `score_event_id` on the `attempt`?

backend/ng/scoring/models/HintRedemption.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        Serialize hint redemption for API response
+        """
+        data = {
+            "id": self.id,
+            "hint_id": self.hint_id,
+            "user_id": self.user_id,
+            "team_id": self.team_id,
+            "score_event_id": self.score_event_id,
+            "timestamp": self.timestamp.isoformat() + "Z" if self.timestamp else None,
```
fpayer: `timestamp` is not nullable, so this ternary is not needed.

backend/ng/scoring/models/Attempt.py
```diff
@@ -0,0 +1,291 @@
...
+        validator = BaseValidator()
+
+        validator.validate_string(
+            data, "submission", config.MAX_SUBMISSION_LENGTH, required=True, friendly_name="Submission"
+        )
+
+        return validator.validate()
+
+    @classmethod
+    def validate_attempt_allowed(
```
fpayer, line 107: Need a TODO to move this logic to Daniel's permission check system in the future.

backend/ng/scoring/models/HintRedemption.py
```diff
@@ -0,0 +1,247 @@
...
+        # LAZY-IMPORT
+        from ...challenge.models.Hint import Hint
+
+        hint = Hint.query.get(hint_id)
+        if not hint:
+            raise ValidationError(f"Hint {hint_id} not found")
+
+        points = -abs(hint.deduction) if hint.deduction > 0 else 0
+
+        validated_data = cls.validate(
```
fpayer, line 156: Include the timestamp in the call to `.validate`

backend/ng/scoring/models/HintRedemption.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        Validate hint redemption data.
+        """
+        validator = BaseValidator()
+
+        validator.validate_model_id(data, "hint_id", "Hint", required=True)
+        validator.validate_model_id(data, "user_id", "User", required=True)
+        validator.validate_model_id(data, "team_id", "Team", required=True)
+
+        validator.validate_integer(data, "points", required=True, max_value=0)
```
fpayer: Validation on `points` is not necessary since it is derived directly from the Hint, which has its own validation.

backend/ng/scoring/models/HintRedemption.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        cls.validate_redemption_allowed(user_id, team_id, hint_id, event_id, challenge_id)
+
+        if timestamp is None:
+            timestamp = utc_now()
+
+        # LAZY-IMPORT
+        from ...challenge.models.Hint import Hint
+
+        hint = Hint.query.get(hint_id)
```
fpayer: Move this after the call to `.validate` so that you don't need to have an additional check that the hint exists. 

Also, shouldn't this be `Hint.find_by_id`?

backend/ng/scoring/models/HintRedemption.py
```diff
@@ -0,0 +1,247 @@
...
+            from .Score import Score
+            from .ScoreEvent import ScoreEvent
+
+            member = TeamMember.find_by_user_and_team(user_id, team_id)
+            if member:
+                score = Score.find_by_team_and_event(team_id, member.event_id)
+                score_event = ScoreEvent.create_score_event(
+                    score_id=score.id, team_id=team_id, points=points, timestamp=timestamp, commit=False
+                )
+                redemption.score_event_id = score_event.id
```
fpayer: Do backrefs make this unnecessary?

backend/ng/scoring/models/HintRedemption.py
```diff
@@ -0,0 +1,247 @@
...
+        db.session.flush()
+
+        if points != 0:
+            # LAZY-IMPORT
+            from ...team.models.TeamMember import TeamMember
+            from .Score import Score
+            from .ScoreEvent import ScoreEvent
+
+            member = TeamMember.find_by_user_and_team(user_id, team_id)
+            if member:
```
fpayer: This if statement is unnecessary - `member` is already confirmed to exist by `validate_redemption_allowed`

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,268 @@
...
+        score = cls.find_by_team_and_event(team_id, event_id)
+        if not score:
+            return None
+
+        higher_scores = cls.query.filter(cls.event_id == event_id, cls.points > score.points).count()
+
+        return higher_scores + 1
+
+    @classmethod
+    def find_by_team_and_event(cls, team_id: int, event_id: int) -> Score:
```
fpayer: Why is `event_id` needed? Can't that be determined from the team?

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,268 @@
...
+        query = cls.query.filter_by(event_id=event_id).order_by(cls.points.desc())
+
+        if limit:
+            query = query.limit(limit)
+
+        scores = query.all()
+        return [score.serialize() for score in scores]
+
+    @classmethod
+    def get_team_rank(cls, team_id: int, event_id: int) -> int | None:
```
fpayer: Why is `event_id` needed? Can't that be determined from the team?

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,268 @@
...
+        return higher_scores + 1
+
+    @classmethod
+    def find_by_team_and_event(cls, team_id: int, event_id: int) -> Score:
+        """
+        Find score for a specific team in an event. Creates one if it doesn't exist.
+        """
+        score = cls.query.filter_by(team_id=team_id, event_id=event_id).first()
+
+        if not score:
```
fpayer: Perform an early-return to remove a level of indentation. 

`if score: return score`

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,268 @@
...
+        validator.validate_datetime(data, "timestamp", required=False)
+
+        return validator.validate()
+
+    @classmethod
+    def create_score(
+        cls,
+        team_id: int,
+        event_id: int,
+        team_name: str,
```
fpayer: `team_name` should be determined from the `team` itself instead of passed in as an argument.

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+        if last_update is None:
+            last_update = utc_now()
+
+        validated_data = cls.validate(
+            {
+                "team_id": team_id,
+                "event_id": event_id,
+                "team_name": team_name,
+                "points": points,
```
fpayer, line 124: Add `last_update` to the call to `.validate`

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,268 @@
...
+
+        validator.validate_datetime(data, "timestamp", required=False)
+
+        return validator.validate()
+
+    @classmethod
+    def create_score(
+        cls,
+        team_id: int,
+        event_id: int,
```
fpayer: `event_id` is unnecessary since it can be determined from the `team`

backend/ng/scoring/models/ScoreEvent.py
```diff
@@ -0,0 +1,183 @@
...
+    def serialize(self, include_admin_fields: bool = False) -> SerializedScoreEvent:
+        """
+        Serialize score event for API response
+        """
+        data = {
+            "id": self.id,
+            "score_id": self.score_id,
+            "team_id": self.team_id,
+            "points": self.points,
+            "timestamp": self.timestamp.isoformat() + "Z" if self.timestamp else None,
```
fpayer: `timestamp` is not nullable, so this ternary is unnecessary.

backend/ng/scoring/models/ScoreEvent.py
```diff
@@ -0,0 +1,183 @@
...
+
+        Returns:
+            ScoreEvent: The created score event
+        """
+        if timestamp is None:
+            timestamp = utc_now()
+
+        validated_data = cls.validate(
+            {
+                "score_id": score_id,
```
fpayer, line 100: Add timestamp to `.validate` call

backend/ng/scoring/models/ScoreEvent.py
```diff
@@ -0,0 +1,183 @@
...
+        query = query.order_by(cls.timestamp.desc())
+        if limit is not None:
+            query = query.limit(limit)
+        return query.all()
+
+    def delete_event(self, commit: bool = True) -> None:
+        """
+        Delete this score event and adjust the associated score
+        """
+        if self.score:
```
fpayer: Unnecessary if statement - `score` is not nullable.

backend/ng/support/models/Ticket.py
```diff
@@ -80,8 +156,12 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
             "subject": self.subject,
             "author_id": self.author_id,
             "status": self.status,
-            "opened_timestamp": self.opened_timestamp.isoformat() + "Z" if self.opened_timestamp else None,
-            "last_updated": self.last_updated.isoformat() + "Z" if self.last_updated else None,
+            "opened_timestamp": self.opened_timestamp.isoformat() + "Z"
```
fpayer: Unnecessary ternary - `opened_timestamp` is not nullable.

backend/ng/support/models/Ticket.py
```diff
@@ -80,8 +156,12 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
+            if self.opened_timestamp
+            else None,
+            "last_updated": self.last_updated.isoformat() + "Z"
```
fpayer: Unnecessary ternary - `last_updated` is not nullable.

fpayer (changes requested):

backend/ng/core/utils/validator.py
CarterPerez-dev: Yea becuase the decorator handles the required logic first, so if value is none here, we already know it's allowed to be empty. I was thinking this would apply for assigning/unassigning tickets or something, where the value could be none.

backend/ng/core/utils/validator.py
```diff
@@ -172,24 +177,94 @@ def validate_model_id(
             self.errors[field] = f"Error validating {friendly_name}: {str(e)}"
             return
 
-    @validation_field
```
fpayer: Why remove the `@validation_field`?

backend/ng/core/utils/validator.py
fpayer: In that case, isn't this `if` statement redundant then? Since the decorator will already perform an early-return if value is not required and is None

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,278 @@
...
+    def validate(cls, data: dict[str, Any]) -> dict[str, Any]:
+        """
+        Validate score data.
+        """
+        validator = BaseValidator()
+
+        validator.validate_model_id(data, "team_id", "Team", required=True)
+        validator.validate_model_id(data, "event_id", "Event", required=True)
+
+        if "points" in data:
```
fpayer: Is this `if` statement necessary if `required=False`?

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,278 @@
...
+            Score: The created score instance
+        """
+        if last_update is None:
+            last_update = utc_now()
+
+        if team is None:
+            # LAZY-IMPORT
+            from ...team.models.Team import Team
+
+            team = db.session.get(Team, team_id)
```
fpayer: Why use `db.session.get` instead of `Team.find_by_id`?

backend/ng/scoring/models/Score.py
```diff
@@ -0,0 +1,278 @@
...
+
+    @classmethod
+    def find_by_team(cls, team_id: int) -> Score:
+        """
+        Find score for a specific team. Creates one if it doesn't exist.
+        """
+        # LAZY-IMPORT
+        from ...team.models.Team import Team
+
+        team = db.session.get(Team, team_id)
```
fpayer: Why not use `Team.find_by_id`?

fpayer (changes requested):

backend/ng/core/utils/validator.py
CarterPerez-dev: Seemed redundent because it's now just a wrapper for `validate_integer`, but it actually makes slightly more sense to keep it now that I think about it.

fpayer (approved):

PR #82, opened 2025-07-21
Integrate dashboard page with backend data

trevnels-cs:
- Connects me/events to the dashboard page and sort into past, present, and future events
- Reworks dashboard/overview components to take model objects instead of individual parameters
- Rough event status badge logic based on timestamps

### DIFF
--- a/frontend/src/components/EventHeader.tsx
+++ b/frontend/src/components/EventHeader.tsx
@@ -1,24 +1,31 @@
+import type { Event } from '@/types';
 import {
-  Heading,
-  Flex,
-  Box,
   AspectRatio,
+  Box,
+  Flex,
+  Heading,
 } from '@radix-ui/themes';
 import type { ReactNode } from 'react';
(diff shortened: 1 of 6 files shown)


frontend/src/components/EventHeader.tsx
```diff
@@ -1,24 +1,31 @@
...
 }: {
-    name: string;
-    description: string;
-    state: 'upcoming' | 'waiting' | 'live' | 'ended';
-    children?: ReactNode;
+  event: Event;
+  children?: ReactNode;
 }) {
+  const now = new Date();
+  let state: 'upcoming' | 'live' | 'ended' | 'waiting' | null = null;
```
cjones-cs, line 25: You should dereference the dates from the event object

cjones-cs: also the other variables

cjones-cs (approved):

PR #84, opened 2025-07-21
Bug fix to allow users to access registered events even if private

droge91:
Bug fix plus a couple tests to verify functionality.

### DIFF
--- a/backend/ng/core/middleware/permission_middleware.py
+++ b/backend/ng/core/middleware/permission_middleware.py
@@ -38,7 +38,7 @@ def event_only_public(f):
         event = kwargs.get('event')
         current_user = kwargs.get('current_user')
         if Demographic.find_by_user_and_event(current_user.id, event.id) is None:
-            if not event or not event.public:
+            if not event.public:
                 return error_response("Event not found", "not_found", 404)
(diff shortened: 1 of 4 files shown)


backend/ng/core/middleware/permission_middleware.py
```diff
@@ -35,8 +36,10 @@ def event_only_public(f):
     @wraps(f)
     def wrapped(*args, **kwargs):
         event = kwargs.get('event')
-        if not event or not event.public:
-            return error_response("Event not found", "not_found", 404)
+        current_user = kwargs.get('current_user')
+        if Demographic.find_by_user_and_event(current_user.id, event.id) is None:
+            if not event or not event.public:
```
fpayer: if not event should come before Demographic.find_by_user_and_event(current_user.id, event.id)

fpayer (changes requested):

backend/ng/core/middleware/permission_middleware.py
droge91: actually the check is probably not necessary in the first place as the previous middlewares will throw an error if the event cant be found.

fpayer (approved):

PR #85, opened 2025-07-22
Fixes kwarg signature issue with auth middleware

fpayer:


trevnels-cs (approved):
Looks good, might be worth doing this for other middleware that sets kwargs too if we aren't already.

PR #86, opened 2025-07-22
Feature/event registration

cjones-cs:


frontend/src/hooks/events.ts
```diff
@@ -46,6 +46,19 @@ export function useUserEvents(userId: number | null) {
   );
 }
 
+/**
+ * Registeres user for a specific event
+ * @param event_id The event id
+ * @param team_name The leaderboard name
+ */
+export function registerMyEvent(eventId: number, teamName: string) {
+  apiMutation(`/events/${eventId}/me/register`, { team_name : teamName }, {
```
RayofLightz, line 55: Are we just doing teamName for now? I know this will also take an invite code.

RayofLightz (approved):
Looks good

frontend/src/hooks/events.ts
cjones-cs: Yes, because we don't have team management in place, it becomes nearly impossible to properly test. This pr only handles individual game.

PR #87, opened 2025-07-24
Modal refactor and form styling updates

trevnels-cs:
Updates the modal component to work with forms and promise-based submit actions. Also adds more flexibility in defining the modal trigger element and allows disabling the submit button when certain conditions are true (or if the form has not been edited.)

As part of this change, I also moved the event editing form into a modal similar to event creation. This is activated by an edit button to prevent accidental changes, which may have occurred with the previous sidebar form.

### DIFF
--- a/frontend/src/components/Modal.tsx
+++ b/frontend/src/components/Modal.tsx
@@ -1,57 +1,139 @@
-import { Button, Dialog, Flex } from '@radix-ui/themes';
-import { useState, type ReactNode } from 'react';
+import {
+  Box,
+  Button,
+  Dialog,
+  Flex,
+} from '@radix-ui/themes';
+import type { accentColors } from '@radix-ui/themes/props';
+import { Form } from 'radix-ui';
+import { useEffect, useState, type ReactNode } from 'react';
 import { TbX } from 'react-icons/tb';
+import { ErrorCallout } from './Callouts';
 
 interface ModalProps {
   title: string,
-  buttonText: string,
-  children: ReactNode,
-  onSubmit: () => void,
-  onSubmitText: string,
-  onSubmitColor?: 'lime' | 'red'
+  description?: string,
+  children?: ReactNode,
+  trigger: ReactNode,
+  onSubmit?: (formData: FormData) => Promise<unknown>,
+  onOpenChange?: (open: boolean) => void,
+  submitVerb?: string,
+  submitColor?: typeof accentColors[number],
+  submitDisabled?: boolean,
+  requireTouchingForm?: boolean,
 }
 
 export default function Modal({
-  title, buttonText, children, onSubmit, onSubmitText, onSubmitColor = 'lime',
-}:ModalProps) {
+  title,
+  description,
+  children,
+  trigger,
+  onSubmit,
+  onOpenChange,
+  submitVerb = 'Submit',
+  submitColor,
+  submitDisabled = false,
+  requireTouchingForm = false,
+} : ModalProps) {
   const [ open, setOpen ] = useState<boolean>(false);
+  const [ error, setError ] = useState<string | null>(null);
+  const [ formTouched, setFormTouched ] = useState<boolean>(false);
+
+  useEffect(() => {
+    if (!open) {
+      // reset state when modal closes
+      setFormTouched(false);
+      setError(null);
+    }
+  }, [ open ]);
 
   return (
-    <Dialog.Root open={open} onOpenChange={setOpen}>
+    <Dialog.Root open={open} onOpenChange={(o) => { onOpenChange?.(o); setOpen(o); }}>
       <Dialog.Trigger>
-        <Button>{buttonText}</Button>
+        {trigger}
       </Dialog.Trigger>
 
-      <Dialog.Content aria-describedby={undefined}>
-        <Flex direction="row">
-          <Dialog.Title
-            className="pb-2"
-          >
-            {title}
-          </Dialog.Title>
-          <Dialog.Close>
-            <button type="button" aria-label="Close" className="absolute left-auto right-4 top-4">
-              <TbX />
-            </button>
-          </Dialog.Close>
-        </Flex>
-        {children}
-        <Flex gap="3" mt="3" justify="end">
-          <Dialog.Close>
-            <Button variant="solid" color="gray">
-              Cancel
-            </Button>
-          </Dialog.Close>
-          <Button
-            color={onSubmitColor}
-            onClick={() => {
-              onSubmit();
+      <Dialog.Content
+        className="flex flex-col gap-3"
+        // aria-describedby should be set to undefined if no description is provided, otherwise it should not be set at all.
+        // as far as i know, prop spreading is the only way to accomplish this.
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        {
+          ...(!description ? { 'aria-describedby' : undefined } : {})
+        }
+      >
+        <Box>
+          <Flex direction="row" justify="between" align="center">
+            <Dialog.Title mb="0">{title}</Dialog.Title>
+            <Dialog.Close>
+              <Button
+                type="button"
+                aria-label="Close"
+                variant="ghost"
+                color="gray"
+                className="!m-0 !p-1"
+              >
+                <TbX className="text-xl" />
+              </Button>
+            </Dialog.Close>
+          </Flex>
+          {description && (
+          <Dialog.Description color="gray">
+            {description}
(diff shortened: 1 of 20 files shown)


frontend/src/routes/admin/events/EventDataForm.tsx
```diff
@@ -191,7 +134,7 @@ export default function EventDataForm({
           <Form.Control asChild>
             <Checkbox defaultChecked={initial?.locked} size="3" />
           </Form.Control>
-          <Form.Label>Locked</Form.Label>
+          <Form.Label>Disable Team Management</Form.Label>
```
cjones-cs, line 137: This may not be correct. Yes, team management is the primary thing we use the locked state for, but there may be other instances, like preventing individual players from changing their leaderboard name or something like that. We should double check with backend

frontend/src/routes/admin/users/UserSidebar.tsx
```diff
@@ -1,20 +1,16 @@
...
+import { useTeamMembers, useUserTeams } from '@/hooks/team';
 import type { Event, Team, User } from '@/types';
+import { Button, Table } from '@radix-ui/themes';
+import AdminDataList from 'components/AdminDataList';
 import AdminSidebar from 'components/AdminSidebar';
+import AdminSidebarHeader from 'components/AdminSidebarHeader';
+import { ErrorCallout } from 'components/Callouts';
+import Entity from 'components/Entity';
 import RoleBadge from 'components/RoleBadge';
+import _ from 'lodash';
```
cjones-cs: Importing the keyBy function instead of the whole library is the style I've been going with on other components

frontend/src/routes/events/OverviewTabs/TeamTab/AssignCaptainModal.tsx
```diff
@@ -1,26 +1,30 @@
...
+      )}
       onSubmit={assignCaptain}
-      onSubmitText="Assign"
-    >
-      <Text>
-        You will no longer be a Team Captain. By assigning a new member as Team Captain, you will lose your ability to add and remove team members.
-      </Text>
-    </Modal>
+      submitVerb="Assign"
+      submitColor="amber"
```
cjones-cs: Assign is supposed to be lime.

frontend/src/routes/events/OverviewTabs/TeamTab/LeaveTeamModal.tsx
```diff
@@ -25,31 +26,45 @@ export default function LeaveTeamModal({ transferCaptain, membersList }: LeaveTe
...
-      onSubmitText="Leave"
-      onSubmitColor="red"
+      onOpenChange={(open) => {
+        if (open) {
+          // reset captain state when modal is closed and reopened
+          setNewCaptain('');
+        }
+      }}
+      submitVerb="Leave"
+      submitColor="red"
```
cjones-cs, line 44: amber or red?

frontend/src/routes/events/OverviewTabs/TeamTab/LeaveTeamModal.tsx
```diff
@@ -25,31 +26,45 @@ export default function LeaveTeamModal({ transferCaptain, membersList }: LeaveTe
...
+        </Button>
+      )}
       onSubmit={leaveTeam}
-      onSubmitText="Leave"
-      onSubmitColor="red"
+      onOpenChange={(open) => {
+        if (open) {
+          // reset captain state when modal is closed and reopened
+          setNewCaptain('');
+        }
```
cjones-cs, line 41: Love this

frontend/src/routes/events/OverviewTabs/TeamTab/LeaveTeamModal.tsx
```diff
@@ -14,9 +15,9 @@ interface LeaveTeamModalProps {
 }
 
 export default function LeaveTeamModal({ transferCaptain, membersList }: LeaveTeamModalProps) {
```
cjones-cs, line 17: Is it better form with swr to get these items in each modal or to pass them from the parent like this? If I recall, the membersList excludes the currently logged in player, which I think is why I had it passed in like this, so the parent component could do the operation once, then pass it everywhere needed.

frontend/src/routes/admin/events/EventDataForm.tsx
trevnels-cs: I had discussed with Franz last week and this is the wording we had arrived at for now. While individual players updating their leaderboard name is still technically team management from the backend, you're right that there is probably a clearer way of wording this to clearly cover individual and team cases. My main concern was that "Locked" is extremely unclear

frontend/src/routes/events/OverviewTabs/TeamTab/RemovePlayerModal.tsx
```diff
@@ -1,28 +1,30 @@
...
     console.log('remove the player', id);
   };
 
   return (
     <Modal
       title={`Are you sure you want to remove ${name}?`}
-      buttonText="Remove"
+      description="They will no longer have access to participate with this team. The invite code for the team will change."
+      trigger={(
+        <Button variant="soft" color="red">
```
cjones-cs, line 20: I put a comment above, but is our standard in other parts of the site to use amber or red for these types of buttons? I used red everywhere here, but I can make note to fix this when I do the hooks

frontend/src/routes/events/OverviewTabs/TeamTab/LeaveTeamModal.tsx
trevnels-cs: I believe SWR recommends using the hooks as close to the places that need the data since it can deduplicate requests, but if there's logic in a parent component you want to share here, I think it makes sense to pass the data in instead of having to do that same logic twice. I'm doing something similar with the admin grids.

frontend/src/routes/events/OverviewTabs/TeamTab/LeaveTeamModal.tsx
trevnels-cs: I think red makes more sense here since this is a destructive operation

frontend/src/routes/events/RegistrationModal.tsx
```diff
@@ -1,43 +1,42 @@
...
+  const register = async (data: FormData) => {
+    const leaderboardName = data.get('leaderboard_name') as string;
+    return registerMyEvent(eventId, leaderboardName);
   };
 
   return (
     <Modal
       title="Are you sure you want to register for this event?"
-      buttonText="Register"
+      description="Please do not use your real name."
```
cjones-cs, line 18: Using the description like this gets rid of the italics I used, which is fine since that was a decision I didn't ask Lauren about. Just wondering if it needs some emphasis or not

frontend/src/routes/events/RegistrationModal.tsx
```diff
@@ -1,43 +1,42 @@
...
+      trigger={(
+        <Button variant="soft">Register</Button>
+      )}
       onSubmit={register}
-      onSubmitText="Register"
+      submitVerb="Register"
+      submitColor="lime"
```
cjones-cs: Doesn't the submit color default to lime? If so, we don't need to pass this over.

frontend/src/components/Modal.tsx
```diff
@@ -1,57 +1,139 @@
...
-  title, buttonText, children, onSubmit, onSubmitText, onSubmitColor = 'lime',
-}:ModalProps) {
+  title,
+  description,
+  children,
+  trigger,
+  onSubmit,
+  onOpenChange,
+  submitVerb = 'Submit',
+  submitColor,
```
cjones-cs: Can we default this to lime? Most of our submit buttons are either going to be lime or red

frontend/src/routes/events/OverviewTabs/TeamTab/RemovePlayerModal.tsx
trevnels-cs: I've been using semantic colors for the trigger buttons depending on what they're for - modals that are used to create things are `lime`, modals that are used to update/modify/edit or involve privilege (captaincy) are `amber`, and modals that are used for deletion/leaving/destructive operations are `red`. I'm also working on another PR that adds constants for some of these semantic colors to more easily reuse them and allow us to change them easily if needed

cjones-cs (changes requested):
Just a few small things. Overall, looks good

cjones-cs (approved):

PR #89, opened 2025-07-28
backend linter fix

CarterPerez-dev:
needs to run from backend/ not root.


CarterPerez-dev:
lol, the first fix worked but when looking at it I for some reason thought it didn't, either way, this one works too.

fpayer (approved):

PR #90, opened 2025-07-28
Fixes install script + add more readme docs

RayofLightz:
I think these should fix the issues you ran into that you mentioned in the dev sync.


trevnels-cs (approved):

trevnels-cs:
We should probably also add `ssl` to the gitignore

PR #91, opened 2025-07-28
Ctfd ext routefixes

droge91:
Fixes admin/user routes and adds test cases for them
Adds better documentation for all routes
A few business logic fixes for event routes.


backend/ng/event/models/Event.py
```diff
@@ -86,56 +86,62 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
-            config.EVENT_NAME_MAX_LENGTH,
-            required=True,
-            friendly_name="Event name",
-        )
-        validator.validate_string(
-            data,
-            "description",
-            config.EVENT_DESCRIPTION_MAX_LENGTH,
-            required=False,
+        if "name" in data:
```
trevnels-cs: Why were these checks added? If name is missing from `data`, don't we still want it go through the validator and raise an exception since it's a required field? It seems like the current code would skip over it and accept data with missing fields.

droge91: its due to creating an event and updating an event sharing the same validator right now. If only some attributes are being updated then it should be able to validate just those. you're right though this causes an issue where its not validating that all required fields are there when creating an event

droge91: I guess since we are switching to using puts instead of patches maybe that logic is unnecessary? if the frontend is just gonna send the full attributes

fpayer (approved):

backend/ng/event/models/Event.py
trevnels-cs: Yes, with PUT all of the fields will be present, even if they haven't changed

PR #92, opened 2025-07-29
Initial challenge page integration

trevnels-cs:
Challenge page now renders description/questions/other challenge info and the VNC iframe. This page is very much an MVP - error/loading states especially are nowhere near final, but the core functionality is there. The events section of the admin panel also now has a dropzone/file-picker for uploading challenges to an event, though there is currently no way to update or delete them (be careful!)

The "connect workspace" button on a challenge page will start the challenge's containers if they are not already, and connect the user's workspace container to the challenge network. Currently there's no way to know what the state of the workspace is from the frontend, so this button will always be available even if the challenge is already connected.

There's also a modal attached to this button that will eventually be used to confirm disconnecting from a previously connected challenge before connecting to the new one (modals also now have a loading state when their submit promises are pending - the submit button becomes a spinner)

I also swapped out the workspace container with one that works consistently for me, and changed the configured docker host IP to point at the host system's docker instance (so it shouldn't need to be changed from the default during development)

Currently looks like this:

image

image


frontend/src/routes/admin/events/ChallengeUploadModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+          />
+          <Heading size="6"><TbUpload /></Heading>
+          <Text>Drag and drop a YAML file here, or click to select one.</Text>
+
+        </Callout.Root>
+      )}
+    </Dropzone>
+  );
+}
+
```
RayofLightz: Is there a particular reason this is its own component?

frontend/src/routes/events/challenge/ChallengeQuestion.tsx
```diff
@@ -0,0 +1,128 @@
...
+              Unanswered
+            </Text>
+          )}
+        </Box>
+        <Text size="2" color="gray">
+          {attempts}
+          {' '}
+          attempt
+          {attempts !== 1 ? 's ' : ' '}
+          left
```
RayofLightz: Use a template literal

frontend/src/routes/admin/events/ChallengeUploadModal.tsx
trevnels-cs: Before the new Modal abstraction the parent of this was a lot more complex so I had split this piece off - now that the Modal is abstracted away it probably doesn't need to be separate

frontend/src/routes/admin/events/ChallengeUploadModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+      }}
+      multiple={false}
+    >
+      {({ getRootProps, getInputProps, isDragActive }) => (
+        <Callout.Root
+          // eslint-disable-next-line react/jsx-props-no-spreading
+          {...getRootProps()}
+          variant={(isDragActive) ? 'surface' : 'outline'}
+          className="cursor-pointer !p-8 !flex flex-col !items-center"
+          color="lime"
```
cjones-cs: We may want to use the generic accent color here instead.

frontend/src/routes/admin/events/ChallengeUploadModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export default function ChallengeUploadModal({ eventId }: { eventId: number }) {
+  const [ fileContent, setFileContent ] = useState<string | null>(null);
+
+  return (
+    <Modal
+      title="Add Challenge"
+      description="Upload a challenge YAML file to create a new challenge for this event."
+      trigger={(
+        <Button variant="soft" color="lime">
```
cjones-cs, line 22: We should probably use accent here

trevnels-cs: I think it makes more sense for this button to be green to help convey that it's used for adding challenges

frontend/src/routes/admin/events/EventSidebar.tsx
```diff
@@ -24,6 +29,15 @@ export default function EventSidebar({ entity }: { entity: Event }) {
         event={entity}
       />
 
+      <AdminSidebarHeader title="Challenges">
+        <ChallengeUploadModal eventId={entity.id} />
```
cjones-cs, line 33: Event is getting passed into this function, but then passed down again to the modal. Can we query swr so we don't have to do this prop passing or is it necessary in this instance?

frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
```diff
@@ -1,53 +1,63 @@
+import type { Challenge, MeChallenge } from '@/types';
```
cjones-cs, line 1: It seems odd to me that challenge and mechallenge are separate types. Making a note to check this below

frontend/src/routes/admin/events/EventSidebar.tsx
trevnels-cs: Passing the event to the sidebar as a prop prevents a duplicate GET for the individual event - the datagrid already has all of the `Event` objects as a list, so indexing into that (in the `AdminGrid` component) is the most efficient way to get the selected Event object and pass it to the sidebar. If we eventually switch to loading the grid in chunks/pages, the sidebar will probably need to perform its own fetch, but for now it's not needed.

For the ChallengeUploadModal, I'm just passing the id of the selected event, not the whole object - I could also get the `id` from search params inside the component, but that would be more lines of code than passing `id` down as a prop.

frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
```diff
@@ -1,53 +1,63 @@
...
-              <TbCheck title="Completed" />
-            </Text>
-          ) : null}
-          <Text size="2" color={color} aria-label={`${points} points`}>
-            {points}
-            pts
+          {complete && (
+          <Text size="2" color={color}>
+            <tb.TbCircleCheckFilled />
           </Text>
```
cjones-cs: indent to match the others

frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
trevnels-cs: This was a backend design decision to separate system-wide challenge info from an individual team's progress - these are the return types for two separate endpoints

frontend/src/routes/admin/events/ChallengeUploadModal.tsx
RayofLightz: Yeah best not to. I know I am not really working on the frontend here, but I have nightmares about multiple components in a single file

frontend/src/routes/events/OverviewTabs/ChallengesTab/index.tsx
```diff
@@ -1,31 +1,62 @@
-import { Container, TextField, Grid } from '@radix-ui/themes';
-import { TbCode, TbSearch } from 'react-icons/tb';
+import { useEventChallenges, useMyChallenges } from '@/hooks/challenge';
+import {
+  Container,
+  Grid,
+  Text,
+  TextField,
+} from '@radix-ui/themes';
+import _ from 'lodash';
```
cjones-cs: Import keyBy directly instead of the whole lib

frontend/src/routes/events/OverviewTabs/ChallengesTab/index.tsx
```diff
@@ -1,31 +1,62 @@
...
+import { useMemo, useState } from 'react';
+import { TbCancel, TbSearch } from 'react-icons/tb';
+import { useParams } from 'react-router';
 import ChallengeCard from './ChallengeCard';
 
 export default function ChallengesTab() {
+  const { idEvent } = useParams<{idEvent: string}>();
+  const { data } = useEventChallenges(Number(idEvent));
+  const { data : myChallenges } = useMyChallenges(Number(idEvent));
```
cjones-cs, line 17: idEvent is being cast from string to Number. Is there a consequence for invalid cast since this comes from the url params? I assume the page may not load at all when the idEvent is bad

frontend/src/routes/events/challenge/ChallengeQuestion.tsx
```diff
@@ -0,0 +1,128 @@
...
+  points: number;
+  attempts: number;
+  status: 'unanswered' | 'incorrect' | 'correct';
+  valueOverride?: string;
+}) {
+  const handleSubmit = useCallback((event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    const { flag } = Object.fromEntries(new FormData(event.currentTarget));
+
+    alert(flag);
```
cjones-cs, line 37: I assume this is one of the messy things you're going to remove later?

frontend/src/routes/events/OverviewTabs/ChallengesTab/index.tsx
trevnels-cs: In this case, it will result in a `NaN`, which will make the SWR hooks not run - you get a navbar and page without any data on it. Better error handling here is something I've thought about, but for MVP it wasn't a priority.

frontend/src/routes/events/challenge/ChallengeQuestion.tsx
trevnels-cs: Yep, this is a placeholder until answer checking is ready

frontend/src/routes/events/challenge/ChallengeQuestion.tsx
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }, [ status ]);
+
+  return (
+    <Flex direction="column">
+      <Flex direction="row" gap="2" align="center" justify="start">
+        <Heading size="5" color={status === 'correct' ? 'lime' : undefined}>
+          {name}
+        </Heading>
+        <Text size="2" color={status === 'correct' ? 'lime' : 'gray'}>
```
cjones-cs, line 57: If you make the color const a function, you may be able to add 'unanswered' as a case and not have to do so many ternary statements.

frontend/src/routes/events/challenge/ChallengeQuestion.tsx
```diff
@@ -0,0 +1,128 @@
...
+                </TextField.Slot>
+              </TextField.Root>
+            </Form.Control>
+            <Form.Message match="valueMissing">Flag cannot be empty</Form.Message>
+          </Form.Field>
+
+          {status !== 'correct' && attempts > 0 && (
+            <Button variant="soft" size="2" type="submit" color="amber">
+              Submit
+            </Button>
```
cjones-cs, line 117: Do they have a max attempts on challenge submits? That may be a case to account for here as well

trevnels-cs: I intend for `attempts` to be the number of remaining attempts - will start at the max and tick down as they submit. Once it reaches 0 the submit button will no longer be available

frontend/src/routes/events/challenge/ChallengeSidebar.tsx
```diff
@@ -0,0 +1,122 @@
...
+                </Flex>
+                <Skeleton loading={!challenge}>
+                  <Text color="gray">
+                    {challenge?.summary || 'No summary available'}
+                  </Text>
+                </Skeleton>
+              </Box>
+            )}
+
+            <ConnectModal eventId={Number(idEvent)} challengeId={Number(idChallenge)} />
```
cjones-cs: Should we be using the params version here or would it be safer to use the ids that come from the swr query?

trevnels-cs: Latter is probably better, will swap that out

frontend/src/routes/events/challenge/ChallengeSidebar.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <Inset side="all" className="shrink !overflow-y-auto">
+          <Flex direction="column" gap="3" className="p-3">
+            <Skeleton loading={!challenge}>
+              {!challenge && (
+              <>
+                {/* Placeholder text to produce a skeleton effect from - actual text is not shown. */}
+                <Text>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</Text>
+                <Text>Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</Text>
+                <Text>Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</Text>
+              </>
```
cjones-cs: Should we go ahead and ask Lauren/etc for some real placeholder text?

frontend/src/routes/events/challenge/ChallengeSidebar.tsx
```diff
@@ -0,0 +1,122 @@
...
+              )}
+              {challenge?.description && (
+              <RadixMarkdown>
+                {challenge.description}
+              </RadixMarkdown>
```
cjones-cs, line 94: The indent on these isn't consistent with the other files. After the `&& (`, the new line should be indented. There are several like that in this file

frontend/src/routes/events/challenge/ChallengeSidebar.tsx
```diff
@@ -0,0 +1,122 @@
...
+            </Skeleton>
+            <Flex direction="column" gap="3">
+              {questions?.map((question) => (
+                <ChallengeQuestion
+                  key={question.id}
+                  name={question.name}
+                  question={question.body}
+                  placeholder={question.placeholder}
+                  points={question.points}
+                  attempts={question.max_attempts - 0}
```
cjones-cs: Why minus zero? Wouldn't that just be itself?

frontend/src/routes/events/challenge/ChallengeSidebar.tsx
trevnels-cs: The content of this text is never shown, it gets masked out to produce the challenge loading state:
image

frontend/src/routes/events/challenge/ChallengeSidebar.tsx
trevnels-cs: This is the default way that they are formatted - I can go through and fix these by hand, but we should look at adjusting the formatter if we want this to be handled consistently

frontend/src/routes/events/challenge/HintsModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+}: {
+    eventId: number;
+    challengeId: number;
+}) {
+  const { data, error } = useChallenge(eventId, challengeId);
+
+  return (
+    <Modal
+      title="Hints"
+      description="Hints can be redeemed in exchange for a deduction from your score."
```
cjones-cs, line 18: Thought for later, but we should probably tell them how much of a deduction it is. Even if it's just a link to some documentation on the subject or something like that

trevnels-cs: The deduction is a field on each hint, and the value for each will be shown in the table here

frontend/src/routes/events/challenge/ChallengeSidebar.tsx
trevnels-cs: The 0 is a placeholder for data we don't have from the backend yet - I was testing the attempts logic for the component by incrementing that value to simulate making submissions

frontend/src/routes/events/challenge/ChallengeQuestion.tsx
cjones-cs: Is it possible for the variable to be named attemptsRemaining? I recall having this confusion with the other product as well. Attempts by itself can mean max attempts or attempts made or attempts remaining.

frontend/src/routes/events/challenge/HintsModal.tsx
cjones-cs: Doesn't the table only show already accepted hints? I had assumed the hints modal was a list of all the hints previously used on the challenge, so if they've not used a hint, they wouldn't have known how much the deduction would be.

frontend/src/routes/events/challenge/ChallengeQuestion.tsx
trevnels-cs: Working on cleaning up some of the color logic and making shared constants for a lot of them in a separate PR, I will make note of this for that set of changes

frontend/src/routes/events/challenge/HintsModal.tsx
trevnels-cs: I was envisioning the table showing all available hints - redeemed ones will show their content and unredeemed ones would have a button to redeem

cjones-cs (approved):

PR #93, opened 2025-07-30
Added more descriptive stack traces to API errors in debug mode

fpayer:


trevnels-cs:
Would you be able to take a look at the data for `APIException`/`ValidationException` as well, specifically making sure that validation error messages get sent back to the client - I believe they're put into a separate field on the exception and not the main exception message, which is why they aren't shown currently.

fpayer:
Sure, I'll look into that

trevnels-cs (approved):

PR #97, opened 2025-08-04
update navbar

cjones-cs:


frontend/src/components/NavBar.tsx
```diff
@@ -1,59 +1,78 @@
...
+    decoration-2`;
 
   return (
     <NavigationMenu.Root className="h-[var(--NavBarHeight)]">
-      <NavigationMenu.List className="flex p-1 shadow-md pr-4">
+      <NavigationMenu.List className="flex p-1 pr-4 dark:border-b-(--gray-a6) dark:border-b-1">
         <div className="flex">
-          <NavigationMenu.Item className="p-2">
+          <NavigationMenu.Item className={location.pathname === '/' ? activeLinkClass : defaultLinkClass}>
             <NavLink to="/">
```
trevnels-cs: I'd recommend putting these `NavLink`s around the `NavigationMenu.Item`s so that the entire box is clickable instead of just the text. When the cursor is over the nav item but not directly over the text, the item still goes into the hover state, even though clicking doesn't do anything - putting the entire nav item inside the link resolves this

trevnels-cs (approved):

PR #99, opened 2025-08-05
Feature/event navigation

cjones-cs:
Improve navigation from events list to event detail

### DIFF
--- a/frontend/src/routes/events/RegistrationModal.tsx
+++ b/frontend/src/routes/events/RegistrationModal.tsx
@@ -11,7 +11,7 @@ export default function RegistrationModal({ eventId }: {eventId : Event['id']}) 
 
   const register = async (data: FormData) => {
     const leaderboardName = data.get('leaderboard_name') as string;
-    return registerMyEvent(eventId, leaderboardName).finally(() => {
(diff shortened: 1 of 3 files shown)


frontend/src/routes/events/RegistrationModal.tsx
```diff
@@ -1,15 +1,19 @@
...
 export default function RegistrationModal({ eventId }: {eventId : Event['id']}) {
   // Todo: alternate case when team game vs individual
+  const navigate = useNavigate();
 
   const register = async (data: FormData) => {
     const leaderboardName = data.get('leaderboard_name') as string;
-    return registerMyEvent(eventId, leaderboardName);
+    return registerMyEvent(eventId, leaderboardName).finally(() => {
+      navigate(`/events/${eventId}`);
+    });
```
RayofLightz: Do we want to `.finally` this? As this will redirect even on a rejection so registration failed

RayofLightz (approved):
Looks mostly good. Just fix the linter errors

frontend/src/routes/events/RegistrationModal.tsx
cjones-cs: Changed it to .then so it only happens on success

PR #100, opened 2025-08-06
Adds most of the admin stuff

RayofLightz:
This should add in most the container admin stuff


backend/ng/containers/controllers/recycle_container.py
```diff
@@ -0,0 +1,5 @@
+from ..models.ContainerInstance import ContainerInstance
+
+def recycle_container(instance_id):
+    instance = ContainerInstance.get_instance_by_id(instance_id)
```
fpayer: Would recommend creating a middleware loader for ContainerInstance to provide graceful error handling if the id does not exist.

backend/ng/containers/controllers/restart_container.py
```diff
@@ -0,0 +1,5 @@
+from ..models.ContainerInstance import ContainerInstance
+
+def restart_container(instance_id):
+    instance = ContainerInstance.get_instance_by_id(instance_id)
```
fpayer: Move to middleware loader

backend/ng/containers/models/ContainerInstance.py
```diff
@@ -21,12 +28,34 @@ def __repr__(self):
     def create_container_instance(cls, blueprint: int, team: int, commit: bool = True):
```
fpayer, line 32: There seems to be a good amount of repeated logic between the `if db_exists` and when it is not branches. Could there be some refactor to consolidate this?

backend/ng/containers/models/ContainerInstance.py
```diff
@@ -78,3 +107,91 @@ def render_container_name(team_id: int, hostname: str, challenge_id: int) -> str
     @staticmethod
     def render_network_name(team_id: int, network_name: str, challenge_id: int) -> str:
         return f"{network_name}-{team_id}-{challenge_id}"
+
+    @staticmethod
+    def parser_network_name(network: str):
```
fpayer: I think this is a typo. Should be called `parse_network_name`?

backend/ng/containers/models/ContainerInstance.py
```diff
@@ -78,3 +107,91 @@ def render_container_name(team_id: int, hostname: str, challenge_id: int) -> str
...
+
+        return SerializedInstanceStats(**data)
+
+    @classmethod
+    def get_instance_by_id(cls, instance_id: int):
+        return cls.query.filter_by(id=instance_id).first()
+
+    def restart(self):
+        client = get_client(config.DOCKER_HOST)
+        ctr = client.containers.get(self.dockerid)
```
fpayer: Do we need any explicit error handling if the container DNE?

backend/ng/containers/models/IndvidualContainer.py
```diff
@@ -107,3 +119,70 @@ def get_novnc_port(self):
...
+        if not current_challenge_network:
+            return None
+
+        parsed_network = ContainerInstance.parse_network_name(current_challenge_network)
+
+        return parsed_network.challenge_id
+
+    def restart(self):
+        client = get_client(self.hostip)
+        ctr = client.containers.get(self.dockerid)
```
fpayer: Do we need any explicit error handling if the container DNE?

backend/ng/user/routes/admin_routes.py
```diff
@@ -64,4 +65,82 @@ class UserTeamsAdminResource(Resource):
...
+    @admin_endpoint()
+    @users_admin_namespace.doc(
+        description="Get current status of user's container",
+        responses={
+            200: "Sucess",
+            400: "Bad request"
+        }
+    )
+    def get(self, user_id):
+        ctr = IndvidualContainer.get_user_indvidual_container(user_id)
```
fpayer: Move to middleware loader

backend/ng/user/routes/admin_routes.py
```diff
@@ -64,4 +65,82 @@ class UserTeamsAdminResource(Resource):
...
+    @admin_endpoint()
+    @users_admin_namespace.doc(
+        description="Get user's container",
+        responses={
+            200: "Sucess",
+            400: "Bad request"
+        }
+    )
+    def get(self, user_id):
+        ctr = IndvidualContainer.get_user_indvidual_container(user_id)
```
fpayer: Move to middleware loader

backend/ng/user/routes/admin_routes.py
```diff
@@ -64,4 +65,82 @@ class UserTeamsAdminResource(Resource):
...
+    @admin_endpoint()
+    @users_admin_namespace.doc(
+        description="Get current challenge user's indvidual container is connected to",
+        responses={
+            200: "Sucess",
+            400: "Bad request"
+        }
+    )
+    def get(self, user_id):
+        ctr = IndvidualContainer.get_user_indvidual_container(user_id)
```
fpayer: Move to middleware loader

backend/ng/user/routes/admin_routes.py
```diff
@@ -64,4 +65,82 @@ class UserTeamsAdminResource(Resource):
...
+    @admin_endpoint()
+    @users_admin_namespace.doc(
+        description="Restart a user's indvidual container",
+        responses={
+            200: "Sucess",
+            400: "Bad request"
+        }
+    )
+    def get(self, user_id):
+        ctr = IndvidualContainer.get_user_indvidual_container(user_id)
```
fpayer: Move to middleware loader

backend/ng/user/routes/admin_routes.py
```diff
@@ -64,4 +65,82 @@ class UserTeamsAdminResource(Resource):
...
+    @admin_endpoint()
+    @users_admin_namespace.doc(
+        description="Delete a user's indvidual container while keeping the db object",
+        responses={
+            200: "Sucess",
+            400: "Bad request"
+        }
+    )
+    def get(self, user_id):
+        ctr = IndvidualContainer.get_user_indvidual_container(user_id)
```
fpayer: Move to middleware loader

fpayer (changes requested):
Might want to consider some additional util functions or decorators to assist with common usage patterns with the docker client.

Right now the main QOL thing I could see would be to have a way to get a container (and error if not found or return None if not found)

There are a lot of try/catches throughout this PR which are related to that.

backend/ng/containers/models/IndvidualContainer.py
RayofLightz: So I want a container doesn't exist error to make it to the frontend at least for restart because in that case they should do a recycle and I don't want the system to auto recycle on not existing I want the end user to know a recycle is happening. I could probably wrap the error in a business logic error

backend/ng/containers/models/ContainerInstance.py
RayofLightz: I have been scratching my head about this for the past few days and honestly before that. I had thought about potentially doing a Mixin to consolidate the structure between IndvidualContainer and ContainerInstance.  I originally decided against this because while it would save some boilerplate, it would make the code much harder to read in the long run and abstract how the function is behaving (Also best to avoid things in which multiple inheritance can be an issue). However I think creating a staticmethod for starting the container and potentially the network creation/attaching would be a good middle ground and actually make the code easier to read

fpayer: This could also be a separate util function that's more generic, so it doesn't need to be associated with a specific class.

backend/ng/containers/routes/admin_routes.py
```diff
@@ -0,0 +1,101 @@
...
+    @admin_container_namespace.doc(
+        description="Get docker node stats",
+        responses={
+            200: "Success",
+            400: "Bad request"
+        },
+    )
+    @admin_endpoint()
+    def get(self):
+        res = get_stats()
```
zoecough, line 46: where is this defined? i was having trouble finding it

backend/ng/core/middleware/error_handler.py
```diff
@@ -112,6 +112,7 @@ def handle_integrity_error(error):
     def handle_sqlalchemy_error(error):
         db.session.rollback()
         db.session.remove()
+        print(error)
```
zoecough, line 115: would there be any benefit to including the error in the error_response as well?

zoecough (changes requested):
check on the ruff linter failures, some of them deal with missing imports which seems significant

backend/ng/containers/routes/admin_routes.py
RayofLightz: Oh yippie this is fun something got messed up with the rebase to fix merge conflicts as far as I can tell

backend/ng/core/middleware/error_handler.py
RayofLightz: I had talked with Franz about this awhile back irrc it was going to get sent over to Carter or Daniel

RayofLightz:
Also I wanted to leave a comment about why I didn't write a get container util function. The reasoning for this was that if wrapped it and threw an error we would still have the try/except. And if we either returned the container or null it would be the same control structure just with an if/else. So I didn't see a benefit to creating a util function for that.

zoecough (approved):
looks good to me

PR #116, opened 2025-08-15
Permissions cleanup

droge91:
Switched everything to use our Role/Permission system instead of CTFD is_admin


backend/ng/permissions/controllers/get_user_roles.py
```diff
@@ -0,0 +1,16 @@
...
+
+def get_user_roles():
+    """
+    Get the current user's role.
+    """
+
+    user = get_current_user()
+    if not user:
+        return []
+    ng_user = User.find_or_create_by_ctfd_id(user.id, commit=False)
```
fpayer: Why `commit=False`?

droge91: I guess it should be true just in case. In theory it should never be making any database changes since you shouldn't be able to be logged in as someone with no associated ctfd user.

fpayer (approved):

PR #119, opened 2025-08-15
Support updated sync

CarterPerez-dev:


backend/ng/core/middleware/loaders/__init__.py
```diff
@@ -13,6 +13,7 @@
 from .load_user import load_user
 from .load_container_instance import load_container_instance
 from .load_indvidual_container_by_user import load_indvidual_container_by_user
+from .load_event_and_team_if_provided import load_event_and_team_if_provided
```
cjones-cs: I don't think we need a loader for this if it's a one off situation. Could this be done in the create controller itself instead of being abstracted?

backend/ng/core/middleware/loaders/load_event_and_team_if_provided.py
```diff
@@ -0,0 +1,59 @@
...
+from ...exceptions import NotFoundError
+
+
+def load_event_and_team_if_provided(
+    source: LoaderType, input_key="event_id", event_output_key="event", team_output_key="team"
+) -> Callable:
+    """
+    Load event and team if event_id is provided.
+    If no event_id, sets both event=None and team=None.
+    If event_id provided, loads event and finds user's team for that event.
```
cjones-cs: Do we actually need to load the whole event? The scope was to just find the team_id based on the event_id and current_user. IDK the specific backend magic to do that, but it's a simple query for the db

backend/ng/support/controllers/user_actions/create_ticket.py
```diff
@@ -6,7 +6,6 @@
 from ....user.models.User import User
 from ...models.Ticket import Ticket
 
-
 def create_ticket(
```
cjones-cs, line 11: Please don't commit things that aren't helpful like this. If there are other changes in this file, it would make sense, but otherwise it's clutter in the pr. `git checkout -p [filepath]` and `git diff [filepath]` is helpful for weeding these out.

backend/ng/support/models/Ticket.py
```diff
@@ -137,6 +139,7 @@ def serialize(self, include_admin_fields: bool = False) -> SerializedTicket:
             "event_id": self.event_id,
             "team_id": self.team_id,
             "challenge_id": self.challenge_id,
+            "challenge_name": self.challenge.name if self.challenge else None,
```
cjones-cs: If we are storing challenge name as none if the challenge isn't passed, then wouldn't the challenge id be the same?

backend/ng/support/tests/test_support_api.py
```diff
@@ -10,8 +10,11 @@
...
 
-    def test_create_ticket_basic(self, logged_in_client, user, event):
+    def test_create_ticket_basic(self, logged_in_client, user, event, team_factory):
         """Test creating a basic support ticket"""
+        # Create a team for the user in this event
+        team_factory(event=event, members=[user])
+
         response = logged_in_client.post(
             "/ng/support/tickets/create",
             json={
```
cjones-cs, line 20: The basic ticket only has text and message. Event_id is an optional field.

backend/ng/core/middleware/loaders/__init__.py
CarterPerez-dev: Yea you're probably right, this is probably a one off situation, I can just do it in the controller.

backend/ng/support/models/Ticket.py
cjones-cs: move this out of the model. Call in controller

cjones-cs (changes requested):
Left some review comments

CarterPerez-dev:
Getting the names in the route is better because the controllers can be used internally without forcing API specific data structure
and we should only enrich data when actually needed for an API response. Its also better to test controllers in isolation.

cjones-cs (approved):

PR #121, opened 2025-08-18
Frontend container management

trevnels-cs:
- Adds admin container management ("Deployments" tab)
    - Can see challenges deployed by teams along with statuses of each service within a deployment
    - Links to filtered deployment view from event/team/challenge entries in the admin panel
    - Buttons to restart/recycle individual service containers for a deployment
    - Button to tail logs from a service container from the admin panel
- Improvements to user-facing challenge page
    - "Connect Workspace" renamed to "Start Challenge"
    - Start challenge button now has a subtle animation to draw attention - first thing the user should see
    - Confirmation modal when switching from another challenge
    - Connected indicator shown on the user's current challenge
    - Future dropdown where restart/recycle controls will eventually go
- A few backend route/type fixes


backend/ng/containers/models/ContainerInstance.py
```diff
@@ -168,13 +169,19 @@ def get_service_group(cls, challenge_id: int, team_id: int) -> list[SerializedCo
     def get_instance_by_id(cls, instance_id: int):
         return cls.query.filter_by(id=instance_id).first()
 
+    def logs(self, tail: int = 500) -> str:
+        client = get_client(config.DOCKER_HOST)
+        ctr = client.containers.get(self.dockerid)
+        return ctr.logs(tail=tail).decode('utf-8')
+
```
RayofLightz: This needs to be in a try catch incase the container doesn't exist on the host.

RayofLightz: Catch on `docker.errors.NotFound`

backend/ng/containers/routes/admin_routes.py
```diff
@@ -17,7 +17,7 @@
 
 admin_container_namespace = Namespace("admin containers", description="admin containers")
 
-@admin_container_namespace.route("/")
+@admin_container_namespace.route("")
```
RayofLightz, line 20: I am assuming this was changed for consistency?

backend/ng/containers/models/ContainerInstance.py
```diff
@@ -168,13 +169,19 @@ def get_service_group(cls, challenge_id: int, team_id: int) -> list[SerializedCo
     def get_instance_by_id(cls, instance_id: int):
         return cls.query.filter_by(id=instance_id).first()
 
+    def logs(self, tail: int = 500) -> str:
```
RayofLightz: We might want to default to something like 200. 500 lines seems a tad high

frontend/src/routes/admin/deployments/DeploymentSidebar.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+      <Table.Cell>
+        <Tooltip content={<Text>{service?.dockerid}</Text>}>
+          <Code color="gray">{service?.dockerid.slice(0, 12)}</Code>
+        </Tooltip>
+      </Table.Cell>
+      <Table.Cell>
+        <Skeleton loading={!status}>
+          {status?.image || 'lorem-ipsum'}
+        </Skeleton>
```
RayofLightz: Move this into its own file

backend/ng/containers/routes/admin_routes.py
trevnels-cs: Yes, with the route set to `/`, a trailing slash is required in the URL when making a request to the endpoint which is not the case with empty string

frontend/src/components/EventCellRenderer.tsx
```diff
@@ -0,0 +1,25 @@
+import { EventIcon } from '@/constants';
+import { useAdminEvent } from '@/hooks/events';
+import { Skeleton } from '@radix-ui/themes';
+import Entity from './Entity';
+
+export default function EventCellRenderer({ value : eventId }: { value: number }) {
```
cjones-cs, line 6: You're only using this in one place, so it shouldn't go in /components. Things in /components should also not be fetching any data

frontend/src/routes/admin/deployments/DeploymentSidebar.tsx
```diff
@@ -0,0 +1,155 @@
...
+import RecycleContainerModal from './RecycleContainerModal';
+import RestartContainerModal from './RestartContainerModal';
+
+function ServiceRow({ service }: {service: ContainerInstance}) {
+  const { data : status, error, isLoading } = useContainerStatus(service.id);
+  return (
+    <Table.Row key={service.id}>
+      <Table.Cell>
+        <Skeleton loading={isLoading}>
+          {status?.name || error?.message || 'lorem-ipsum'}
```
cjones-cs: Remove fake text. Opt for generic default message instead like "Loading..." or something

frontend/src/routes/admin/deployments/DeploymentSidebar.tsx
```diff
@@ -0,0 +1,155 @@
...
+        </Skeleton>
+      </Table.Cell>
+      <Table.Cell>
+        <Skeleton loading={!status}>
+          <Badge color={status?.status === 'running' ? COLOR_POSITIVE : COLOR_NEGATIVE}>
+            {upperCase(status?.status) || 'lorem-ipsum'}
```
cjones-cs: fake text

frontend/src/routes/admin/deployments/RecycleContainerModal.tsx
```diff
@@ -0,0 +1,33 @@
+import { COLOR_NEGATIVE } from '@/constants';
+import { recycleContainer, useContainerStatus } from '@/hooks/container';
+import { Button, Text } from '@radix-ui/themes';
+import Modal from 'components/Modal';
+import { TbRecycle } from 'react-icons/tb';
+
+export default function RecycleContainerModal({ containerId }: {containerId: number}) {
```
cjones-cs, line 7: Probably overkill, but could use type ContainerInstance['id'] instead of just 'number'

frontend/src/components/EventCellRenderer.tsx
trevnels-cs: This is used in both the team grid and deployments grid, so I pulled it out into a separate component. Due to how the grid works, each cell has to perform the event id to name lookup and have them be deduplicated by SWR - otherwise all of these requests have to happen at once in a blocking manner instead of being done asynchronously as the grid scrolls

frontend/src/routes/admin/deployments/DeploymentSidebar.tsx
trevnels-cs: I can change this, but keep in mind that this text is never shown - it only exists to set the width of the skeleton component wrapping it while in the loading state.

frontend/src/routes/admin/deployments/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+import EventCellRenderer from 'components/EventCellRenderer';
+import DeploymentSidebar from './DeploymentSidebar';
+
+function TeamCellRenderer({ data }: {data: Deployment}) {
+  return <Entity icon={TeamIcon} label={data.team_name} to={`/admin/teams?id=${data.team}`} />;
+}
+
+function ChallengeCellRenderer({ data }: {data: Deployment}) {
+  return <Entity icon={ChallengeIcon} label={data.challenge_name} to={`/admin/events?id=${data.event_id}`} />;
+}
```
cjones-cs: You can actually condense this into one cellRenderer component by using `cellRendererParams`, then just call Entity with {icon, label, link} as params

`cellRenderer: EntityLinkCell,
      cellRendererParams: (params) => ({
        label: params.data.challenge_name, ...
      }),`

frontend/src/routes/events/challenge/ConnectModal.tsx
```diff
@@ -1,14 +1,15 @@
...
 import { useState } from 'react';
-import { TbCheck, TbPlug } from 'react-icons/tb';
+import { TbCheck, TbPlayerPlay } from 'react-icons/tb';
 
 export default function ConnectModal({ eventId, challengeId }: {
   eventId: number;
   challengeId: number;
 }) {
-  const currentChallenge = null; // get from server once route is ready
+  const { data : currentChallenge } = useCurrentChallengeId();
```
cjones-cs: are you handling the error case?

frontend/src/components/EventCellRenderer.tsx
trevnels-cs: Also, once the event name consistently comes down with event ids from the backend, this renderer will be removed - it's currently just a stopgap for getting event names into the table for routes where names are not provided.

cjones-cs: Okay, adding this to our todo list for frontend when backend gets cleaned a bit

cjones-cs (approved):

PR #122, opened 2025-08-18
initial notification models, placeholders, and model test

CarterPerez-dev:
- initial notification models, placeholders, and model test
- deleted the websocket folder to avoid confusion and clean up unused files


backend/ng/notifications/models/Notification.py
```diff
@@ -0,0 +1,334 @@
...
+        nullable = False
+    )
+    data = db.Column(db.JSON, nullable = True)
+
+    recipient_id = db.Column(db.Integer,
+                             db.ForeignKey("users.id"),
+                             nullable = True,
+                             index = True)  # None = broadcast
+    sender_id = db.Column(db.Integer, db.ForeignKey("users.id"), nullable = True)
+    read = db.Column(db.Boolean, default = False, nullable = False)
```
fpayer: This might be better as a DateTime instead of a Bool so we can track when notifications were read.

backend/ng/notifications/models/Notification.py
```diff
@@ -0,0 +1,334 @@
...
+        nullable = False
+    )
+    title = db.Column(db.String(config.NOTIFICATION_TITLE_MAX_LENGTH), nullable = False)
+    message = db.Column(
+        db.String(config.NOTIFICATION_MESSAGE_MAX_LENGTH),
+        nullable = False
+    )
+    data = db.Column(db.JSON, nullable = True)
+
+    recipient_id = db.Column(db.Integer,
```
fpayer: Recipient_id should be required. Otherwise, the "read" field wouldn't work since it wouldn't be clear who read the notification.

backend/ng/notifications/models/Notification.py
```diff
@@ -0,0 +1,334 @@
...
+        db.Enum(NotificationPriority),
+        default = NotificationPriority.NORMAL,
+        nullable = False
+    )
+    title = db.Column(db.String(config.NOTIFICATION_TITLE_MAX_LENGTH), nullable = False)
+    message = db.Column(
+        db.String(config.NOTIFICATION_MESSAGE_MAX_LENGTH),
+        nullable = False
+    )
+    data = db.Column(db.JSON, nullable = True)
```
fpayer: Do you have a use-case in mind for the data field? If not, we should remove it to shrink the scope down.

backend/ng/notifications/models/Notification.py
```diff
@@ -0,0 +1,334 @@
...
+
+    recipient_id = db.Column(db.Integer,
+                             db.ForeignKey("users.id"),
+                             nullable = True,
+                             index = True)  # None = broadcast
+    sender_id = db.Column(db.Integer, db.ForeignKey("users.id"), nullable = True)
+    read = db.Column(db.Boolean, default = False, nullable = False)
+    created_at = db.Column(db.DateTime, default = utc_now, nullable = False)
+
```
fpayer, line 82: Might be good to also have an `expires_at` field so we can automatically clear out old notifications.

fpayer (changes requested):
I think there's actually a need for 2 models:
- Announcement
- Notification

Announcement would be used to display a banner message on the site itself (e.g. telling people about some change in event times). Notification would be used to send a message to a particular individual (e.g. letting someone know their support ticket was updated).

backend/ng/notifications/models/Notification.py
CarterPerez-dev: Yea it's probably redundant

CarterPerez-dev:
I rethought  the priority field approach for notifications/announcements. Given that the vast majority will be just a 'standard' priority, I'm torn between three options: required, optional, or no priority field whatsoever. What's your take?

fpayer:
I don't have a strong feeling on it. Mainly because I'm not sure what the difference would be in how the notification will be presented to the user. Are you thinking the notifications would be different colors or some other difference?

fpayer (approved):

PR #123, opened 2025-08-19
admin route fixes + impersonation

droge91:
Fixed some of the old /admin routes that were nonfunctional and added the namespace back.
Added endpoints to allow admins to perform session based impersonation of users.


backend/ng/user/models/User.py
```diff
@@ -195,6 +195,15 @@ def delete(self) -> None:
         db.session.delete(self)
         db.session.commit()
 
+    @classmethod
```
fpayer, line 198: Any reason for adding this function? It doesn't seem to get called anywhere. It also looks like it has copy+paste comments from the Team model.

droge91: the get_data_counts controllers uses it, but it got removed in one of the refactors so I added it back so the associated route functioned again. The comment needs fixing though.

fpayer (approved):

PR #124, opened 2025-08-19
 new admin endpoint for retrieving attempts by team

droge91:


backend/ng/scoring/routes/admin_routes.py
```diff
@@ -83,3 +84,20 @@ def get(self, event_id: int, team_id: int, event, team, **kwargs):
...
+class ScoreAttempts(Resource):
+    @scoring_admin_namespace.doc(**GET_SCORE_HISTORY_DOC)
+    @admin_endpoint()
+    @load_event(LoaderType.PARAM)
+    @load_team(LoaderType.PARAM)
+    def get(self, event_id: int, team_id: int, event, team, **kwargs):
+        """
+        Get scoring attempts for a team
+        """
+        limit = request.args.get("limit", config.DEFAULT_SCORE_HISTORY_LIMIT, type=int)
```
trevnels-cs, line 98: If we include a limit param here, we should also have an offset param and return the total count along with the data so the frontend can paginate to ensure all attempts are still accessible if the number of them goes beyond MAX_SCORE_HISTORY_LIMIT

droge91:
gonna close this as carters branch adds a route that does a similar thing

PR #125, opened 2025-08-19
User Support page

cjones-cs:
Support Detail page:
- User can close (not reopen) ticket
- User can reply to ticket (only when ticket isn't closed)

Create Ticket Page:
- User must input a subject and message
- Event, and Challenge fields are optional
- Team is auto-filled based on event and user in the backend

Support Index page:
- Displays a table of the current user's tickets

### DIFF
--- a/frontend/src/routes/support/Detail.tsx
+++ b/frontend/src/routes/support/Detail.tsx
@@ -8,26 +8,26 @@ import {
   Heading,
   Separator,
   Text,
-  Section
+  Section,
 } from '@radix-ui/themes';
 import { StatusBadge } from 'components/StatusBadge';
 import { ErrorCallout } from 'components/Callouts';
 import { TbArrowLeft } from 'react-icons/tb';
 import { useNavigate, useParams } from 'react-router';
 import { isNil, isUndefined, map } from 'lodash';
-import RichTextEditor from 'components/RichTextEditor'
+import RichTextEditor from 'components/RichTextEditor';
 import { useMyTicketMessages, addNewTicketMessage, resolveMyTicket } from '@/hooks/support';
 
 export default function Detail() {
   const navigate = useNavigate();
   const { idTicket } = useParams();
-  const { data, error: errorMessages } = useMyTicketMessages(Number(idTicket));
-  const [version, setVersion] = useState<number>(0) // To reinit the RichTextEditor
-  const [newText, setNewText] = useState<string>('');
-  const [replyError, setReplyError] = useState<string | null>(null);
-  const [resolveError, setResolveError] = useState<boolean>(false);
-  const [resolveLoading, setResolveLoading] = useState<boolean>(false);
-  const [replyLoading, setReplyLoading] = useState<boolean>(false);
+  const { data, error : errorMessages } = useMyTicketMessages(Number(idTicket));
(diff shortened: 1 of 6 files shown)


frontend/src/routes/support/Detail.tsx
```diff
@@ -1,101 +1,174 @@
...
+          </Box>
           <div>
-            Status
-            <Separator size="4" />
-            {StatusBadge(status)}
+            {map(messages, (message) => (
+              <Card className="mt-2" key={message.id}>
+                <Flex justify="between">
+                  <Text weight="bold" size="2">{message.author_name}</Text>
+                  <Text weight="bold" size="2">{new Date(message.created_at).toString()}</Text>
```
trevnels-cs: Our SWR fetching logic will automatically parse dates coming from the backend, so `new Date(...)` is not needed here - `message.created_at` will already be a `Date` at this point.

Also, consider using `toLocaleString` instead of `toString` for a more human-readable date format

frontend/src/routes/support/Detail.tsx
```diff
@@ -1,101 +1,174 @@
...
+              </ErrorCallout>
+            )}
+          </Flex>
         </Flex>
-      </Card>
-    </Flex>
+
+        <Card size="3" className="w-2/7 h-fit">
+          <Flex direction="column" gap="4">
+            <Section size="1">
```
trevnels-cs, line 143: Consider hiding these sections if their value is not present instead of displaying N/A, I think that would make the UI a bit cleaner

frontend/src/routes/support/index.tsx
```diff
@@ -10,48 +10,47 @@ import type { ColDef } from 'ag-grid-community';
...
-    { field : 'status', cellRenderer : StatusBadgeCell },
-    { field : 'created', valueFormatter : (params) => new Date(params.value).toString() },
-    { field : 'updated', valueFormatter : (params) => new Date(params.value).toString() },
+  const { data, error } = useMyTickets();
+
+  const colDefs: ColDef<Ticket>[] = [
+    { field : 'subject', headerName : 'Subject' },
+    { field : 'event_name', headerName : 'Event Name' },
+    { field : 'status', headerName : 'Status', cellRenderer : StatusBadgeCell },
+    { field : 'opened_timestamp', headerName : 'Created Date', valueFormatter : (params) => new Date(params.value).toString() },
```
trevnels-cs: `new Date(...)` is also not needed here since params.value is already a `Date` for these columns - consider `params.value.toLocaleString()`

frontend/src/routes/support/Detail.tsx
cjones-cs: Generally, I'd agree, but not in this use case. In this case I want the user to aware of these optional values since they have a corresponding place on the index's table. I don't think it's good design to have a table, then a detail associated with that table record, that doesn't display all the data fields (or potentially more fields as needed)

frontend/src/routes/support/Detail.tsx
cjones-cs: Good call. fixed

frontend/src/routes/support/index.tsx
cjones-cs: good call, fixed

trevnels-cs (approved):

PR #126, opened 2025-08-20
challenge progress, submission history

CarterPerez-dev:
A lot of it is just updating documentation, and adding test cases, the relevant additions are:

- [backend/ng/event/controllers/user/get_challenge_progress.py](https://github.com/CyberSkyline/ctf-ng/compare/backend-TODOs-one-point-O?expand=1#diff-cfd2b2335d796fea5980da9dc8af23ab657b4d712baaa327f76b0d7e38a66177)
- [backend/ng/event/routes/user_routes.py](https://github.com/CyberSkyline/ctf-ng/compare/backend-TODOs-one-point-O?expand=1#diff-7b3991d69dec488af5992f00298135f95cbbb2cbda9c3c05d0fd812e4e06ad75)
- [backend/ng/scoring/controllers/admin_actions/get_submission_history.py](https://github.com/CyberSkyline/ctf-ng/compare/backend-TODOs-one-point-O?expand=1#diff-9fb30309f0eb0135717abd2cf2f15f001dd1bc097b3351c9e3ad3058ba490311)
- [backend/ng/scoring/routes/admin_routes.py](https://github.com/CyberSkyline/ctf-ng/compare/backend-TODOs-one-point-O?expand=1#diff-04bc8d20a43399f454c4da2ce93c9406fdde143a0137490a48dc196b50ad6b2d)

Doing more serialized ID - name correspondence next.


backend/ng/scoring/routes/admin_routes.py
```diff
@@ -77,9 +99,105 @@ def get(self, event_id: int, team_id: int, event, team, **kwargs):
...
+        result = get_submission_history(
+            team_id = team_id,
+            event_id = event_id,
+            limit = limit
+        )
+
+        enriched_score_events = []
+        for score_event in result["score_events"]:
+            event_data = score_event.serialize(include_admin_fields = True)
+            event_data["team_name"] = team.name
```
trevnels-cs: Avoid adding extra data after serialization - these fields should be added within the serialize function and be present on the serialized type definition

backend/ng/scoring/routes/admin_routes.py
```diff
@@ -77,9 +99,105 @@ def get(self, event_id: int, team_id: int, event, team, **kwargs):
...
+                event_data["source_id"] = score_event.hint_redemptions[0].id
+            elif score_event.manual_awards:
+                event_data["source_type"] = "manual_award"
+                event_data["source_id"] = score_event.manual_awards[0].id
+            enriched_score_events.append(event_data)
+
+        enriched_attempts = []
+        for attempt in result["attempts"]:
+            attempt_data = attempt.serialize(include_admin_fields = True)
+            attempt_data["team_name"] = team.name
```
trevnels-cs: Same thing - this data should be added as part of serialization and not afterwards

backend/ng/scoring/routes/admin_routes.py
```diff
@@ -77,9 +99,105 @@ def get(self, event_id: int, team_id: int, event, team, **kwargs):
...
+            if attempt.challenge:
+                attempt_data["challenge_name"] = attempt.challenge.name
+            if attempt.question:
+                attempt_data["question_name"] = attempt.question.name
+            enriched_attempts.append(attempt_data)
+
+        enriched_hints = []
+        for redemption in result["hint_redemptions"]:
+            hint_data = redemption.serialize(include_admin_fields = True)
+            hint_data["team_name"] = team.name
```
trevnels-cs: Same thing - this data should be added as part of serialization and not afterwards

backend/ng/scoring/routes/admin_routes.py
```diff
@@ -77,9 +99,105 @@ def get(self, event_id: int, team_id: int, event, team, **kwargs):
...
+                hint_data["user_name"] = user_data["name"]
+
+            if redemption.hint:
+                hint_data["hint_preview"] = redemption.hint.preview
+            enriched_hints.append(hint_data)
+
+        enriched_awards = []
+        for award in result["manual_awards"]:
+            award_data = award.serialize(include_admin_fields = True)
+            award_data["team_name"] = team.name
```
trevnels-cs: Same thing - this data should be added as part of serialization and not afterwards

backend/ng/scoring/controllers/admin_actions/get_submission_history.py
```diff
@@ -0,0 +1,48 @@
...
+    Attempt,
+    HintRedemption,
+    ManualPointAward,
+)
+
+
+def get_submission_history(
+    team_id: int,
+    event_id: int,
+    limit: int = config.DEFAULT_SUBMISSION_HISTORY_LIMIT
```
trevnels-cs: Having a limit here may result in some logical issues since there are multiple lists of varying lengths being returned - if an attempt from the attempts list references a score event that's outside of the requested limit, handling that on the frontend will get tricky.

I do think that the number of score events for a single team will be relatively low enough that a limit is not needed in this case, and all can be returned at once. @fpayer any thoughts?

backend/ng/scoring/controllers/admin_actions/get_submission_history.py
```diff
@@ -0,0 +1,48 @@
...
+from .... import config
+from ....scoring.models import (
+    ScoreEvent,
+    Attempt,
+    HintRedemption,
+    ManualPointAward,
+)
+
+
+def get_submission_history(
```
trevnels-cs: Is there a better name than submission history for this? Submission history makes me think this will only return attempts, even though there is other data present too.

backend/ng/scoring/routes/admin_routes.py
CarterPerez-dev: https://github.com/CyberSkyline/ctf-ng/pull/119/files/8ab81f6e3b2971127a707f51db0303ab1ec63d5e#r2279712144
@cjones-cs - Seems there may be some conflicting approaches being suggested for name enrichment.

- Enrichment not in the model, handle it in controller/route (which
   I've been doing for the support routes)
   
 - Put put name enrichment within the serialize() method 
 
So What do you both think would be the best consistent pattern going forward? Should we:
- Do all name enrichment to controller/route level, or
- Move all name enrichment into the serialize() methods?

trevnels-cs: My take is that if we're adding data to serialized models, it should be part of the underlying type definition - i.e. in this case `team_name` should be added to the `SerializedScoreEvent` type and populated as part of the model serialize method.
Adding fields after serialization seems a little more fragile - certain controllers/routes might end up adding fields while others don't (or if we forget to.) Since the goal is to have model return types be consistent across the entire backend, I think it makes sense to do as part of the model to ensure that anywhere the model is returned includes those fields. This also avoids having to copy/paste this logic across all controllers/routes where these models are serialized and returned.

tl;dr we shouldn't be "enriching" serialized model objects after the fact - if we want to add fields, we should be adding them during serialization to ensure consistency. Serialized model type signatures are effectively a contract with the frontend as to what values will be there - anything not present on the type signature doesn't exist as far as the frontend is concerned.

backend/ng/scoring/controllers/admin_actions/get_submission_history.py
CarterPerez-dev: - get_team_score_history?
- get_team_scoring_events?
- get_team_score_data?
- get_scoring_activity?

What do you recommend?

backend/ng/scoring/controllers/admin_actions/get_submission_history.py
trevnels-cs: Another idea to simplify this would be to serialize the `attempt`/`hint_redemption`/`manual_award` fields as part of the ScoreEvent - that way we can page through them without worrying about jagged lists and the frontend would not have to perform that joining logic.

backend/ng/scoring/routes/admin_routes.py
CarterPerez-dev: Yea I agree, since we will be doing it for a lot of endpoints, makes more sense to do it in the model/serialize function.

backend/ng/scoring/controllers/admin_actions/get_submission_history.py
fpayer: I agree - no need to add complexity with handling limits until we run into issues.

trevnels-cs:
FYI - swagger docs do not seem to load on this branch:
image

backend/ng/scoring/routes/admin_routes.py
```diff
@@ -114,90 +114,20 @@ def get(self, event_id: int, team_id: int, event, team, **kwargs):
...
-
-        return success_response(
-            {
-                "score_events": enriched_score_events,
-                "attempts": enriched_attempts,
-                "hint_redemptions": enriched_hints,
-                "manual_awards": enriched_awards
-            }
-        )
+        return success_response({"score_events": score_events})
```
trevnels-cs: May be cleaner to return the list directly, doesn't need to be nested within another key

CarterPerez-dev:
> FYI - swagger docs do not seem to load on this branch:
> 
> image
> 
> 

It could be that example for the get team score events doc is causing issues, I'll figure it out.

backend/ng/scoring/models/HintRedemption.py
```diff
@@ -25,6 +25,10 @@ class SerializedHintRedemption(TypedDict):
     score_event_id: int | None
     timestamp: str
     points: int
+    # Name enrichment fields
```
trevnels-cs, line 32: Would it be possible to add the challenge id/name to this as well?

trevnels-cs:
Actually there is another issue with this approach - incorrect attempts don't have a score event attached to them, so they aren't included in this log. We might need to think about this a little more

I think the best approach for this would just be to have `/attempts`, `/hint_redemptions`, and `/manual_awards` routes that an admin can hit to get attempt, hint redemption, and manual award objects for a team without needing them to be attached to a score event

trevnels-cs (approved):

PR #129, opened 2025-08-21
Move name serialization to model serialization function

CarterPerez-dev:


fpayer:
Will defer to how frontend wants this, but code looks fine to me

backend/ng/support/models/Ticket.py
```diff
@@ -155,6 +169,11 @@ def serialize(self, include_admin_fields: bool = False) -> SerializedTicket:
                 }
             )
 
+            if self.author:
```
trevnels-cs: I don't think author name needs to be admin-only

backend/ng/support/models/Ticket.py
```diff
@@ -30,11 +30,18 @@ class SerializedTicket(TypedDict):
     challenge_id: int | None
     message_count: int
     tags: list[str]
```
cjones-cs: Tags needs to be admin-only. Regular users won't need it

backend/ng/support/models/Ticket.py
```diff
@@ -30,11 +30,18 @@ class SerializedTicket(TypedDict):
...
+    event_name: NotRequired[str | None]
+    team_name: NotRequired[str | None]
+    challenge_name: NotRequired[str | None]
     # Admin-only fields
     assigned_to: NotRequired[int | None]
     closed_timestamp: NotRequired[str | None]
     muted: NotRequired[bool]
     first_admin_response_timestamp: NotRequired[str | None]
+    # Admin name enrichment fields
+    author_name: NotRequired[str]
```
cjones-cs: Author_name should have the same permission level as author_id, so non-admin

cjones-cs (changes requested):
Small adjustment to admin vs regular fields

cjones-cs (approved):

PR #130, opened 2025-08-21
MINOR OAuth2 integration with Okta

zoecough:
CTFd is really picky about how you log in a user, and it won't recognize a user as logged in without the nonce and hash, so I had to make a placeholder hash since OAuth users won't have passwords.


zoecough:
https://dev-presidentscup.cisa.gov/ng/authenticate/okta/login

^ click testable here; you should be able to make a user (don't upload any personal information; when asked for images on the account creation flow put in any random image you might have and a fake SSN but you'll need a working phone number that can receive texts)

Currently redirects you to /hello

fpayer:
I'm getting a 404 with that link

conf/nginx/http.conf
fpayer, line 1: We should create a separate `http.conf` for dev/prod. Otherwise, the devs will get errors trying to run locally since they won't have the ssl cert.

zoecough:
https://dev-presidentscup.cisa.gov/ng/authenticate/okta/login

The link should be working now; env files will need to be configured with blank values for the oauth flow once this goes live. i can add to the install script to generate those lines when the env files are generated; not sure what's best to do when those files already exist to get those lines added for everybody. it'll just look like this:
```
# OAuth
OKTA_CLIENT_ID = -
OKTA_CLIENT_SECRET = -
OKTA_DOMAIN = -
SERVER_DOMAIN = -
```

conf/nginx/http.conf
```diff
@@ -18,6 +18,8 @@ http {
 
   server {
     listen 80;
+    return 301 https://$host$request_uri;
```
fpayer: Should there be this redirect to https?

docker-compose.yaml
```diff
@@ -26,15 +26,15 @@ services:
...
+    # command : -c "exec bash"                                                    
     command : -c "/usr/bin/supervisord -c /etc/supervisor/supervisord.conf && supervisorctl tail -f ctfd"
 
   nginx:
     image: nginx:1.28
     container_name: ng-nginx
     restart: always
     volumes:
-      - ./conf/nginx/http.conf:/etc/nginx/nginx.conf:ro
+      - ./conf/nginx/staging-http.conf:/etc/nginx/nginx.conf:ro
```
fpayer: Is this the correct filename?

fpayer (approved):

PR #132, opened 2025-08-22
add two endpoints for roles

CarterPerez-dev:
Screenshot_20250822_065254
Screenshot_20250822_064815

### DIFF
--- a/backend/ng/permissions/controllers/get_users_by_roles.py
+++ b/backend/ng/permissions/controllers/get_users_by_roles.py
@@ -30,7 +30,7 @@ def get_users_with_roles(role_names: list[str]) -> list[User]:
     return list(users)
 
 
-def get_assignable_users() -> list[User]:
+def get_support_role_users() -> list[User]:
     """
     Get all users who can be assigned support tickets
     """
--- a/backend/ng/permissions/controllers/get_users_by_roles.py
+++ b/backend/ng/permissions/controllers/get_users_by_roles.py
@@ -1,0 +1,40 @@
+"""
+Get users filtered by their roles - part of the permissions system.
+"""
+
+from ...core.exceptions import ValidationError
+
+from ..models.Role import Role
+from ..models.enums import RoleEnum
+from ...user.models.User import User
+
+
+def get_users_with_roles(role_names: list[str]) -> list[User]:
+    """
+    Get all users who have any of the specified roles
+    """
+    users = set()
+
+    for role_name in role_names:
+        try:
+            role_enum = RoleEnum(role_name.lower())
+        except ValueError:
+            valid_roles = [role.value for role in RoleEnum]
+            raise ValidationError(
+                    f"Invalid role '{role_name}'. Valid roles are: {', '.join(valid_roles)}"
+                    ) from None
+
+        role_users = Role.get_users_with_role(role_enum)
+        users.update(role_users)
+
+    return list(users)
+
+
+def get_assignable_users() -> list[User]:
+    """
+    Get all users who can be assigned support tickets
+    """
+    return get_users_with_roles(
+            [RoleEnum.ADMIN.value,
+             RoleEnum.SUPPORT.value]
+            )
--- a/backend/ng/permissions/routes/admin_routes.py
+++ b/backend/ng/permissions/routes/admin_routes.py
@@ -6,7 +6,7 @@ from ...core.middleware.auth import admin_endpoint
 from ...core.exceptions import ValidationError
 from ..models.Role import Role
 from ..models.UserRole import UserRole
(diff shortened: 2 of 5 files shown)


CarterPerez-dev:
Could've done empty params = all assignable users, but went with explicit 
endpoints for clarity, but what do ya'll think / prefer?

droge91 (commented):
This all looks fine, only thing I'd note is that /assignable_users is maybe a bit vague, including ticket in the route name might be better to specify exactly what the route does. There's documentation for it though so maybe it doesn't matter

CarterPerez-dev:
> This all looks fine, only thing I'd note is that /assignable_users is maybe a bit vague, including ticket in the route name might be better to specify exactly what the route does. There's documentation for it though so maybe it doesn't matter

Yea you're probably right, I'll add ticket.

backend/ng/permissions/controllers/get_users_by_roles.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    users = set()
+
+    for role_name in role_names:
+        try:
+            role_enum = RoleEnum(role_name.lower())
+        except ValueError:
+            valid_roles = [role.value for role in RoleEnum]
+            raise ValidationError(
+                    f"Invalid role '{role_name}'. Valid roles are: {', '.join(valid_roles)}"
```
cjones-cs, line 24: Should we be sending back what valid roles are? Wouldn't that be a potential security bug since roles are a security level item?

backend/ng/permissions/controllers/get_users_by_roles.py
```diff
@@ -0,0 +1,40 @@
...
+                    ) from None
+
+        role_users = Role.get_users_with_role(role_enum)
+        users.update(role_users)
+
+    return list(users)
+
+
+def get_assignable_users() -> list[User]:
```
cjones-cs: I agree that the name here is vague. Documentation is nice, but maybe something like "get_support_admin_users" or something would be better? We can brainstorm this

cjones-cs: get_support_role_users

backend/ng/permissions/tests/test_users_by_roles.py
```diff
@@ -0,0 +1,963 @@
...
+
+from ..controllers.get_users_by_roles import get_users_with_roles
+from ..controllers.assign_role_to_user import assign_role_to_user
+from ..controllers.get_users_by_roles import get_assignable_users
+
+
+pytestmark = pytest.mark.db
+
+
+class TestUsersByRolesHappyPaths:
```
cjones-cs, line 22: Is "happyPaths" supposed to be in the name?

cjones-cs (changes requested):
Just a minor renaming thing for clarity when debugging

backend/ng/permissions/controllers/get_users_by_roles.py
CarterPerez-dev: Not necessarily, I understand your point, but if a user is not an admin, they would first see the error messgae "Unauthorized", so only admins would ever see that message, I think it's just better UX since it will tell the admins which roles are available if they don't know.

backend/ng/permissions/tests/test_users_by_roles.py
CarterPerez-dev: It's just a class which essentially categorizes what I test in that class of tests. Its means:

"A "happy path" is the ideal or optimal scenario in which a user or system follows the intended, expected steps to achieve a goal without errors or exceptional conditions. It represents a frictionless, successful journey, often used in user experience (UX) design to create smooth user flows and in software testing to verify core functionality."

backend/ng/permissions/routes/admin_routes.py
```diff
@@ -174,8 +174,8 @@ def get(self, **kwargs):
         return success_response(users)
 
 
-@permissions_admin_namespace.route("/tickets/assignable_users")
-class AssignableUsers(Resource):
+@permissions_admin_namespace.route("/support_role_users")
+class SupportRoleUsers(Resource):
```
cjones-cs, line 178: Does this have a _docs.py associated with it? Since this is a new route, I don't see a routes/_docs.py file that goes with it?

CarterPerez-dev, line 178: When I make routes, I put all the namespace doc decorator documentation in a seperate file because I like cleaner looking routes and one central place to edit/write the documentation, but you can write them in the route decorator itself too,  it's just a style/preference thing. So the docs are written in the route for these permission routes.

cjones-cs (approved):

PR #133, opened 2025-08-23
Challenge Permissions + Model Changes

droge91:
Updated Team model to use can_manage_team instead of locked
Updated Event model to include hints_enabled and a toggle

Expended permissions middleware to include can_view_challenges and can_play_challenges
Adds new permissions checks in relevant routes.


fpayer (commented):
What are your thoughts about creating a `check_permission` middleware that would handle this `if` in the middleware instead?

e.g.
```
@get_permissions
@check_permission(PermissionEnum.CAN_VIEW_CHALLENGES, error_message = "...")
def get(self, ...)
  ...
```

We could set default error messages for each permission within `check_permission`

This would ensure consistency with the return error code/message and cut down on the repeated `return error_response...`

droge91:
> What are your thoughts about creating a `check_permission` middleware that would handle this `if` in the middleware instead?
> 
> e.g.
> 
> ```
> @get_permissions
> @check_permission(PermissionEnum.CAN_VIEW_CHALLENGES, error_message = "...")
> def get(self, ...)
>   ...
> ```
> 
> We could set default error messages for each permission within `check_permission`
> 
> This would ensure consistency with the return error code/message and cut down on the repeated `return error_response...`

Yeah I like this idea, would probably consolidate the permission getting and checking into one middleware in that case though.
The one issue is that it makes some of the error_responses pretty generic, like an action failing because the event is over or locked would just give whatever generic error message you give instead of specifying the exact reason

fpayer (approved):

PR #134, opened 2025-08-25
update support

CarterPerez-dev:


backend/ng/support/routes/_docs.py
```diff
@@ -356,12 +356,12 @@
...
+            "required": True,
             "type": "integer",
             "example": 1
         },
         "team_id": {
-            "description": "Team ID to associate with ticket",
+            "description": "Team ID to associate with ticket (optional - auto-derived from ticket author's team if not provided)",
             "required": False,
             "type": "integer",
             "example": 42
```
cjones-cs: I would just remove team_id from the params since we can always query it directly.

backend/ng/support/routes/admin_routes.py
```diff
@@ -226,8 +226,8 @@ def put(self, ticket_id: int, ticket, current_user: User, json_data, **kwargs):
         """
         updated_ticket = set_ticket_event(
             event_id=json_data.get("event_id"),
-            team_id=json_data.get("team_id"),
             ticket=ticket,
+            team_id=json_data.get("team_id"),
```
cjones-cs: same here, team_id doesn't necessarily need to be passed since we will always know the value based on the author of the ticket we can derive from ticketId/Ticket object

cjones-cs (changes requested):
We can probs remove team id from endpoint params since we would know that based on ticket id and event id

backend/ng/support/routes/_docs.py
CarterPerez-dev: true

cjones-cs (approved):

PR #135, opened 2025-08-26
Hints need names

johnmarsden-edu:
- Makes hints require a name so that we can update them later.


lib/parser/src/cyber_skyline/chall_parser/compose/challenge_info.py
```diff
@@ -60,6 +60,7 @@ class Hint:
     Hints can be either structured (TextHint) or simple strings.
     Each hint has a preview and costs points when opened.
     """
+    name: str = attr.ib(validator=v.instance_of(str))  # Developer facing name for the hint (e.g., "hint1", "hint2")
```
johnmarsden-edu, line 62: The hints field needs a uniqueness constraint on the names.

johnmarsden-edu: Same with the questions field

PR #136, opened 2025-08-26
Admin grid improvements

trevnels-cs:
- Consistently use entity names for column fields/filters now that they are provided by backend
- Configure grids to use radix UI font for cells
- Handle user `role` -> `roles` transition


frontend/src/routes/admin/events/index.tsx
```diff
@@ -1,88 +1,89 @@
...
+    filter : true,
+    floatingFilter : true,
+  },
+  {
+    field : 'id',
+    width : 100,
+    headerName : 'ID',
+    filter : true,
+    floatingFilter : true,
+  },
```
cjones-cs: If you don't actually need to display the id of the event, you can use `hide: true`, otherwise, it may be better to have this field near the leftmost side of the table

frontend/src/routes/admin/teams/index.tsx
```diff
@@ -1,12 +1,56 @@
...
+    cellRendererParams : (params: { data: {event_name?: string, event_id: number} }) => ({
+      icon : EventIcon,
+      label : params.data.event_name ?? `UNKNOWN (${params.data.event_id})`,
+      to : `/admin/events?id=${params.data.event_id}`,
+    }),
+  },
+  {
+    headerName : 'Members',
+    width : 100,
+    field : 'member_count',
```
cjones-cs: nit, but I would put field as the first item, just so it's consistent with other usages.

frontend/src/routes/admin/users/index.tsx
```diff
@@ -1,54 +1,60 @@
...
+    filter : true,
+    floatingFilter : true,
+  },
+  {
+    field : 'id',
+    width : 100,
+    headerName : 'ID',
+    filter : true,
+    floatingFilter : true,
+  },
```
cjones-cs: same thing here about the ids

frontend/src/components/AdminDataList.tsx
```diff
@@ -21,6 +21,14 @@ export default function AdminDataList({ data }: {data: Record<string, unknown>})
               if (key === 'role') {
                 return <RoleBadge value={value!.toString()} />;
               }
+              if (key === 'roles') {
+                return (value as string[]).map((role) => (
+                  <>
+                    <RoleBadge key={role} value={role} />
+                    &nbsp;
```
cjones-cs, line 28: Technically, wouldn't this have a trailing space after the last RoleBadge? There's probably a smart css thing we could do here to eliminate the extra space, but don't worry about it if it's too much effort.

cjones-cs (approved):

frontend/src/components/AdminDataList.tsx
trevnels-cs: It's difficult to use flex etc. in grid cells without having to redo a lot of the alignment styling ourselves, so I think spaces are the simplest approach while keeping the cell contents inline

PR #138, opened 2025-08-26
simple test runner

zoecough:


.github/workflows/chall-check-testrunner.yml
```diff
@@ -0,0 +1,29 @@
+name: chall-check test runner
+
+on:
+  pull_request:
+    paths:
+      - 'tools/chall-check/**'
```
johnmarsden-edu, line 6: This should also be applied to the lib/parser folder.

.github/workflows/chall-check-testrunner.yml
```diff
@@ -0,0 +1,29 @@
...
+          python-version: '3.12'
+
+      - name: Install dependencies
+        run: |
+          cd tools/chall-check
+          pip install -e . pytest
+
+      - name: Run tests
+        working-directory: tools/chall-check
+        run: python -m pytest test/test_cli.py -v
```
johnmarsden-edu: Is there not a way to run all tests available at the same time?

johnmarsden-edu (changes requested):

johnmarsden-edu (approved):
Looks good, looks like it runs just fine

PR #139, opened 2025-08-26
Staging deploy workflow

zoecough:
PR for testing


zoecough:
this runner implementation uses my Amazon WorkSpaces instance to host the runner since the runner has to be within the subnet our EC2 instances are on. I can create a written guide to set this up on other WorkSpaces instances if mine has to be deleted for whatever reason

zoecough:
because we've switched from Workspaces to a VPN connection, this is no longer relevant

PR #141, opened 2025-08-27
Validate serialized event instead of model

trevnels-cs:
Fix event time range update operations - would previously fail since validator expects serialized data (datetimes as strings), not model data (datetimes as datetime objects)

### DIFF
--- a/backend/ng/event/models/Event.py
+++ b/backend/ng/event/models/Event.py
@@ -204,7 +204,7 @@ class Event(db.Model):
             if hasattr(self, key):
                 setattr(self, key, value)
 
-        Event.validate(kwargs)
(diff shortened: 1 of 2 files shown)


trevnels-cs:
Noticed a subtle logical error with the time range validation here unrelated to the contents of this PR - once an event has started, it can no longer be updated, since sending a past start date with the PUT request will fail validation (because submitted start dates must be in the future.) Currently the only way to update an event in progress is to move the start date to some time in the future. I believe this also applies to the registration window as well.

I think it would make sense to allow submitting past dates within these ranges to resolve this. Setting (or keeping) a past date would just mean registration and/or the event would be opened if within the defined window.

fpayer (approved):

PR #145, opened 2025-08-28
Ctfd version upgrade

fpayer:


johnmarsden-edu:
@fpayer updated version now out.

PR #149, opened 2025-08-29
Feature/admin support

cjones-cs:
Admin Support Tickets table and sidebar functionality.

### DIFF
--- a/frontend/src/components/StatusBadge.tsx
+++ b/frontend/src/components/StatusBadge.tsx
@@ -1,14 +1,12 @@
-import { COLOR_INFO, COLOR_POSITIVE, COLOR_WARNING } from '@/constants';
+import { COLOR_POSITIVE, COLOR_WARNING } from '@/constants';
 import { Badge } from '@radix-ui/themes';
 import type { ICellRendererParams } from 'ag-grid-community';
 
 const CLOSED = 'closed';
-const IN_PROGRESS = 'inprogress';
 
 function StatusBadge(status: string) {
   switch (status) {
     case CLOSED: return <Badge color={COLOR_POSITIVE}>Closed</Badge>;
-    case IN_PROGRESS: return <Badge color={COLOR_INFO}>In Progress</Badge>;
     default: return <Badge color={COLOR_WARNING}>Open</Badge>;
   }
 }
@@ -16,7 +14,6 @@ function StatusBadge(status: string) {
 function StatusBadgeCell({ value }: ICellRendererParams) {
   switch (value) {
     case CLOSED: return <Badge color={COLOR_POSITIVE}>Closed</Badge>;
-    case IN_PROGRESS: return <Badge color={COLOR_INFO}>In Progress</Badge>;
     default: return <Badge color={COLOR_WARNING}>Open</Badge>;
   }
 }
--- a/frontend/src/components/StatusBadge.tsx
+++ b/frontend/src/components/StatusBadge.tsx
@@ -4,8 +4,7 @@ import type { ICellRendererParams } from 'ag-grid-community';
 
 const CLOSED = 'closed';
 
-function StatusBadge({status, size = "1"} : {status: string, size?: string}) {
-  console.log('size', size)
+function StatusBadge({ status, size = '1' } : {status: string, size?: string}) {
   switch (status) {
     case CLOSED: return <Badge color={COLOR_POSITIVE} size={size}>Closed</Badge>;
     default: return <Badge color={COLOR_WARNING} size={size}>Open</Badge>;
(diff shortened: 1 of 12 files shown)


frontend/src/hooks/users.ts
```diff
@@ -28,3 +28,13 @@ export function useUser(userId: number | null) {
 export function useCurrentUser() {
   return useSWR<User>('/users/me');
 }
+
+/**
+ * Admin Only endpoint
+ * Get events for a specific user
+ */
+export function useUserEvents(userId: number | undefined) {
```
trevnels-cs, line 36: This hook already exists in `events.ts`

frontend/src/routes/admin/tickets/index.tsx
```diff
@@ -1,81 +1,153 @@
...
         columnDefs={colDefs}
+        loading={isLoading}
+        getRowId={(params) => params.data.id.toString()}
+        sidebarComponent={MessagesSidebar}
         gridOptions={{
-          rowSelection : {
-            mode : 'singleRow',
-            checkboxes : false,
-            enableClickSelection : true,
+          onCellClicked : (e: CellClickedEvent) => {
```
trevnels-cs: Might be worth moving this into AdminGrid and having it take stopCellSelection as a prop so this logic can be more easily reused across other grids

frontend/src/components/StatusBadge.tsx
```diff
@@ -1,22 +1,19 @@
...
   switch (status) {
-    case CLOSED: return <Badge color={COLOR_POSITIVE}>Closed</Badge>;
-    case IN_PROGRESS: return <Badge color={COLOR_INFO}>In Progress</Badge>;
-    default: return <Badge color={COLOR_WARNING}>Open</Badge>;
+    case CLOSED: return <Badge color={COLOR_POSITIVE} size={size}>Closed</Badge>;
+    default: return <Badge color={COLOR_WARNING} size={size}>Open</Badge>;
   }
 }
 
 function StatusBadgeCell({ value }: ICellRendererParams) {
```
trevnels-cs, line 14: Could the StatusBadge component be used as the cell renderer instead of having a separate component? Might be better for consistency

frontend/src/routes/admin/tickets/index.tsx
```diff
@@ -1,81 +1,153 @@
...
-import { AgGridReact } from 'ag-grid-react';
-import { radixTheme } from '@/grid';
+  ChallengeIcon,
+  EventIcon,
+  TeamIcon,
+  UserIcon,
+} from '@/constants';
+import MessagesSidebar from './MessagesSidebar';
+
+function NameLinkCell(
```
trevnels-cs, line 18: This is a nice way of handling entity cells and empty values, I may eventually update Entity to have this logic so it can more easily be used as a renderer directly

frontend/src/components/StatusBadge.tsx
cjones-cs: Probably, but I did this to standardize the color usage and naming display across all usages of badging for statuses. It probably made more sense when I thought there were going to be 3 options instead of 2, but I think it's fine to leave it

frontend/src/hooks/users.ts
cjones-cs: Should we move it to the users section here? I've been grouping things based on their api documentation and route since that makes it easier to trace

frontend/src/routes/admin/tickets/index.tsx
cjones-cs: I agree, this undefined/isNull check would be a good thing to have to safeguard the ui from a potential broken link. Thankfully, in the usages I know of, we are guaranteed to have those values available. I had asked about that on one of your previous prs

frontend/src/routes/admin/tickets/index.tsx
cjones-cs: Done

frontend/src/components/StatusBadge.tsx
trevnels-cs: My concern is that since there's two separate implementations of the StatusBadge, one for normal UI and one for grids, they could potentially get out of sync

frontend/src/hooks/users.ts
trevnels-cs: I initially placed it in `events.ts` since it returns a list of events, but your interpretation also makes sense

cjones-cs: We will have a follow up pr to handle the dups and organize them by endpoint name

trevnels-cs (approved):

PR #150, opened 2025-08-29
Feature/admin support tags

cjones-cs:
Admin Tag management. Support tags are the only ones currently available, but we have the option of adding additional tag types


frontend/src/routes/admin/tags/SupportTags/TagModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+) {
+  const handleSubmit = async (data: FormData) => {
+    const entries = Object.fromEntries(data.entries());
+    const newTag: Omit<TicketTag, 'id | ticket_count'> = {
+      name : entries.name as string,
+      color : entries.color as string,
+      description : entries.description as string,
+    };
+
+    if (isUndefined(defaultValues)) {
```
trevnels-cs: Consider pulling isUndefined(defaultValues) into a variable like `isCreating` since it's used a lot here

trevnels-cs (approved):

PR #151, opened 2025-08-29
Miscellaneous event/challenge/dashboard improvements

trevnels-cs:
- Progress indicator on challenge cards, some display fixes
- Show "Teams of 2" instead of "Teams of 2-2" on team size 2 event headers
- Don't show the team tab if the URL tab parameter is changed to `team` in an individual event
- Disable browser autocompletion for answer fields
- Refactor user dashboard page for better loading/error states and reduced prop-drilling

### DIFF
--- a/frontend/src/components/EventHeader.tsx
+++ b/frontend/src/components/EventHeader.tsx
@@ -1,17 +1,17 @@
+import { DATEFORMAT } from '@/constants';
+import type { Event } from '@/types';
 import {
   AspectRatio,
   Box,
-  Link as LinkTheme,
   Flex,
   Heading,
+  Link as LinkTheme,
   Text,
 } from '@radix-ui/themes';
 import { isNull } from 'lodash';
 import type { ReactNode } from 'react';
 import { TbCalendar, TbUser } from 'react-icons/tb';
 import { Link } from 'react-router';
-import type { Event } from '@/types';
-import { DATEFORMAT } from '@/constants';
 import EventBadge from './EventBadge';
 import RadixMarkdown from './RadixMarkdown';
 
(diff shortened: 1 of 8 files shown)


frontend/src/components/EventHeader.tsx
```diff
@@ -1,17 +1,17 @@
+import { DATEFORMAT } from '@/constants';
```
cjones-cs, line 1: Since we are using toLocaleString for dates, I do wonder if this is still necessary.

frontend/src/routes/dashboard/index.tsx
```diff
@@ -24,30 +25,27 @@ export default function Dashboard() {
   return (
     <>
       <HeaderContainer>
-        {liveEvents.length === 0 && (
-          <InfoCallout>
-            No events are currently running. This should eventually be something more interesting, i.e. the first upcoming event or practice area.
-          </InfoCallout>
+        {(liveEvents === undefined || liveEvents.length === 0) && (
```
cjones-cs: You can use lodash isEmpty for this entire check

frontend/src/components/EventHeader.tsx
trevnels-cs: Since this is used for `formatRange`, toLocaleString doesn't apply here. However, we can remove the params from `new Intl.DateTimeFormat` to format date ranges according to the user's selected locale, which is the equivalent of toLocaleString

cjones-cs (changes requested):
Few ideas

frontend/src/components/EventHeader.tsx
trevnels-cs: I think the `DATEFORMAT.range` config we have set is still important here though for user-facing date ranges

cjones-cs (approved):

PR #156, opened 2025-09-03
Migrated from yarn to pnpm

fpayer:


fpayer:
To make the switch from yarn to pnpm:

```
corepack enable pnpm
pnpm install
```

backend/.scripts/pip_install.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source venv/bin/activate
```
zoecough, line 5: on my machine, this is executing before the venv is created, causing `install.sh` to error out.

image

trevnels-cs (approved):
`pnpm` seems to be fully working here. I did have to force delete the old node_modules beforehand due to some permission issues with `.vite` (likely related to the container mount)

PR #159, opened 2025-09-04
Consistent Docs and other TODO's

droge91:
Changes all route documentation to be consistently Inline
Adds a Start Team endpoint
Adds time_limit to Event and end_time to Team and updates the permission checks
Adds a CAN_START_TEAM_TIMER permission for captains


backend/ng/event/routes/user_routes.py
```diff
@@ -340,10 +431,31 @@ def post(self, event_id, event, team, current_user, **kwargs):
...
+            200: "Team started successfully",
+            403: "Forbidden if user does not have permission to start the event",
+            404: "Not Found if event or team does not exist",
+        },
+    )
+    def post(self, event_id, event, team, permissions, **kwargs):
+        """
+        Start the event for the user's team
+        """
+        team.set_start_timestamp(utc_now())
```
fpayer, line 452: Should have some sort of server-side check that prevents a team from calling this endpoint multiple times to extend their start/end time.

fpayer (changes requested):

backend/ng/event/routes/user_routes.py
droge91: This is handled by the permission check

droge91: The CAN_START_TEAM_TIMER permission will not be granted if a team has already started

fpayer (approved):

PR #160, opened 2025-09-05
ecr-pull + container reboot/recycle

RayofLightz:
Container reboot recycle to face the user. Foundation for pulling from a private container repo


backend/ng/event/routes/user_routes.py
```diff
@@ -419,3 +421,44 @@ def get(self, team: Team, current_user: User, challenge_id: int, event_id: int,
...
+        },
+        responses={
+            200: "Sucess",
+            400: "Bad request",
+        },
+    )
+    @user_endpoint()
+    @load_event(source=LoaderType.PARAM)
+    @load_team_by_user_and_event()
+    def get(self, team: Team, current_user: User, challenge_id: int, event_id: int, event: Event):
```
trevnels-cs: I believe these should be POST requests instead of GETs to indicate they're performing an action on the server instead of just retrieving data. Same goes for `EventChallengeStartContainers` above

trevnels-cs (approved):

PR #161, opened 2025-09-07
Notifications pr 3/3

CarterPerez-dev:


backend/ng/notifications/controllers/user_actions/mark_read.py
```diff
@@ -0,0 +1,35 @@
...
+) -> Notification:
+    """
+    Mark a single notification as read.
+    """
+    notification = Notification.find_by_id(notification_id)
+
+    if not notification:
+        raise NotFoundError(f"Notification {notification_id} not found")
+
+    if notification.recipient_id != user_id:
```
fpayer: Should use Daniel's permission check middleware system

backend/ng/notifications/controllers/user_actions/mark_read.py
```diff
@@ -0,0 +1,35 @@
...
+def mark_notification_read(
+    notification_id: int,
+    user_id: int,
+) -> Notification:
+    """
+    Mark a single notification as read.
+    """
+    notification = Notification.find_by_id(notification_id)
+
+    if not notification:
```
fpayer: Can we handle this check via a loader middleware for notifications?

backend/ng/notifications/routes/_docs.py
fpayer, line 1: Switch to inline docs

backend/ng/notifications/routes/user_routes.py
```diff
@@ -1,6 +1,104 @@
...
+        Get unread notification count
+        """
+        count = get_unread_count(user_id = current_user.id)
+        return success_response({"count": count})
+
+
+@notifications_user_namespace.route("/me/<int:notification_id>/read")
+class MarkRead(Resource):
+    @user_endpoint()
+    @notifications_user_namespace.doc(**MARK_NOTIFICATION_READ_DOC)
```
fpayer: Create a loader middleware for notifications

backend/ng/notifications/services/notification_service.py
```diff
@@ -1,6 +1,320 @@
...
+        user_ids = None,
+        team_id = None,
+        event_id = None
+    ):
+        """
+        Notify frontend to refetch data at path
+        """
+        try:
+            if user_ids:
+                emit_to_users("refetch", {"path": path}, user_ids)
```
fpayer: Use an enum for the event names to help prevent potential future issues with typos.

fpayer (changes requested):

backend/ng/notifications/controllers/user_actions/mark_read.py
CarterPerez-dev: Yeah, so my opinion is that this is more of an ownership check rather than a permission check,  it's similar to load_ticket_with_user checking if the user owns the ticket. We could use @get_permissions to check if user has a "manage_own_notifications" permission, but that seems like overkill for basic ownership validation. I did make it all happen in the loader middleware which seems cleaner but Let me know what you think.

fpayer: Ah, I see your point. Yeah, it's probably overkill to have a `manage_own_notifications` permission, but this is still a type of check that should happen in middlware. 

What about a `check_user_is_notification_recipient` middleware that just handles this `notification.recipient_id != user_id` check? That would help maintain the paradigm of permissions checks happening in middleware.

CarterPerez-dev: Yea true, if you want a specific permission type of middleware but just for this instance, I think it may be better to sorta globalize it then, this could be overengineered, but it then would handle the ticket ownership check as well, which is the same case. It could also be used for any other resource - ownership checks that may happen in the future, or any current ones I dont know of. Meaning, maybe some thing like this?

```python
def check_ownership(resource_key: str, user_field: str, error_message: str = None):
    def decorator(f):
        @wraps(f)
        def decorated_function(*args, **kwargs):
            resource = kwargs.get(resource_key)
            current_user = kwargs.get("current_user")
            
            if not resource:
                raise ValueError(f"Resource '{resource_key}' must be loaded first")
            if not current_user:
                raise ValueError("current_user must be loaded first")
                
            resource_user_id = getattr(resource, user_field)
            if resource_user_id != current_user.id:
                message = error_message or f"You can only access your own {resource_key}"
                raise PermissionError(message)
                
            return f(*args, **kwargs)
        return decorated_function
    return decorator
 ```   
and then in the routes:
    
Notifications
```python
@user_endpoint()
@load_notification(source=LoaderType.PARAM)
@check_ownership(resource_key="notification", user_field="recipient_id")
def mark_read(self, notification_id: int, notification, current_user, **kwargs):
```
Tickets (replaces `load_ticket_with_user` loader)
```python
@user_endpoint()
@load_ticket(source=LoaderType.PARAM)  
@check_ownership(resource_key="ticket", user_field="author_id")
def close_my_ticket(self, ticket_id: int, ticket, current_user, **kwargs):
```

fpayer (approved):

PR #167, opened 2025-09-17
wbsocket redis testing

CarterPerez-dev:
```json
                                                                                                                                                                            

         

                                                                                                
minor semver because I added a workflow



                                                                  

```


backend/ng/notifications/tests/integration/test_notification_e2e.py
```diff
@@ -0,0 +1,148 @@
...
+
+class TestExternalIntegration:
+    """
+    Tests that run outside Docker
+    """
+    def test_ctfd_is_running(self):
+        """
+        Verify CTFd is accessible
+        """
+        resp = requests.get("http://localhost:8001/")
```
fpayer: Recommend having the base path (`http://localhost:8001/`) be a const so that we can easily update the test cases if the port needs to be changed for any reason in the future.

backend/ng/notifications/tests/test_notification_system_works.py
```diff
@@ -0,0 +1,101 @@
+"""
+Test Basic (with info)
+"""
+import pytest
+from flask_socketio import SocketIOTestClient
+
+from ..sockets import (
+        user_connections,
```
fpayer: Why 8-tab spacing?

.github/workflows/notification-e2e.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pytest requests python-socketio[client] -r backend/ng/requirements.txt -r external/CTFd/requirements.txt
+
+      - name: Start Docker services
+        working-directory: backend/ng/notifications/tests/integration
+        run: |
+          docker compose -f docker-compose.test.yml up -d
```
zoecough, line 39: I think this may be overcomplication; is this separate docker compose spin-up just for redis? this can be done with the normal test runner, take a look at the github marketplace for options to stand up a redis server in github actions

backend/ng/notifications/tests/integration/docker-compose.test.yml
zoecough, line 1: Potentially unnecessary if using the redis server marketplace item for github actions if the only purpose this serves is to spin up redis on the github action

zoecough (commented):
especially in cases like this where your PR has a couple different things happening, commit messages that are specific about what the changes in the commit are are helpful to the rest of the team both for reviewing purposes and for future development

backend/ng/notifications/tests/test_notification_system_works.py
CarterPerez-dev: I use and copy over my personal .style.yapf formatter and forgot to change that setting, I'll  revert it back to 4.

.github/workflows/notification-e2e.yml
CarterPerez-dev: True, though this is just the first phase - testing core WebSocket auth directly to ctfd before adding our nginx/infrastructure layers which would be much easier and more realistic through a compose, so right now its really basic and does look overcompcliated, but will probably be needed down the line.

fpayer (approved):

PR #170, opened 2025-09-18
Admin event challenges route

trevnels-cs:
The admin page challenges list previously used the user-facing event challenges endpoint, which now has stricter access control that prevents admins from managing challenges unless they're registered for and have started an event.

This creates a separate admin endpoint to get event challenge lists which is not subject to user-facing permission controls so that admins can view challenges associated with any event through the admin panel.

### DIFF
--- a/backend/ng/event/routes/admin_routes.py
+++ b/backend/ng/event/routes/admin_routes.py
@@ -217,6 +217,24 @@ class EventRegister(Resource):
 @events_admin_namespace.route("/<int:event_id>/challenges")
 class EventChallenges(Resource):
     @events_admin_namespace.doc(
+        description="Get all challenges for a specific event",
+        responses={
+            200: "Success - Returns list of challenges for the event",
+            404: "Not found - Event does not exist",
+            403: "Forbidden - Admin access required",
+            500: "Internal Server Error",
+        },
+    )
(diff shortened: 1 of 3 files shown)


backend/ng/event/routes/admin_routes.py
```diff
@@ -216,6 +216,24 @@ def post(self, event_id, user_id, **kwargs):
...
+            500: "Internal Server Error",
+        },
+    )
+    @admin_endpoint()
+    @load_event(source=LoaderType.PARAM)
+    def get(self, event: Event, **kwargs):
+        """
+        Get all challenges for an event
+        """
+        challenges = event.get_all_challenges()
```
RayofLightz, line 234: Looking at this model this might have serialization issues? This is returning the raw db object. You might want to map over `challenge.serialize` As this will ensure a typed dict

RayofLightz (approved):
This should be fine, just consider the serialization stuff.

backend/ng/event/routes/admin_routes.py
trevnels-cs: This is the same implementation as `/event/challenges`, which has been working - serialize is called implicitly inside of `success_response` for objects that have that method defined

PR #171, opened 2025-09-18
Initial working event start button

trevnels-cs:
Start button for event pages. Currently missing a lot of permission checking logic, since the necessary permission data is not available to the frontend currently, but having the button in place is important for being able to test event/challenge functionality going forwards.

### DIFF
--- a/frontend/src/hooks/events.ts
+++ b/frontend/src/hooks/events.ts
@@ -114,6 +114,20 @@ export function useMyTeamMembers(eventId: number | undefined) {
 }
 
 /**
+ * Starts the user's team for a specific event. User must be captain.
+ * @param eventId The id of the event
+ */
+export function startMyTeam(eventId: number) {
+  return apiMutation(`/events/${eventId}/me/team/start`, undefined, {
+    method : 'POST',
+  }).then(() => {
+    mutate(`/events/${eventId}/me/team`);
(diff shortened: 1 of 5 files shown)


frontend/src/hooks/events.ts
```diff
@@ -113,6 +113,20 @@ export function useMyTeamMembers(eventId: number | undefined) {
...
+ * @param eventId The id of the event
+ */
+export function startMyTeam(eventId: number) {
+  return apiMutation(`/events/${eventId}/me/team/start`, undefined, {
+    method : 'POST',
+  }).then(() => {
+    mutate(`/events/${eventId}/me/team`);
+    mutate(`/events/${eventId}/challenges`);
+    mutate(`/events/${eventId}/me/challenges`);
+  });
```
cjones-cs, line 127: wouldn't only the /me endpoints need updating? /challenges may not have a field you need to update

cjones-cs (approved):

frontend/src/hooks/events.ts
trevnels-cs: Currently the challenges endpoint goes from a 403 to loading the data when starting an event, so the mutate is needed to trigger that update. Once the frontend is aware of permissions, loading the challenges will be guarded behind that and won't need to be refetched like this

PR #181, opened 2025-09-23
Add the ability to pull all containers for a challenge

RayofLightz:
Ads an endpoint for pulling containers from ECR. We have to report status of pulls over websockets because otherwise we will run into http time out issues when pulling 1gb+ images.


backend/ng/event/routes/admin_routes.py
```diff
@@ -283,3 +288,37 @@ def post(self, event_id, event, json_data, **kwargs):
...
+        responses={
+            200: "Success - Returns true",
+            404: "Not found - Event does not exist",
+            403: "Forbidden - Admin access required",
+            500: "Internal Server Error",
+        },
+    )
+    @admin_endpoint()
+    @load_challenge(source=LoaderType.PARAM)
+    def get(self, challenge: Challenge, current_user, **kwargs):
```
trevnels-cs, line 304: Since this is a server-side action that the client calls, I think this should be a post instead of get

backend/ng/event/routes/admin_routes.py
```diff
@@ -283,3 +288,37 @@ def post(self, event_id, event, json_data, **kwargs):
         challenge = import_challenge_from_yaml(event, json_data)
         return success_response(challenge)
 
+@events_admin_namespace.route("/<int:event_id>/challenge/<int:challenge_id>/pull")
```
trevnels-cs, line 291: I think it makes more sense for this to go under `containers/routes/admin_routes.py` as `/challenge/<int:challenge_id>/pull` since it's a container operation that will be used alongside the endpoints defined there.

RayofLightz: Those routes are for containers with active instances. This is honestly a challenges scoped operation, but for some reason we have been keeping the challenge operations under the event routes

backend/ng/challenge/models/ContainerBlueprint.py
```diff
@@ -136,4 +142,9 @@ def render_environment(self, team_seed: str) -> dict[str, str] | list[str]:
         if isinstance(self.environment, list):
             return self.environment
 
-        return {k: (v(team_seed) if isinstance(v, Callable) else v) for k, v in self.environment.items()}
\ No newline at end of file
+        return {k: (v(team_seed) if isinstance(v, Callable) else v) for k, v in self.environment.items()}
+
+    def pull_image(self):
```
fpayer, line 170: How do the ECR auth creds get passed to the docker client?

RayofLightz: Its in `backend/ng/containers/utils/Client.py` It is currently passed as a config

backend/ng/event/routes/admin_routes.py
trevnels-cs: With the challenge admin panel, I've been putting a lot of challenge admin routes under `/challenge` instead of under `/event`, and I think it would be best to continue doing that moving forward, especially because having to populate the event id for routes that don't need it can make things more difficult on the frontend side

backend/ng/challenge/models/ContainerBlueprint.py
fpayer: This should probably get set via a .env file

RayofLightz: I don't disagree but I couldn't see anywhere in the code that we are actually using the env files. I can port it over to using a env file

backend/ng/event/routes/admin_routes.py
RayofLightz: I agree I will move it under `/challenge`

trevnels-cs (approved):

backend/ng/challenge/models/ContainerBlueprint.py
fpayer: https://github.com/CyberSkyline/ctf-ng/blob/development/backend/ng/core/routes/views.py

All of the env and config values get merged together and can be accessed via `get_app_config`

```
from CTFd.utils import get_app_config
static_build_path = get_app_config("STATIC_BUILD_PATH")
```

PR #183, opened 2025-09-23
Support static ips in networks

johnmarsden-edu:


lib/parser/src/cyber_skyline/chall_parser/compose/service.py
```diff
@@ -29,9 +29,17 @@
...
+from cyber_skyline.chall_parser.compose.validators import is_ipv4
 
 type CapAdd = Literal['NET_ADMIN', 'SYS_PTRACE']
 
+@define
+class ServiceNetwork:
+    """Represents a network attachment for a service.
+    """
+    ipv4_address: str | None = field(default=None, validator=v.optional(is_ipv4))  # Static IPv4 address for the service on this network
+
```
RayofLightz, line 41: Does this need to get used further down in the service def?

johnmarsden-edu: You would be completely correct and in fact I did catch that after I wrote a test

.github/workflows/chall-check-testrunner.yml
```diff
@@ -20,7 +20,12 @@ jobs:
         with:
           python-version: '3.12'
 
-      - name: Install dependencies
+      - name: Install parser dependencies
+        run: |
+          cd lib/parser
+          pip install -e .
+
+      - name: Install chall-check dependencies
```
johnmarsden-edu, line 28: @zoecough this is what I did

RayofLightz (approved):
Looks good

PR #184, opened 2025-09-23
Frontend websocket integration

trevnels-cs:
The backend can now trigger SWR refetches on the frontend via websocket. SWR hooks that use affected endpoints will automatically reload their data from the server and re-render the components they are used by.

Had to update `serve_debug.py` to use `socketio.run` as described in the flask-socket.io documentation, as `app.run` is not compatible with websockets (and results in a 500 error when connecting.)

I also configured Vite to do a full refresh of the page when `socket.ts` is modified, as the default hot-reload behavior caused issues with multiple sockets/listeners being created unnecessarily.

### DIFF
--- a/conf/ctfd/serve_debug.py
+++ b/conf/ctfd/serve_debug.py
@@ -42,4 +42,4 @@ if args.profile:
     toolbar.init_app(app)
     print(f" * Flask profiling running at http://0.0.0.0:${args.port}/flask-profiler/")
 
-app.run(debug=True, threaded=True, host="0.0.0.0", port=args.port)
+app.extensions["socketio"].run(app, host="0.0.0.0", port=args.port, debug=True)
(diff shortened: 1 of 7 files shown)


conf/ctfd/serve_debug.py
```diff
@@ -42,4 +42,4 @@
     toolbar.init_app(app)
     print(f" * Flask profiling running at http://0.0.0.0:${args.port}/flask-profiler/")
 
-app.run(debug=True, threaded=True, host="0.0.0.0", port=args.port)
+app.extensions["socketio"].run(app, host="0.0.0.0", port=args.port, debug=True)
```
RayofLightz, line 45: Did you mean to remove the threaded part?

trevnels-cs: The server that socket.io starts is backed by gevent, which uses its own concurrency system by default. `threaded` is an option specific to werkzeug, which is no longer being used here

fpayer (approved):

trevnels-cs:
Also note that changes to the debug entrypoint will require a stop/start cycle (container rebuild) after pulling

PR #185, opened 2025-09-23
Support for public pages and other things

### LINKED ISSUE #177: Support for public pages
To be used for things like the public leaderboard.

- [x] Create a `@public` endpoint decorator
- [ ] Change behavior of authed endpoints to be a 401 instead of a redirect to CTFd login
- [ ] Create endpoints to list permissions per event
- [x] Have a way for the frontend to tell if the user is authenticated or not (this could be the same as the permissions endpoint)

droge91:
Changes the behavior of several endpoints to be publicly accessible without authentication
Changes unauthed requests to 401 instead of redirect
adds an endpoint to get user permissions for an event

Fixes #177

### DIFF
--- a/backend/ng/core/middleware/auth.py
+++ b/backend/ng/core/middleware/auth.py
@@ -64,7 +64,7 @@ def _auth_handler(f, auth_required, json_required, validation_func):
         if auth_required:
             current_user = get_current_user()
             if not current_user:
-                abort(401)
+                raise PermissionError("Authentication is required to access this endpoint.")
             kwargs["current_user"] = User.find_or_create_by_ctfd_id(current_user.id)
         if json_required:
             if not request.is_json:
--- a/backend/ng/core/middleware/auth.py
+++ b/backend/ng/core/middleware/auth.py
@@ -106,7 +106,7 @@ def admins_only(f):
 
 
 # Convenience decorators for common patterns
-def user_endpoint(auth_required=True, json_required=False, validation_func=None):
+def user_endpoint(json_required=False, validation_func=None):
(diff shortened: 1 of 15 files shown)


backend/ng/core/middleware/auth.py
fpayer, line 1: Is it the changes to this file that is causing the 401 instead of the redirect or is that from something else?

backend/ng/permissions/routes/user_routes.py
```diff
@@ -0,0 +1,40 @@
...
+from ...core.middleware.permission_middleware import (
+    check_permissions,
+)
+
+
+permissions_user_namespace = Namespace("/permissions", description="Permissions management endpoints for users")
+logger = get_logger(__name__)
+
+
+@permissions_user_namespace.route("/<int:event_id>/me")
```
trevnels-cs, line 18: We should also have an endpoint for checking global permissions outside of the context of an event, i.e. can this user access the admin panel?

backend/ng/core/middleware/auth.py
droge91: Yeah, previously the api endpoint decorator was using CTFd's authed_only decorator which is what had the redirect behaviour

backend/ng/permissions/routes/user_routes.py
droge91: There are no specific permissions for viewing admin specific stuff, we could make one like `CAN_VIEW_ADMIN` that would be returned from such an endpoint or just make a user route for getting your own roles so you can use whether or not someone is an admin for managing their views. what do you think?

fpayer (approved):

PR #187, opened 2025-09-24
Pass None values through validator to allow clearing fields

### LINKED ISSUE #179: PUT requests can't clear optional fields
Once an optional field has been set on a model (i.e. event), future PUT requests are unable to reset the field to None by sending a null value, since the empty field is erased from the validated data. Instead, we need to pass None values through so they will be written to model objects.

trevnels-cs:
Resolves #179. `None` values passed for optional fields will be passed through the validator, allowing optional model fields to be cleared.

For this to happen, JSON input must include the fields with a value of `null` - fields omitted entirely won't be passed through as `None` due to how the update/validation logic is used in existing code. This kind of blurs the lines between `PUT` and `PATCH` semantics which feels a little weird, but with the way existing update/validation code works there's not much that can be done about that without significant refactoring.

### DIFF
--- a/backend/ng/challenge/tests/test_challenge_model.py
+++ b/backend/ng/challenge/tests/test_challenge_model.py
@@ -167,15 +167,15 @@ class TestChallengeValidation:
 
     def test_validate_optional_fields_missing_should_pass(self, event):
         """Test that validation passes when optional fields are missing."""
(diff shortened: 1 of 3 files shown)


droge91 (approved):
This all looks fine to me

PR #189, opened 2025-09-24
Event leaderboard UI

trevnels-cs:
Show an ag-grid populated from API data on each event's Leaderboard tab, along with the user's team's current score.

Also fixes a cherry-pick mismatch with previous start modal changes that caused the event page to crash immediately.

image


frontend/src/routes/events/StartModal.tsx
```diff
@@ -1,13 +1,12 @@
...
 import { startMyTeam, useMyTeam } from '@/hooks/events';
-import type { Event } from '@/types';
 import { Button } from '@radix-ui/themes';
 import Modal from 'components/Modal';
 import { TbPlayerPlay } from 'react-icons/tb';
 
-export default function StartModal({ event }: {event: Event}) {
-  const { data : myTeam } = useMyTeam(event.id);
-  const handleStart = async () => startMyTeam(event.id);
+export default function StartModal({ eventId }: {eventId: number}) {
```
cjones-cs, line 7: Have you double checked that all the cases where this is called is eventid instead of event? I recall an emergency fix we did for this last week.

trevnels-cs: This was intended to be the correct fix for that issue, but was missed in the original cherry-pick - StartButton should only be taking the id instead of model object. I have now reverted the previous fix in this PR, so the two files should be back in sync now.

cjones-cs (approved):

PR #192, opened 2025-09-25
Container management and challenge import fixes

trevnels-cs:
- Add missing ChallengeVariable type string during question validation, resolves failing YAML imports
- Update admin-facing restart/recycle endpoints to POST for consistency with user-facing ones
- Update frontend container management hooks to use POST
- Fix missed refactor in `get_instance_group` that was causing exceptions
- Fix `start_containers` type signature
- Catch exception thrown when user workspace does not exist in `get_current_challenge`

### DIFF
--- a/backend/ng/challenge/models/Question.py
+++ b/backend/ng/challenge/models/Question.py
@@ -90,6 +90,7 @@ class Question(db.Model):
             validator.validate_model_id(
                 data,
                 "answer_variable_id",
+                "ChallengeVariable",
                 required=templated,
                 friendly_name="Answer Variable ID",
             )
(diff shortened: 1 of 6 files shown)


backend/ng/challenge/models/Question.py
```diff
@@ -90,6 +90,7 @@ def validate(cls, data: dict[str, Any]) -> dict[str, Any]:
             validator.validate_model_id(
                 data,
                 "answer_variable_id",
+                "ChallengeVariable",
```
johnmarsden-edu, line 93: This should be fixed now in development

trevnels-cs: The PR diff in the GitHub UI was generated before those changes were merged, so it's stale - this PR should still merge cleanly as-is and not double up that change

RayofLightz (approved):

PR #208, opened 2025-09-27
Email Notifications with AWS SES for Support Tickets

CarterPerez-dev:
# Email Notifications with AWS SES for Support Tickets
***closes issue https://github.com/CyberSkyline/ctf-ng/issues/174***
---
* AWSEmailService: boto3 integration with graceful fallback when not configured
* TicketEmailTemplates: basic HTML/text templates for all ticket events,
  (not sure how we wanted to do the html templates so I made them basic for now)
* Email routing in NotificationService based on ticket context:
---
  - New tickets → admin support inbox - `ADMIN_SUPPORT_INBOX_EMAILS`
  - User replies to unassigned tickets → admin support inbox
---
  - Admin replies to a team associated ticket → all team members get email
  - Admin replies to an individual ticket ( no association ) → ticket author only gets email
---
  - User replies to an admin assigned ticket → assigned admin gets email
---
* Implicit assignment: admin replies to unassigned ticket auto-assigns to the replying admin
* AWS send from email configurable via .env as `AWS_SES_FROM_EMAIL`
* Configurable admin support inbox/s for unassigned ticket routing/new tickets - via .env as `ADMIN_SUPPORT_INBOX_EMAILS`
----
* Pytests testing business logic, smart routing, and error handling with mocked AWS calls
* Test script for real AWS email verification and sending - `/backend/ng/scripts/_test/_test_email_send.py`
* .env setup example in `/conf/.env.default `
* Dependencies added: `types-boto3==1.40.40`, `types-redis==4.6.0.20241004`

PR #211, opened 2025-09-29
fix notification reply/assigment logic

CarterPerez-dev:
- If user creates a ticket (which is always going to be unnasigned initially) - then notify all *support* admins there is a new ticket
- Admin replies to that ticket while being unnasigned - user still gets a notification
***(Or if admin is assigned and replies - user still also gets a notification)***
- User then replies to that same ticket with an unnasigned admin - all *support* admins get notification (because its still unassigned) - (OR if admin IS assigned then only THAT admin gets a notification of user reply)
***NOTE: (PR: https://github.com/CyberSkyline/ctf-ng/pull/208 adds the functionality to where when an admin replies to a ticket while being unnasigned - that admin automatically gets assigned to that ticket (for now thats not the case due to it being tied to the email service which is not on this branch - therefore - for now, all admins get notified until an admin is assigned)***

### DIFF
--- a/backend/ng/notifications/services/notification_service.py
+++ b/backend/ng/notifications/services/notification_service.py
@@ -134,25 +134,8 @@ class NotificationService:
         subject: str,
     ) -> None:
         """
-        Notify all support staff about a new support ticket with DB notifications
+        Notify all support staff about a new support ticket via refetch only
         """
-        support_users = get_support_role_users()
-
-        for support_user in support_users:
-            notification = Notification.create_notification(
-                notification_type = NotificationType.TICKET_CREATE,
-                title = "New Support Ticket",
-                message = f"New ticket created: {subject}",
-                recipient_id = support_user.ctfd_user.id,
-                sender_id = author_id,
-                ticket_id = ticket_id,
-                commit = False,
-            )
-
-            NotificationService._emit_notification(notification)
-
-        db.session.commit()
-
         NotificationService._emit_refetch(
             path = "/ng/support/tickets",
         )
--- a/backend/ng/notifications/services/notification_service.py
+++ b/backend/ng/notifications/services/notification_service.py
@@ -155,7 +155,6 @@ class NotificationService:
 
         NotificationService._emit_refetch(
             path = "/ng/support/tickets",
-            user_ids = None
         )
 
     @staticmethod
@@ -185,7 +184,6 @@ class NotificationService:
 
         NotificationService._emit_refetch(
             path = f"/ng/support/tickets/{ticket_id}",
-            user_ids = None
         )
 
     @staticmethod
--- a/backend/ng/notifications/services/notification_service.py
+++ b/backend/ng/notifications/services/notification_service.py
@@ -22,6 +22,7 @@ from ..models import (
     AnnouncementType,
 )
 from ...team.models import TeamMember
+from ...permissions.controllers import get_support_role_users
 
 
 class WebSocketEvent(str, Enum):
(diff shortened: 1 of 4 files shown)


backend/ng/notifications/services/notification_service.py
```diff
@@ -133,11 +134,58 @@ def notify_new_ticket(
...
+                recipient_id = support_user.ctfd_user.id,
+                sender_id = author_id,
+                ticket_id = ticket_id,
+                commit = False,
+            )
+
+            NotificationService._emit_notification(notification)
+
+        db.session.commit()
+
```
cjones-cs: Should we really be notifying support people about new tickets? Don't they usually just check the table manually so we don't get slammed with notifs?

backend/ng/notifications/services/notification_service.py
```diff
@@ -133,11 +134,58 @@ def notify_new_ticket(
...
+        NotificationService._emit_refetch(
+            path = f"/ng/support/tickets/{ticket_id}",
+            user_ids = None
```
cjones-cs: Does this exclude muted tickets?

backend/ng/support/controllers/all_actions/create_ticket_message.py
```diff
@@ -30,19 +29,31 @@ def create_ticket_message(
...
+                author_id=author_id,
+                recipient_id=ticket.assigned_to,
+                is_admin_reply=False,
+            )
+        else:
+            # Ticket is unassigned - notify all support staff
+            NotificationService.notify_unassigned_ticket_reply(
+                ticket_id=ticket.id,
+                author_id=author_id,
+            )
```
cjones-cs: Same question about muted

backend/ng/notifications/services/notification_service.py
```diff
@@ -133,11 +134,58 @@ def notify_new_ticket(
...
+            )
+
+            NotificationService._emit_notification(notification)
+
+        db.session.commit()
+
         NotificationService._emit_refetch(
             path = "/ng/support/tickets",
-            user_ids = None  # Via _emit_refetch fallback
+            user_ids = None
```
RayofLightz: User_ids defaults to None in the emit refetch so this is unneeded

backend/ng/notifications/services/notification_service.py
```diff
@@ -133,11 +134,58 @@ def notify_new_ticket(
...
+            )
+
+            NotificationService._emit_notification(notification)
+
+        db.session.commit()
+
+        NotificationService._emit_refetch(
+            path = f"/ng/support/tickets/{ticket_id}",
+            user_ids = None
         )
```
RayofLightz, line 170: What I mentioned earlier

RayofLightz (commented):
@cjones-cs asked me to take a look at some of this since its backend. Just a few comments

backend/ng/support/tests/test_support_api.py
```diff
@@ -595,6 +603,322 @@ def test_ticket_list_includes_names(self, logged_in_client, user, event, challen
...
+        self,
+        logged_in_client,
+        user,
+        admin,
+        assigned_ticket,
+        db_session,
+    ):
+        """
+        Test that when a user replies to an assigned ticket,
+        the assigned admin receives a notification
```
cjones-cs, line 714: Should be the assigned support person, since we made the distinction now

backend/ng/notifications/services/notification_service.py
CarterPerez-dev: ahh.. you're saying that new tickets shouldn't be notifications then?

cjones-cs: I'm torn on it, so I'll ask the group

cjones-cs: Yeah, we don't need to alert support staff on creation. Just alert users they created a ticket

backend/ng/support/tests/test_support_api.py
```diff
@@ -595,6 +620,344 @@ def test_ticket_list_includes_names(self, logged_in_client, user, event, challen
...
+            ticket_id=assigned_ticket.id,
+            sender_id=user.id
+        ).all()
+
+        assert len(admin_notifications) >= 1
+        notification = admin_notifications[-1]
+        assert notification.type.value == "ticket_message"
+        assert notification.sender_id == user.id
+
+    def test_user_replies_to_unassigned_ticket_all_support_staff_get_notification(
```
cjones-cs: Is this extraneous?

cjones-cs (changes requested):
Double check that one test case matches expected behavior

CarterPerez-dev:
> Double check that one test case matches expected behavior

test quote reply

cjones-cs (approved):

PR #216, opened 2025-10-02
Use react-hook-form in modals

trevnels-cs:
- Use react-hook-form to perform validation in Modal forms, and to allow cleaner use of checkboxes, selects, etc. without manually maintaining their state separate from the form values
  - Validation rules have been kept as-is, but we could adopt a library like yup or zod in the future to enforce more advanced validation schemas on the client-side
  - Form submit handler now consumes typed rhf form data as its parameter, instead of a `FormData` object that has to be unpacked.
- Updates the event registration modal form to include Terms/Conditions (as markdown, so they can be easily configurable in the future), as well as an option to manually enter/paste an invite code (good practice to provide an alternative to "magic links")
- Updates the tag create/edit modal to use a color picker input instead of manual hex code entry

### DIFF
--- a/frontend/src/routes/admin/events/EventDataForm.tsx
+++ b/frontend/src/routes/admin/events/EventDataForm.tsx
@@ -22,11 +22,9 @@ export default function EventDataForm({
         {(injected) => (
           <TextField.Root
             placeholder="Event Name"
-            // eslint-disable-next-line react/jsx-props-no-spreading
             {...register('name', {
               required : 'Event name is required',
             })}
-            // eslint-disable-next-line react/jsx-props-no-spreading
             {...injected}
           />
         )}
@@ -38,9 +36,7 @@ export default function EventDataForm({
             placeholder="Event Description"
             rows={4}
             resize="vertical"
-            // eslint-disable-next-line react/jsx-props-no-spreading
             {...register('description')}
-            // eslint-disable-next-line react/jsx-props-no-spreading
             {...injected}
           />
         )}
@@ -65,7 +61,6 @@ export default function EventDataForm({
                     name={field.name}
                     ref={field.ref}
                     size="3"
-                    // eslint-disable-next-line react/jsx-props-no-spreading
                     {...injected}
                   />
                 </Box>
@@ -91,7 +86,6 @@ export default function EventDataForm({
                     name={field.name}
                     ref={field.ref}
                     size="3"
-                    // eslint-disable-next-line react/jsx-props-no-spreading
                     {...injected}
                   />
                 </Box>
@@ -117,7 +111,6 @@ export default function EventDataForm({
                     name={field.name}
                     ref={field.ref}
                     size="3"
-                    // eslint-disable-next-line react/jsx-props-no-spreading
                     {...injected}
                   />
                 </Box>
@@ -133,7 +126,6 @@ export default function EventDataForm({
           {(injected) => (
             <TextField.Root
               type="number"
-              // eslint-disable-next-line react/jsx-props-no-spreading
               {...register('max_team_size', {
                 required : 'Max team size is required',
                 valueAsNumber : true,
@@ -146,7 +138,6 @@ export default function EventDataForm({
                   message : 'Max team size cannot exceed 8',
                 },
               })}
-              // eslint-disable-next-line react/jsx-props-no-spreading
               {...injected}
             />
           )}
@@ -158,7 +149,6 @@ export default function EventDataForm({
               type="number"
               placeholder="No time limit"
               step={1}
-              // eslint-disable-next-line react/jsx-props-no-spreading
               {...register('time_limit_minutes', {
                 valueAsNumber : true,
                 min : {
@@ -166,7 +156,6 @@ export default function EventDataForm({
                   message : 'Time limit must be at least 1 minute',
                 },
               })}
-              // eslint-disable-next-line react/jsx-props-no-spreading
               {...injected}
             />
           )}
@@ -178,11 +167,9 @@ export default function EventDataForm({
           {(injected) => (
             <TextField.Root
               type="datetime-local"
-              // eslint-disable-next-line react/jsx-props-no-spreading
               {...register('registration_start_date', {
                 valueAsDate : true,
               })}
-              // eslint-disable-next-line react/jsx-props-no-spreading
               {...injected}
             />
 
@@ -192,11 +179,9 @@ export default function EventDataForm({
           {(injected) => (
             <TextField.Root
               type="datetime-local"
-              // eslint-disable-next-line react/jsx-props-no-spreading
               {...register('registration_end_date', {
                 valueAsDate : true,
               })}
-              // eslint-disable-next-line react/jsx-props-no-spreading
               {...injected}
             />
           )}
@@ -208,11 +193,9 @@ export default function EventDataForm({
           {(injected) => (
             <TextField.Root
               type="datetime-local"
-              // eslint-disable-next-line react/jsx-props-no-spreading
               {...register('start_time', {
                 valueAsDate : true,
               })}
-              // eslint-disable-next-line react/jsx-props-no-spreading
               {...injected}
             />
           )}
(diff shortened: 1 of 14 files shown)


cjones-cs:
The option to input an invite code was intentionally excluded as it was cited as "confusing to users". If we are going to change that, then you should include logic to rip the invite code out of the url. You can rewrite it, or snag it from the tm on cs.

Also worth noting, if you are changing this modal functionality, they also wanted the team's name displayed when joining a team. That way it's more clear which one they are joining.

frontend/src/routes/admin/events/EventDataForm.tsx
```diff
@@ -1,151 +1,233 @@
...
+      <FormField label="Description" error={errors.description}>
+        {(injected) => (
           <TextArea
-            defaultValue={initial?.description || ''}
             placeholder="Event Description"
             rows={4}
             resize="vertical"
+            // eslint-disable-next-line react/jsx-props-no-spreading
+            {...register('description')}
+            // eslint-disable-next-line react/jsx-props-no-spreading
```
cjones-cs: Would it be better to disable this lint rule vs escaping out of it? Looks like we are going to have to do it a lot.

frontend/src/routes/events/RegistrationModal.tsx
```diff
@@ -1,111 +1,186 @@
...
+              <Text as="label" size="2">
+                <Flex gap="2">
+                  <Checkbox
+                    id="termsConditions"
+                    checked={field.value}
+                    onCheckedChange={(checked) => field.onChange(checked)}
+                    onBlur={field.onBlur}
+                    ref={field.ref}
+                  />
+                  I agree to the Terms and Conditions
```
cjones-cs, line 169: Instead of agreeing to the terms and conditions, we could just put the checkbox next to the markdown text.

frontend/src/routes/events/RegistrationModal.tsx
```diff
@@ -1,111 +1,186 @@
...
+                  />
+                )}
+              </FormField>
+            </>
+          )}
+
+          <Box className="text-sm">
+            <RadixMarkdown>
+              {`I acknowledge that I have read and understand the [eligibility criteria](/) and [contest rules](/)
+          for CISA's President's Cup Cybersecurity Competition. I agree to (1) comply with these criteria and rules and
```
cjones-cs, line 144: Have you asked OTA/CISA for these links? I think they probably exist on their website somewhere already.

cjones-cs: Either that, or make sure to put in a new github issue to follow up on this so it isn't forgotten

frontend/src/routes/events/RegistrationModal.tsx
cjones-cs: Upon testing this, I discovered that the Register button goes green when I error out the checkbox and have something in the name field. Minor bug, but we want to ensure that they have checked the box before allowing a green registration button.

frontend/src/routes/admin/events/EventDataForm.tsx
```diff
@@ -1,151 +1,233 @@
...
+              })}
+              // eslint-disable-next-line react/jsx-props-no-spreading
+              {...injected}
             />
-          </Form.Control>
-          <Form.Label>Registration Open</Form.Label>
-        </Form.Field>
+          )}
+        </FormField>
       </Flex>
```
cjones-cs, line 214: The logic for the form is also not correct. Testing revealed putting in a registration opens date allowed the create button to go green. 

Image

frontend/src/routes/admin/tags/SupportTags/TagModal.tsx
```diff
@@ -1,28 +1,21 @@
...
 import { createSupportTag, putSupportTag } from '@/hooks/support';
 import type { TicketTag } from '@/types';
-import { isUndefined } from 'lodash';
-import { COLOR_POSITIVE, COLOR_WARNING } from '@/constants';
+import { Button, Flex, TextField } from '@radix-ui/themes';
+import FormField from 'components/FormField';
+import Modal from 'components/Modal';
+import { isUndefined, omit } from 'lodash';
 
 export default function CreateTagModal(
```
cjones-cs, line 9: I think we may be checking for touched instead of values here. Seems I can touch all the fields to get the green, but it's not actually checking the values are non-null
Image

frontend/src/routes/admin/tags/SupportTags/TagModal.tsx
```diff
@@ -39,45 +32,55 @@ export default function CreateTagModal(
...
+                <input
+                  type="color"
+                  className="rounded"
+                  // eslint-disable-next-line react/jsx-props-no-spreading
+                  {...register('color', {
+                    required : 'Please enter a color.',
+                  })}
+                  // eslint-disable-next-line react/jsx-props-no-spreading
+                  {...injected}
+                />
```
cjones-cs, line 60: Is there a way we can get the picker to default to the hex input instead of defaulting to showing rgb? 

Image

frontend/src/routes/admin/teams/ScoreAdjustModal.tsx
```diff
@@ -2,16 +2,18 @@ import { COLOR_WARNING } from '@/constants';
...
-import { Form } from 'radix-ui';
 import { TbPlusMinus } from 'react-icons/tb';
 
 export default function ScoreAdjustModal({ team }: { team: Team }) {
-  const handleSubmit = async (data: FormData) => {
-    const { points, reason } = Object.fromEntries(data.entries());
-
-    return adjustPoints(team.event_id, team.id, Number(points), reason as string);
-  };
+  const handleSubmit = async ({
```
cjones-cs, line 10: Same issues as the others with validation

cjones-cs (changes requested):

frontend/src/routes/admin/tags/SupportTags/TagModal.tsx
trevnels-cs: This is up to the browser/OS implementation, so as far as I know this isn't configurable from our end

trevnels-cs:
Regarding the invite codes, I still think the URL should be the primary means of handling them, and that functionality still works as expected. However, I think that clicking the register modal and only having the option to create a team, with no indication that I can join one, seems off to me - there should at least be some instruction for players looking to join a team that they will need a link/code from their captain.

I do also think it makes sense to allow manual invite code entry, since there are special cases where that may be useful (i.e. I was sent the code to my phone and want to join on my computer - would be easier for me to type out the code than figuring out how to move that link to the other device.) We could hide it behind an additional button like "Enter code manually" if it's too confusing, but I think there is value in preserving that option for the cases where it's useful.

> Also worth noting, if you are changing this modal functionality, they also wanted the team's name displayed when joining a team. That way it's more clear which one they are joining.

I like this in principle, but it means there would need to be an endpoint that maps invite codes to team info - we'd need to make sure this is limited such that someone can't try to brute-force enumerate invite code -> team mappings.

frontend/src/routes/events/RegistrationModal.tsx
trevnels-cs: These two are different tabs within the same page on CISA's PC7 site, and there doesn't seem to be a way to parameterize which tab to select in the URL. We will probably need to wait until after the shutdown to clarify how these should work.

cjones-cs: If it's on the same page, then we can just make this a single link and go to the page where both of those things are found. I think it's too much for us to request they make a tabbing functionality change for our links to have tab specific redirection.

trevnels-cs: Tracking in #220. I don't think it makes sense to request changes to the tabbing, but we should make sure that linking to the page instead of the specific tab is acceptable, especially since rules/eligibility are not visible by default - the user has to switch to those tabs manually to see the associated info.

cjones-cs (approved):

PR #218, opened 2025-10-03
Add yaml linting

johnmarsden-edu:


backend/ng/event/controllers/admin/lint_challenge.py
```diff
@@ -0,0 +1,153 @@
...
+    elif isinstance(error, ValueError):
+        return [SerializedLintMessage(
+            message=f"Value error: {str(error)}"
+        )]
+    else:
+        return [SerializedLintMessage(
+            message=f"Error: {str(error)}"
+        )]
+
+def lint_challenge(json_data) -> SerializedLint | None:
```
trevnels-cs: Not essential, but I think that consuming the already-decoded yaml string here and letting the route handle decoding the data makes more sense to me organizationally

johnmarsden-edu:
```suggestion
def lint_challenge(yaml_content: str) -> SerializedLint | None:
```

backend/ng/event/controllers/admin/lint_challenge.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    Import a challenge from a YAML definition.
+
+    :return: Any warnings or errors that were returned
+    """
+    payload = base64.urlsafe_b64decode(json_data["yaml"])
```
johnmarsden-edu:
```suggestion
```

backend/ng/event/controllers/admin/lint_challenge.py
```diff
@@ -0,0 +1,153 @@
...
+
+    try:
+        compose_file = parse_compose_string(payload.decode("utf-8"))
```
johnmarsden-edu:
```suggestion
        compose_file = parse_compose_string(yaml_content)
```

trevnels-cs:
It looks like the tests need to be updated for the new function signature, but other than that it looks good

trevnels-cs (approved):

backend/ng/challenge/controllers/admin/lint_challenge.py
```diff
@@ -0,0 +1,149 @@
+from typing import NotRequired, TypedDict
```
RayofLightz, line 1: Consider putting this under the challenge directory

backend/ng/challenge/models/Hint.py
```diff
@@ -49,14 +52,15 @@ def serialize(self, team=None, include_admin_fields=False) -> SerializedHint:
...
-        data = {
+        data: SerializedHint = {
             "id": self.id,
             "name": self.name,
             "challenge_id": self.challenge_id,
             "preview": self.preview,
             "body": None,
             "deduction": self.deduction,
             "is_redeemed": False,
+            "index": self.index,
```
trevnels-cs: I don't think index needs to be included in the serialized versions of things, since it should only be used to order items in backend db queries

backend/ng/challenge/utils/challenge_yaml.py
```diff
@@ -0,0 +1,42 @@
...
+        self.question_id = question_id
+        self.template = template
+
+    def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
+        return {
+            "event_id": self.event_id,
+            "challenge_id": self.challenge_id,
+            "question_id": self.question_id,
+            "template": self.template.eval_str,
+        }
```
johnmarsden-edu, line 24: @trevnels-cs this was my approach to improving the serialization of the partial functions

trevnels-cs: Does this ensure that the env vars get rendered when they are passed into the container init? It looked like the variable was just being populated based on a toString instead of the rendered value

trevnels-cs: Yeah, it looks like `__call__` is doing that here, should be good. I'm still not sure about coupling the question ID to the seed though - should we have the ability to use randomized env variables that aren't necessarily tied to a question?

johnmarsden-edu: @fpayer that seems like a question for you

fpayer: For env vars specifically, there shouldn't be a `question_id` since the 1-1 correlation is between container-challenge. Multiple questions could share the same container. 

The `question_id` should only be used for the randomized questions/answers

johnmarsden-edu: I currently require every single environment variable be tied to a question. Because I didn't have an explicit way to differentiate between the two, though I guess I could theoretically make a differentiation. But that's not how it was designed earlier.

johnmarsden-edu: But now I'm changing that to address this particular requirement

backend/ng/conftest.py
```diff
@@ -1114,6 +1117,7 @@ def _factory(challenge: Challenge | None = None, answer_variable: ChallengeVaria
         else:
             defaults['answer_variable_id'] = answer_variable.id
         defaults.update(kwargs)
+        print(defaults)
```
RayofLightz: Nit: Remove the print

RayofLightz (approved):
This looks good, in the future a lot of these things should be separate smaller PRs.

backend/ng/conftest.py
johnmarsden-edu:
```suggestion
```

PR #230, opened 2025-10-06
Hotfix Add 404 handling, fix public endpoint, fix error message

droge91:


trevnels-cs (approved):

trevnels-cs (changes requested):
Notification test doesn't seem to be authenticating anymore - should investigate and make sure cookies are still being set properly

trevnels-cs (approved):

PR #231, opened 2025-10-07
Login page

### LINKED ISSUE #178: Set up frontend for login page
- [ ] Give user option to select between "Single Sign On" or "Expo Login"

### LINKED ISSUE #205: Logged in Redirection
- Make event listing and individual event pages public. Redirect from `/` to `/events` for unauthed users. Maybe also include a login prompt
- Move register button onto individual event pages for authed users? - can keep event listing page clean and consistent
- Public/authed leaderboard pages have some differences to account for

trevnels-cs:
Login page - directs users to SSO login, while also allowing "expo login" for standard user accounts using our own login endpoint.

Adds a basic RequireUser wrapper component that allows routes to redirect unauthenticated users to the login page.

Adds user name to the NavBar when logged in, and switches between Log In and Log Out in the profile dropdown depending on auth state.

Note that login/logout must perform browser navigation (doing a full refresh) instead of using react-router, since the csrfToken in window.init has to be updated when auth changes.

### DIFF
--- a/frontend/src/components/NavBar.tsx
+++ b/frontend/src/components/NavBar.tsx
@@ -1,14 +1,21 @@
-import { NavigationMenu } from 'radix-ui';
-import { NavLink, useLocation } from 'react-router';
-import { TbUserCircle } from 'react-icons/tb';
-import { twMerge } from 'tailwind-merge';
+import { apiMutation } from '@/fetchers';
+import { useCurrentUser } from '@/hooks/users';
 import ThemeToggle from 'components/ThemeToggle';
 import { useTheme } from 'next-themes';
+import { NavigationMenu } from 'radix-ui';
+import { TbUserCircle } from 'react-icons/tb';
+import { NavLink, useLocation } from 'react-router';
+import { twMerge } from 'tailwind-merge';
 
 export default function NavBar() {
   const logout = () => {
-    // perform a logout
+    apiMutation('/users/logout', {}, { method : 'POST' }).then(() => {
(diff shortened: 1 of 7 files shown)


frontend/src/components/NavBar.tsx
```diff
@@ -101,8 +108,9 @@ export default function NavBar() {
           </NavigationMenu.Item>
 
           <NavigationMenu.Item>
-            <NavigationMenu.Trigger className={twMerge(defaultLinkClass, 'pt-3')}>
-              <TbUserCircle />
+            <NavigationMenu.Trigger className={twMerge(defaultLinkClass)}>
```
cjones-cs: You removed the pt-3, so there's no need for twMerge here

frontend/src/routes/login/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <Card size="3">
+        <Flex direction="column" className="max-w-96" gap="3" align="center">
+          <Heading size="8">Log in</Heading>
+          <Text color="gray">
+            To continue, please log in with your organization&apos;s Single Sign-On (SSO) provider.
+          </Text>
+
+          <a href="/ng/authenticate/okta/login" className="w-full">
+            <Button size="4" className="!w-full">Log in with SSO</Button>
+          </a>
```
cjones-cs, line 24: This seems wrong

cjones-cs (changes requested):

cjones-cs (approved):

Issue #223, opened 2025-10-06
Gap between profile button and dropdown causes menu to close when moving mouse downward


trevnels-cs:
When hovering over the profile button, the associated dropdown will open, however as the mouse is moved slowly downward and gets to the gap between the button and menu, the dropdown will close again before the mouse can reach it to keep it open. This results in some odd flickering behavior, and can make it difficult to access the menu unless the mouse is moved quickly to reach the dropdown from the button before it closes.

<img width="162" height="278" alt="Image" src="https://github.com/user-attachments/assets/ac31f7d2-d4fe-4381-bece-a162a0bbcc2c" />

trevnels-cs:
Also take a look at why this dropdown is still transparent in some cases. It isn't on my end though?

cjones-cs:
Made the dropdowns not open on hover, must explicitly click to open the dropdowns. Same with notifications dropdown. I also wasn't able to reproduce the transparency issues, so we will just mark that as non-reproducable.

PR #237, opened 2025-10-07
Notifications in nav bar

cjones-cs:
Add notifications dropdown to nav bar.
Disable hover opening for both the notification and profile dropdowns
Handles issues: https://github.com/CyberSkyline/ctf-ng/issues/223

### DIFF
--- a/frontend/src/constants.ts
+++ b/frontend/src/constants.ts
@@ -66,16 +66,15 @@ export const ROLES: Record<string, { color: AccentColor; icon: IconType }> = {
   },
 };
 
-
 /* Find these const defined in NotificationType.py */
 export const NOTIF_TYPE = {
   TICKETS : [
-    "ticket_create",
-    "ticket_message",
-    "ticket_status_change",
-    "ticket_assigned",
+    'ticket_create',
+    'ticket_message',
+    'ticket_status_change',
+    'ticket_assigned',
   ],
-  TEAM: [
-    "team_invitation",
-  ]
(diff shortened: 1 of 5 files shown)


frontend/src/constants.ts
```diff
@@ -65,3 +65,16 @@ export const ROLES: Record<string, { color: AccentColor; icon: IconType }> = {
...
+/* Find these const defined in NotificationType.py */
+export const NOTIF_TYPE = {
+  TICKETS : [
+    'ticket_create',
+    'ticket_message',
+    'ticket_status_change',
+    'ticket_assigned',
+  ],
+  TEAM : [
+    'team_invitation',
```
trevnels-cs: This notification type seems to be unused - we don't have a way to invite users to teams within the platform itself

cjones-cs: It may be unused currently, but I included it to more closely match what the backend _can_ currently send. Since we decided that the backend will remove the challenge_released type, it was excluded. Also, announcements are handled separately as well. See `backend/ng/notifications/models/Notification.py` for current notif types listed on the model

trevnels-cs: While the type does exist on the backend, the code to send that type of notification is never used anywhere and I don't think there are any plans to, so I think that type should be removed as well.

cjones-cs: removed extra type

trevnels-cs (approved):

Issue #209, opened 2025-09-29
Accessibility - Fix "clickable" components


fpayer:
All clickable element must have:

- The `pointer` mouse cursor when hovered
- A tab-index so that the element can be tab-selected by the keyboard
- Can be triggered using the "enter" key when tab-selected

fpayer:
Tabs on the event page

<img width="755" height="107" alt="Image" src="https://github.com/user-attachments/assets/d4fc33af-98fa-4502-a842-801baa4760f7" />

fpayer:
The user icon in the top-right of the navbar

<img width="255" height="247" alt="Image" src="https://github.com/user-attachments/assets/8acfdbe7-c639-402b-ba77-d7c1e195d172" />

fpayer:
All of the action buttons/links on the team management page

<img width="1106" height="323" alt="Image" src="https://github.com/user-attachments/assets/715e1ca8-06ac-40c0-9e9b-101a09391c31" />

cjones-cs:
> The user icon in the top-right of the navbar
> 
[...]

PR #239, opened 2025-10-07
Set all active buttons to use pointer cursor

cjones-cs:
Changes buttons to cursor
Disabled buttons do not change to cursor
https://github.com/CyberSkyline/ctf-ng/issues/209


cjones-cs:
Closing: https://www.w3.org/WAI/ARIA/apg/patterns/button/examples/button/

PR #240, opened 2025-10-08
exec-service

RayofLightz:
Exec service for admins to exec into challenge containers

### DIFF
--- a/.github/workflows/services-linter.yml
+++ b/.github/workflows/services-linter.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Setup pnpm
       uses: pnpm/action-setup@v4
       with:
-        run_install: false
+        run_install: true
 
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v4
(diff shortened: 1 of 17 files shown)


.github/workflows/services-linter.yml
zoecough, line 1: this is probably skipping installing devDependencies because i can see eslint in there but the workflow is failing

zoecough (commented):
everything else looks good. will the exec service fail if TLS certs arent generated on install?

services/exec-service/server.ts
```diff
@@ -0,0 +1,44 @@
...
+interface DockerHeaderRequest extends Request {
+  headers : Headers & {
+    'docker-id' : string,
+  }
+}
+
+wss.on('connection', (socket: WebSocket, request: DockerHeaderRequest) => {
+  const { 'docker-id' : dockerId } = request.headers;
+
+  const docker = new Docker({
```
trevnels-cs, line 16: Do you think it would make sense for the docker client to be a singleton instead of creating a new one for each socket connection?

RayofLightz: Sorry just saw this feedback. I mainly did this just so any event emiitters/listeners created from the client and its associated streams get marked by the GC at the end of the connection. This is admittedly a bit of an over optimization so I can change it if you think it would be easier to read

zoecough (approved):

PR #242, opened 2025-10-08
Move registration to the event page, more auth checks

### LINKED ISSUE #205: Logged in Redirection

### LINKED ISSUE #229: Move registration button to event pages
# Event Header
For authenticated users:
- Show register button if registration is open and they are not registered
  - A `CAN_REGISTER` permission might be useful to track eligibility consistently with other permissions, though we currently have the eligibility endpoint for this.
- Show "Registered on [team name]"/"Registered as [leaderboard name]" if registered

For unauthenticated users
- If event hasn't ended, show a message like "To participate in this event, please log in." w/ link to login page

Will look something like this:
Image

trevnels-cs:
- Registration modal is now located on the event page
  - If registered, event page will show "Registered on [team name]" or "Registered as [leaderboard name]"
- Helpers for checking auth/registration state are now used to conditionally show controls
  - This PR is a first pass, more thorough permission checks to come soon
  - Unauthed/unregistered users only see leaderboard
- Added a few missing `mutate`s to ensure the UI updates immediately when performing server-side actions
- AvailableEvents now shows a grid of cards for all public events
  - Event cards include start date, time limit, max team size
  - Will be split into sections based on registration/event status in a later PR - needs a bit more planning that I don't want to block the other changes on


frontend/src/routes/events/Overview.tsx
```diff
@@ -1,23 +1,47 @@
...
   const [ searchParams, setSearchParams ] = useSearchParams();
   const { idEvent } = useParams();
-  const currentTab = searchParams.get('tab') ?? 'challenges';
 
-  const { data } = useEvent(Number(idEvent));
+  const { data, error } = useEvent(Number(idEvent));
+  const { isAuthenticated, isUnauthenticated } = useAuth();
+  const { isRegistered, isUnregistered } = useRegistration(Number(idEvent));
+
+  const challengesTabAvailable = isRegistered;
```
cjones-cs, line 26: Since there's only one condition here, I would normally suggest we remove this, but I think it adds clarity in the render function, so it's up to you

frontend/src/routes/events/RegistrationLine.tsx
```diff
@@ -0,0 +1,33 @@
...
+import type { Event } from '@/types';
+import { Text } from '@radix-ui/themes';
+import { TbCheck } from 'react-icons/tb';
+import RegistrationModal from './RegistrationModal';
+
+export default function RegistrationLine({ event }: {event: Event}) {
+  const { isRegistered, isUnregistered, team } = useRegistration(event.id);
+
+  // don't check eligibility if already registered since we know it will fail
+  const { data : eligibility, error : eligibilityError } = useMyEligibility(isUnregistered ? event.id : null);
```
cjones-cs, line 13: smart. If you pass null, then swr will ignore the request

frontend/src/routes/events/RegistrationLine.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  if (isRegistered) {
+    return (
+      <Text color={COLOR_POSITIVE}>
+        <TbCheck className="inline me-1" />
+        You are registered
+        {' '}
+        {event.max_team_size > 1 ? 'on' : 'as'}
```
cjones-cs: Can we change "on" to be "on team: " or "with team: ". I think it's important to denote explicitly if it's a team or not

cjones-cs (changes requested):
Minor nits

frontend/src/routes/events/Overview.tsx
trevnels-cs: I think it's better for clarity and consistency with how the team tab works - also if we were to add more conditions to showing the challenges tab than just registration in the future, having it already split out like this would be useful

cjones-cs (approved):

PR #244, opened 2025-10-09
updated notification logic

CarterPerez-dev:
Add delete notification
option to send notification with announcement
remove team invation type and method
add expired at param to endpoint
add type param to system announcement endpoint


fpayer:
Is there a way to make a "challenge updated" announcement separate from updating the yaml? I'm thinking about the case where someone forgets to create the announcement while they're updating the yaml and would like to make the announcement afterwards.

backend/ng/announcements/services/announcement_service.py
```diff
@@ -105,8 +112,25 @@ def send_system_announcement(
             title = title,
             message = message,
             sender_id = sender_id,
+            expires_at = expires_at,
         )
 
+        if send_notification:
+            all_users = Users.query.all()
```
fpayer: If the announcement is for a specific event, only users in that event should get a notification.

CarterPerez-dev:
> Is there a way to make a "challenge updated" announcement separate from updating the yaml? I'm thinking about the case where someone forgets to create the announcement while they're updating the yaml and would like to make the announcement afterwards.

Added a standalone controller.

fpayer (approved):
For the future - you should be doing a rebase to get the changes from the `development` branch into your feature branch. You did a merge here instead, which means that the diff view not only shows your changes, but also all of the changes from all of the commits that got merged in to `development` while you were working on this feature branch. This makes it super confusing to review cause I can't tell what changes you made (which need to be reviewed) vs someone else made (were already reviewed).

PR #248, opened 2025-10-09
Specify team timestamps as UTC

trevnels-cs:
Serialized team timestamps were missing Z suffix, so the frontend was not parsing them as `Date`.


droge91 (approved):
Looks Fine

PR #253, opened 2025-10-09
Hotfix for empty team cleanup issue

### LINKED ISSUE #287: Removing last member with /admin/teams/<id>/kick should disband team
If an admin kicks the last player on a team, the team should be disbanded instead of leaving a 0-member team in the system.

An alternative to this would be having a dedicated endpoint for admins to disband a team that would kick all players and remove the team.

droge91:
Fixes empty teams not being cleaned up in /leave as well as ensuring their corresponding scores get deleted
also fixes a couple duplicate route deletions in core.

Updated to also close #287


backend/ng/event/routes/user_routes.py
```diff
@@ -416,14 +416,16 @@ def post(self, event_id, event, team, current_user, **kwargs):
                 403,
             )
         try:
-            team_member.remove_team_member(commit = False)
+            team_member.remove_team_member(commit = True)
```
trevnels-cs: Ideally this would be done atomically instead of in two steps, so that if any piece fails the entire transaction won't go through. Is there a different way to make the below logic work without committing until the entire process is done?

droge91: We could potentially change the check to `len(team.members == 1` to check for teams that will be empty after the commit but we would also have to get rid of the protection on `team.disband_team` that prevents disbanding teams that still have a member since the member will still be there when its called

trevnels-cs: I think it makes sense to do that and allow `disband_team` to work regardless. You could also ensure that `disband_team`  cascades the delete to associated TeamMember objects and just call disband if member count is 1 to delete the team and remaining member in a single operation. This may also prove useful for #287

backend/ng/team/controllers/remove_member.py
```diff
@@ -0,0 +1,21 @@
...
+    Remove a member from the team and regenerate the team's join code.
+    """
+    user = user.ctfd_user if hasattr(user, "ctfd_user") else user
+    team_member = TeamMember.find_by_user_and_team(user.id, team.id)
+    team_member.remove_team_member(commit = False)
+
+    demographic = Demographic.find_by_user_and_event(user.id, team.event_id)
+    demographic.delete(commit = False)
+    if len(team.members) == 1:
+        Score.query.filter(Score.team_id == team.id).delete()
```
trevnels-cs: I believe this should also have `commit = False` to keep everything in a single transaction.

Also, consider moving Score deletion into the `disband_team` function instead of doing it here

trevnels-cs (approved):

PR #254, opened 2025-10-10
added decorator to oauth routes and set up swarm for prodmode

zoecough:
I removed "staging" specific stuff since this config works for staging and anything additional that's prod-specific will be handled by the additional prod infra repo rather than within this one.


RayofLightz (approved):
Looks good but where is the actual swarm setup?

PR #257, opened 2025-10-10
Select component to use with react hook forms

cjones-cs:
A standard select component for use with the new react hook forms

### DIFF
--- a/frontend/src/components/SelectDropdown.tsx
+++ b/frontend/src/components/SelectDropdown.tsx
@@ -1,0 +1,77 @@
+import { Flex, Select } from '@radix-ui/themes';
+import { map } from 'lodash';
+import {
+  Controller,
+  type Control,
+  type FieldValues,
+  type FieldError,
+  type RegisterOptions,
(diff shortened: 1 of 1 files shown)


frontend/src/components/SelectDropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+    <FormField label={label} error={error}>
+      {(injected) => (
+        <Controller
+          name={name}
+          control={control}
+          rules={rules}
+          render={({ field }) => (
+            <Select.Root
+              value={field.value}
+              onValueChange={field.onChange}
```
trevnels-cs, line 62: Might be wise to map the "None" value here to undefined/null at the RHF level for easier integration with APIs and other systems

- If field.value is null/undefined, send "None" to the Select's value
- If the value passed to onValueChange is "None", pass null/undefined to the controller's onChange function

This would also allow cleanly using `required` on the controller, etc.

trevnels-cs (approved):

PR #262, opened 2025-10-10
Fix Event Detail 404 bug

### LINKED ISSUE #234: Accessing private events always results in a 404, even for registered users/admins
Admin users are unable to open/access private events, even when they are registered for one - the endpoint comes back with a 404 regardless. This behavior can be observed with event ID 2 in the sample data - even though the admin is registered for it and it shows on their dashboard, event info can't be accessed via the /events/[id] endpoint.

Either private events should be removed from `/users/me/events`, `/users/me/teams`, and other event endpoints so that they are completely inaccessible, or private events should be available to admins and users that are registered for them. I think the latter makes sense, i.e. to allow admins to test events and challenges before opening them to the public.

droge91:
Fixes Admins and registered users no longer being able to access private event details, Closes #234

### DIFF
--- a/backend/ng/core/middleware/permission_middleware.py
+++ b/backend/ng/core/middleware/permission_middleware.py
@@ -52,7 +52,7 @@ def event_only_public(f):
     @wraps(f)
     def wrapped(*args, **kwargs):
         event = kwargs.get('event')
-        current_user = kwargs.get('current_user') or get_current_user()
+        current_user = kwargs.get('current_user')
         if current_user and RoleEnum.ADMIN in get_user_roles(current_user.id):
             return f(*args, **kwargs)
         id = current_user.id if current_user else None
--- a/backend/ng/core/middleware/permission_middleware.py
+++ b/backend/ng/core/middleware/permission_middleware.py
@@ -5,8 +5,9 @@ from ..utils.logger import get_logger
 from ...permissions.controllers.get_team_management_permissions import get_team_management_permissions
 from ...permissions.controllers.get_user_permissions import get_user_permissions
 from ...permissions.controllers.get_challenge_permissions import get_challenge_permissions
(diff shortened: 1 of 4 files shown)


backend/ng/core/middleware/permission_middleware.py
```diff
@@ -51,7 +52,9 @@ def event_only_public(f):
     @wraps(f)
     def wrapped(*args, **kwargs):
         event = kwargs.get('event')
-        current_user = kwargs.get('current_user')
+        current_user = kwargs.get('current_user') or get_current_user()
+        if current_user and RoleEnum.ADMIN in get_user_roles(current_user.id):
+            return f(*args, **kwargs)
         id = current_user.id if current_user else None
         if Demographic.find_by_user_and_event(id, event.id) is None:
```
trevnels-cs, line 59: Something in this logic still seems to be off - the admin registered for the private event by the setup script is getting a 404, even though the existing logic here should have allowed access for registered users, regardless of role. I don't think adding the role check is a complete solution here, since there still seems to be an underlying issue with the Demographic check.

droge91: what are you getting the 404 on? /events/id? I'm not able to get it to 404 in my testing. the main issue that was fixed was that the endpoint was now public so `current_user` was no longer being injected so it needed to be retrieved if not present.

trevnels-cs: Ah, that makes sense. Do you think it would make sense to have `public_endpoint` still inject `current_user` if it's present, so that the middleware functions that may be chained after it won't need to have the `or get_current_user()`?

droge91: Yeah I like that better, That way we don't have deal with it separately for any more middleware that might need it

trevnels-cs (approved):

Issue #173, opened 2025-09-22
Add support for image uploads to AWS S3 on support tickets


fpayer:
Create a backend endpoint to allow image upload. Don't worry about frontend - that will be handled later. 

- Endpoint should have a configurable MAX_FILESIZE. Default should be 5MB
- Endpoint should ONLY accept webp filetypes (file will be converted on the frontend)
- Test cases should attempt to upload different test files to verify the filetype + filesize requirements.

PR #263, opened 2025-10-14
aws upload support tickets

CarterPerez-dev:
AWS S3 image upload support for tickets
AWS S3 image upload service, controller, model, test files
Test script to test image upload to s3 bucket
closes issue: https://github.com/CyberSkyline/ctf-ng/issues/173


backend/ng/support/controllers/__init__.py
```diff
@@ -10,6 +10,7 @@
     create_ticket_message,
     get_ticket,
     list_tickets,
+    upload_ticket_image,
```
fpayer: rename to `upload_ticket_attachment`. Will allow us to future-proof in case we allow other file types in the future.

backend/ng/support/controllers/all_actions/upload_ticket_image.py
```diff
@@ -0,0 +1,78 @@
...
+
+from .... import config
+from ....core.exceptions import ValidationError
+
+from ...models import Ticket, TicketAttachment
+from ...services import get_s3_upload_service
+
+
+
+def upload_ticket_image(
```
fpayer: Rename to `upload_ticket_attachment`

backend/ng/support/controllers/all_actions/upload_ticket_image.py
```diff
@@ -0,0 +1,78 @@
...
+
+    file_ext = file.filename.lower().split('.')[-1] if '.' in file.filename else ''
+    if file_ext not in config.TICKET_IMAGE_ALLOWED_TYPES:
+        allowed_types_str = ', '.join(f".{t}" for t in config.TICKET_IMAGE_ALLOWED_TYPES)
+        raise ValidationError(
+            f"Only {allowed_types_str} files are allowed",
+            errors={"file": f"File must be in one of these formats: {allowed_types_str}"}
+        )
+
+    file.seek(0, 2)
```
fpayer: Is there a better way to determine filesize? This seems sketchy to me.

backend/ng/config.py
```diff
@@ -23,6 +23,16 @@
 TICKET_TAG_DESCRIPTION_MAX_LENGTH = 200
 TICKET_TAG_COLOR_MAX_LENGTH = 7 # Hex color code like #FF0000
 
+# Support ticket image upload config
+TICKET_IMAGE_MAX_SIZE = 5 * 1024 * 1024  # 5MB in bytes
+TICKET_IMAGE_ALLOWED_TYPES = ['webp']
```
fpayer: Rename to `TICKET_ATTACHMENT_ALLOWED_TYPES`

backend/ng/config.py
```diff
@@ -23,6 +23,16 @@
 TICKET_TAG_DESCRIPTION_MAX_LENGTH = 200
 TICKET_TAG_COLOR_MAX_LENGTH = 7 # Hex color code like #FF0000
 
+# Support ticket image upload config
+TICKET_IMAGE_MAX_SIZE = 5 * 1024 * 1024  # 5MB in bytes
```
fpayer: Rename to `TICKET_ATTACHMENT_MAX_SIZE`

backend/ng/support/controllers/all_actions/upload_ticket_image.py
```diff
@@ -0,0 +1,78 @@
...
+        file_extension=file_ext,
+    )
+
+    if not upload_result:
+        raise ValidationError(
+            "Failed to upload image. Please try again.",
+            errors={"file": "Upload failed"}
+        )
+
+    content_type = f"image/{file_ext}"
```
fpayer: Is there a way to automatically determine this without relying on the file extension?

backend/ng/support/services/test-upload-img.webp
fpayer, line 1: Can you find a plain image file for the test? This is going into a gov project.

fpayer (changes requested):
Instead of using the s3 presign url, create an endpoint for the webapp to connect to s3 and return the contents of the file (acting as a proxy to S3). Otherwise, we'll run into issues down the line with the presign URLs expiring. 

Break out the `TicketAttachment` model into two separate ones.

- A generic `FileUpload` model the file-specific info
- The `TicketAttachment` model to map the FileUpload to the specific support ticket. You might not even need a separate model for this if you can fit it directly within the existing Ticket model. 

Something to keep in mind - we need to enforce permission checks on accessing the S3 files. For the support tickets, we don't want one player being able to see the support ticket file uploads for another players. There should be some field on the `FileUpload` model to specify that URL to the s3 proxy endpoint for this file. That way, we can have multiple different s3 proxy endpoints with different levels of permission checks.

fpayer (approved):

Issue #261, opened 2025-10-10
Notification with a removed type results in a 500 for the notification list


trevnels-cs:
Since we removed `CHALLENGE_RELEASED` as a notification type, I'm now permanently getting a 500 from the notifications endpoint with `LookupError: 'CHALLENGE_RELEASED' is not among the defined enum values. Enum name: notificationtype. Possible values: TICKET_CREA.., TICKET_MESS.., TICKET_STAT.., ..., EVENT_ANNOU..`. Clearing them from the db will resolve this for now, but we should ensure that this situation is handled gracefully by the endpoint in the future if we ever need to remove or update the enum.

PR #271, opened 2025-10-16
fix 500 error from removed enum type

CarterPerez-dev:
## Closes issue https://github.com/CyberSkyline/ctf-ng/issues/261
- Updated Notification and Announcement models to fetch IDs first, then load objects individually with try-catch to skip invalid enums
- Added UNKNOWN enum value as a fallback for future enum removals
- Added tests that simulate removed enum values and handle gracefully


backend/ng/notifications/models/Notification.py
```diff
@@ -391,17 +408,43 @@ def find_filtered_notifications(
...
         ).filter((cls.expires_at.is_(None))
-                 | (cls.expires_at > utc_now())).count()
+                 | (cls.expires_at > utc_now()))
+
+        count = 0
+        notification_ids = [n[0] for n in query.with_entities(cls.id).all()]
+
+        for notification_id in notification_ids:
+            try:
+                notification = cls.query.get(notification_id)
```
trevnels-cs: I'm not sure about this solution performance-wise, since running a single query to retrieve all rows is significantly faster than querying each row sequentially like this. Consider something like the solution discussed at the end of https://blog.wrouesnel.com/posts/sqlalchemy-enums-careful-what-goes-into-the-database/ - wrapping the Enum type to let SQLAlchemy handle unknowns within a single large query

fpayer:
Going to defer to Trevor's review on this.

trevnels-cs:
Is the Notification E2E test failure a residual of when those tests were broken, or is that an issue introduced by this change? May be worth syncing this branch up to current development to verify.

CarterPerez-dev:
> Is the Notification E2E test failure a residual of when those tests were broken, or is that an issue introduced by this change? May be worth syncing this branch up to current development to verify.

Good call, I initially assumed it was from a residual of when the tests were broken, which still might've been, but when I synced with development there actually was a bug introduced with this change causing it to fail as well.

trevnels-cs (approved):

Issue #259, opened 2025-10-10
Field Validation incomplete


cjones-cs:
Milkdown had an error, but I found this bug. The backend is still making a db record for a new ticket, even if not all the required fields are in place. For this example, the message field was empty (bc milkdown was broken)

<img width="1181" height="431" alt="Image" src="https://github.com/user-attachments/assets/2729eeb3-80e6-4dfb-b96a-f78410d6843d" />

<img width="1181" height="431" alt="Image" src="https://github.com/user-attachments/assets/0464ad66-1d0d-46f6-b06a-d250cb974d7b" />

trevnels-cs:
This is because the ticket object is created first and committed to the db, then the message is added afterwards - the message doesn't get validated until the message object is created. This needs to instead be done in a single transaction instead of committing the ticket before we know whether the message was added successfully or not.

PR #272, opened 2025-10-16
fix db record created when not all fields in place

CarterPerez-dev:
## Closes issue https://github.com/CyberSkyline/ctf-ng/issues/259
- Created atomic model method to solve the issue of db record being created when message is empty


RayofLightz (approved):

RayofLightz:
Wait for @cjones-cs to take a look as well

backend/ng/support/models/Ticket.py
```diff
@@ -229,6 +229,76 @@ def create_ticket(
 
         return ticket
 
+    @classmethod
+    def create_ticket_with_initial_message(
```
cjones-cs: Why are you creating a new function for this instead of updating the create_ticket function to properly validate required fields?

cjones-cs (changes requested):
Needs explaination

backend/ng/support/models/Ticket.py
CarterPerez-dev: Good question, the problem isn't validation itself - it's that we were doing two separate database operations. It's a transaction timing issue. The controller was calling `create_ticket(commit=True)` which immediately saves the ticket, then calling `add_message()` separately. If the message part failed, the ticket was already in the database. Basically, the create_ticket controller used the `@classmethod create_ticket()` model method *then* the `@classmethod add_message()` model method, so it was two different functions that made up the  `create_ticket.py ` controller in order to create the ticket object + a message object when a user creates a ticket (because you shouldn't be able to just create a ticket with no message)

This was being done non atomically which caused the issue you had of how a ticket object/db record was created when the message field (seperate function) wasn't present. Think of it like this - when they are two different functions, I can validate if all required fields are present for *that* function itself, but I can't just validate  if both fields from different functions *together* are present.
So heres a simple example of what I mean:

- function1 needs field1 required/present
- function2 needs field2  required/present

So I already was validating if all required fields are present, but for the function/s itself, but how do I make sure field1 AND field2 are BOTH present, and if not, then dont commit/create db record? I have to create an atomic transaction which essentially automatically requires both fields to be present because its now one transaction...

So, I *could* techinically do some validation code in the controller, but that simply would have just been creating the same kinda code that the `@classmethod def create_ticket_with_initial_message()` does, but in a much messier/hacky way, therefore the logical solution which makes the most sense was to just make an atomic method in the model, then call it in the controller. And if you're then wondering why I didn't just replace the existing model methods is because those are used by themselves in other controllers, hence, adding another model method that combines them as an atomic method regardless if its just used for one controller made the cleanest sense.

backend/ng/support/models/Ticket.py
```diff
@@ -229,6 +229,76 @@ def create_ticket(
...
+            author_id=validated_ticket_data["author_id"],
+            event_id=validated_ticket_data.get("event_id"),
+            team_id=validated_ticket_data.get("team_id"),
+            challenge_id=validated_ticket_data.get("challenge_id"),
+        )
+
+        message = TicketMessage(
+            text=validated_message_data["text"],
+            author_id=author_id,
+        )
```
RayofLightz: It looks like you are just copy pasting the create functions of these. Call the create functions and pass in `Commit=False`

backend/ng/support/controllers/user_actions/create_ticket.py
```diff
@@ -27,30 +27,21 @@ def create_ticket(
...
         challenge_id=challenge_id,
-        commit=True,
-    )
-
-    ticket.add_message(
-        text=text,
-        author_id=current_user.id,
-        is_admin=False,
-        commit=True,
     )
```
RayofLightz: You could just pass `commit=False` to these and save the session/rollback on exception/success later in the code.

RayofLightz (changes requested):
I took a second look after looking at some of the other tickets code and there are some ways that we could improve how we are doing this.

backend/ng/support/controllers/user_actions/create_ticket.py
```diff
@@ -27,14 +30,33 @@ def create_ticket(
...
+            author_id=current_user.id,
+            event_id=event_id,
+            team_id=team_id,
+            challenge_id=challenge_id,
+            commit=False,
+        )
+
+        # Flush to assign ticket ID
+        db.session.flush()
+
```
RayofLightz: Consider adding the flush to create_ticket method

RayofLightz (approved):
One last comment worth considering

cjones-cs (approved):

PR #273, opened 2025-10-16
remove challenge updated notification

CarterPerez-dev:
Removed challenge updated notification


cjones-cs (changes requested):
Were going to remove challenge updated since we can use the standard event announcement to perform an announcement/notif on this instead.

cjones-cs (approved):

Issue #251, opened 2025-10-09
Page titles


trevnels-cs:
Pages currently don't have titles and just show their URL in the tab bar.

PR #276, opened 2025-10-16
Feature/titles

cjones-cs:
https://github.com/CyberSkyline/ctf-ng/issues/251


frontend/src/routes/login/index.tsx
```diff
@@ -13,6 +13,7 @@ import ExpoLoginForm from './ExpoLoginForm';
 export default function Login() {
   return (
     <Flex className="absolute inset-0 overflow-hidden bg-dots-1" align="center" justify="center" direction="column" gap="3" p="3">
+      <title>President&apos;s Cup - Login</title>
```
trevnels-cs: Should the login page still have President's Cup in the title when the other pages don't? For consistency, I think we should either remove it here or add it to the other page titles

trevnels-cs (approved):

PR #278, opened 2025-10-16
Ctfd routeflatten

### LINKED ISSUE #206: API - Flatten endpoints
LOW PRIORITY - This is a cleanup item we can handle after the initial game 
- flatten endpoints to account for automatic associations (team to event, etc)
    - especially a problem for challenge ids
    - Adjust/make proper middleware checks for these association checks
      - i.e. `load_event_by_challenge` - already implemented in a stash
- Update frontend endpoints to match

- EX: `/admin/scoring/events/{event_id}/teams/{team_id}/history` ⇒ `/admin/scoring/teams/{team_id}/history`

droge91:
Flattens many API routes to be less verbose
Also had to refactor a bit of scoring to get it to work
Modifies frontend hooks for the changed routes

Closes #206 

Not going to be merged in until V2


frontend/src/hooks/challenge.ts
droge91, line 1: I deleted the submitAnswer function as it appeared to be superseded by submitFlag and it wasn't used anywhere that I could see. Let me know if it should stay

backend/ng/event/routes/user_routes.py
```diff
@@ -477,12 +478,11 @@ def get(self, event_id: int, event: Event, permissions, **kwargs):
         return success_response(challenges)
 
 
-@events_user_namespace.route("/<int:event_id>/challenges/<int:challenge_id>")
+@events_user_namespace.route("/challenges/<int:challenge_id>")
```
trevnels-cs: Since these routes are no longer parameterized by event, I think it would make sense to move them to the challenges namespace instead. `/events/challenges/x` -> `/challenges/x`

backend/ng/scoring/routes/user_routes.py
```diff
@@ -81,22 +82,20 @@ class MyTeamScore(Resource):
...
     @load_event(LoaderType.PARAM)
     @load_team_by_user_and_event()
-    @load_score_by_team_and_event()
-    def get(self, event_id: int, event, team, score, current_user: User, **kwargs):
+    @load_score_by_team()
+    def get(self, team, score, current_user: User, **kwargs):
         """
         Get my team's score
         """
         result = get_team_score(score=score)
```
trevnels-cs: `get_team_score` just returns the same score object that was passed in, so we can just return `score` directly here

frontend/src/hooks/challenge.ts
```diff
@@ -23,60 +23,41 @@ export function useMyChallenges(eventId: number | null) {
...
-  });
-}
-
 export function redeemHint(
-  eventId: number,
   challengeId: number,
   hintId: number,
 ) {
-  return apiMutation(`/events/${eventId}/challenges/${challengeId}/hint/${hintId}/redeem`, undefined, {
+  return apiMutation(`/hint/${hintId}/redeem`, undefined, {
```
trevnels-cs: Should this be `/scoring/hint/${hintId}/redeem` since that endpoint is in the scoring namespace?

backend/ng/event/routes/user_routes.py
droge91: agreed, I almost did that initially

PR #299, opened 2025-10-21
Challenge update modal and YAML download

### LINKED ISSUE #221: Integrate challenge update functionality
Now that the backend supports updating challenges, we need to:
- [ ] Expose that functionality as a PUT request
- [ ] Create UI in the challenge admin page to upload new YAML for an existing challenge that uses the endpoint
- [ ] Set up `challenge_updated` notification to notify users of updates made *only after they have started the event*

trevnels-cs:
Closes #221.

Modal for updating existing challenges in the system, as well as a button to download the YAML content of a challenge as a file. Also includes a backend fix to properly update container blueprints based on name matching instead of creating new ones with each update.

image

image

The modal uses `react-codemirror` for YAML editing, which handles highlighting, indentation, etc. - we can eventually connect to our challenge linting output to codemirror's parse tree for lint highlighting as well. The editor is wrapped with `react-dropzone` to allow dragging files in, as well as selecting a file via the upload button in the corner.

As a code editor, codemirror captures tab events when focused for indentation purposes - to keyboard-navigate out of codemirror, users need to press Ctrl+M first to switch from tab indentation to navigation.

This PR also reworks the existing challenge upload modal to use this same editor for consistency. Eventually the two will also be able to share linting logic.

### DIFF
--- a/frontend/src/components/Modal.tsx
+++ b/frontend/src/components/Modal.tsx
@@ -72,6 +72,8 @@ export default function Modal<T extends FieldValues>({
 
       <Dialog.Content
         className={twMerge('flex flex-col gap-3', className)}
+        // clear tabIndex to ensure compatibility with CodeMirror's focus management
+        tabIndex={undefined}
         // aria-describedby should be set to undefined if no description is provided, otherwise it should not be set at all.
         // as far as i know, prop spreading is the only way to accomplish this.
         {
--- a/backend/ng/challenge/models/ContainerBlueprint.py
+++ b/backend/ng/challenge/models/ContainerBlueprint.py
@@ -44,6 +44,7 @@ class ContainerBlueprint(db.Model):
         """
         return {
             "id": self.id,
+            "name": self.name,
             "image": self.image,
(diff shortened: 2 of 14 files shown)


johnmarsden-edu (approved):
The logic looks good to me

frontend/src/components/Modal.tsx
```diff
@@ -72,6 +72,8 @@ export default function Modal<T extends FieldValues>({
 
       <Dialog.Content
         className={twMerge('flex flex-col gap-3', className)}
+        // clear tabIndex to ensure compatibility with CodeMirror's focus management
+        tabIndex={undefined}
```
cjones-cs, line 76: Just to confirm. This doesn't interfere with other modals right?

cjones-cs (approved):

frontend/src/components/Modal.tsx
trevnels-cs: I tested keyboard navigation inside other modals, and it continues to work as expected for me after making this change

PR #302, opened 2025-10-23
email domain checks backend

### LINKED ISSUE #293: Implement backend for email domain checks on event registration
Update the Event model to keep a list of allowed domains that users must in order to register for the event. Make sure to update the serialize function to include that new list. 

Update the event registration endpoint to validate that the user has a domain that is allowed for the event. If no list is provided, then any email address is allowed.

droge91:
Adds allowed_domains to the event model
Now checks if the users domain is eligible during the eligibility check for registration
adds allowed_domains to serialization

Closes #293


backend/ng/event/models/Event.py
```diff
@@ -384,4 +395,13 @@ def check_eligibility(self, user):
         if Demographic.find_by_user_and_event(user.id, self.id):
             raise BusinessLogicError("User is already registered for this event.")
 
+        if self.allowed_domains:
+            user_email_domain = user.ctfd_user.email.split("@")[-1].lower()
+            allowed_domains_lower = [domain.lower() for domain in self.allowed_domains]
+            for domain in allowed_domains_lower:
+                if domain in user_email_domain:
+                    break
+            else:
```
fpayer, line 404: Should this `else` be here?

backend/ng/event/models/Event.py
```diff
@@ -384,4 +395,13 @@ def check_eligibility(self, user):
         if Demographic.find_by_user_and_event(user.id, self.id):
             raise BusinessLogicError("User is already registered for this event.")
 
+        if self.allowed_domains:
+            user_email_domain = user.ctfd_user.email.split("@")[-1].lower()
+            allowed_domains_lower = [domain.lower() for domain in self.allowed_domains]
```
fpayer, line 400: Need to either update the model to also ensure that all of the strings in `allowed_domains` will always be lowercase or also call `lower` on each domain in `allowed_domains`.

fpayer (changes requested):

backend/ng/event/models/Event.py
droge91: yeah the else will be hit if the for loop finishes without a break meaning that the users domain is not allowed.
`for`

`else:`

is valid python its just not used that often

backend/ng/event/models/Event.py
droge91: I don't understand what you mean, allowed_domains_lower is what is used in the check and it has been lowercased.

fpayer: You're right. I must have glazed over. No issue here

fpayer (approved):

PR #303, opened 2025-10-23
Fix current challenge exception when workspace has not yet been created

trevnels-cs:
Current challenge endpoint comes back with an error while the user workspace is being created, which results in error message flicker on the challenge page. Instead, the endpoint should indicate that no challenge is connected for the user (which is true, since the workspace doesn't exist yet.)


RayofLightz (approved):
Good catch and thank you for the hot fix

Issue #280, opened 2025-10-17
Implement sponsor backend


fpayer:
Create a "sponsor" model. It will need fields for
- Name
- URL to logo image

Create some sort of db record to track the sponsor for each user (investigate the best way of doing this)

We will also need to create API endpoints in order to:
- [x] Create a new sponsor organization
- [x] Edit an existing sponsor organization
- [ ] Perform a text-search of available sponsor organizations
- [x] Update a user's sponsor

PR #304, opened 2025-10-24
Sponsor Backend

droge91:
Adds sponsor model 
Routes for sponsor creation ,editing, assigning to user, and searching existing sponsors

part of #280

### DIFF
--- a/backend/ng/sponsors/models/Sponsor.py
+++ b/backend/ng/sponsors/models/Sponsor.py
@@ -1,0 +1,73 @@
+from CTFd.models import db
+from typing import Any
+from ...core.utils.validator import BaseValidator
+
+
+class Sponsor(db.Model):
+    __tablename__ = "ng_sponsors"
+
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(128), nullable=False)
+    logo = db.Column(db.String(256), nullable=True)
+
+    users = db.relationship("User", back_populates="affiliation", cascade="all, delete-orphan")
+
+
+    def __repr__(self):
+        return f"<Sponsor id={self.id} name={self.name}>"
+
+    def serialize(self, include_admin_fields = False) -> dict[str, Any]:
+        data =  {
+            "id": self.id,
+            "name": self.name,
+        }
+        if include_admin_fields:
+            data["logo"] = self.logo
+        return data
+
+    @classmethod
+    def create_sponsor(cls, name, logo=None, commit=True):
+        sponsor = cls(name=name, logo=logo)
+        db.session.add(sponsor)
+        if commit:
+            db.session.commit()
+        return sponsor
+
+    def update(self, commit=True, **kwargs):
+
+        for key, value in kwargs.items():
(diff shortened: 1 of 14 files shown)


backend/ng/sponsors/models/Sponsor.py
```diff
@@ -0,0 +1,73 @@
...
+    def __repr__(self):
+        return f"<Sponsor id={self.id} name={self.name}>"
+
+    def serialize(self, include_admin_fields = False) -> dict[str, Any]:
+        data =  {
+            "id": self.id,
+            "name": self.name,
+        }
+        if include_admin_fields:
+            data["logo"] = self.logo
```
trevnels-cs: I believe logo should not be an admin field, as they will need to be user-accessible to display on the leaderboard

backend/ng/sponsors/models/Sponsor.py
```diff
@@ -0,0 +1,73 @@
+from CTFd.models import db
+from typing import Any
+from ...core.utils.validator import BaseValidator
+
+
+class Sponsor(db.Model):
+    __tablename__ = "ng_sponsors"
+
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(128), nullable=False)
```
trevnels-cs, line 10: Would it make sense to have both name and abbreviation fields for sponsors? i.e. "Cybersecurity and Infrastructure Security Agency" and "CISA" could both be stored under a sponsor. @fpayer Any thoughts on this?

backend/ng/user/routes/user_routes.py
```diff
@@ -66,6 +67,60 @@ def get(self, current_user, **kwargs):
...
+
+    )
+    def put(self, current_user, json_data, **kwargs):
+        """Update my sponsor information"""
+
+        sponsor = json_data.get("sponsor")
+        if not sponsor:
+            return {"success": False, "errors": {"sponsor": "Sponsor data is required"}}, 400
+
+        sponsor = Sponsor.find_by_name(sponsor)
```
trevnels-cs: This endpoint should probably take an ID instead of a name

backend/ng/user/models/User.py
```diff
@@ -53,6 +60,7 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
                 "email": "",
                 "role": "unknown",
                 "registered_at": "",
+                "affiliation": self.affiliation.name if self.affiliation else "",
```
trevnels-cs: Might want to include both ID and name here - the frontend may want to display the logo etc. from this data, which would be easier to do based on an ID.

We could also just include the entire affiliation object under the affiliation field (with id, name, and logo) - and have the field be `null` if no affiliation is present.

backend/ng/sponsors/models/Sponsor.py
droge91: Very easily implemented it there's a use for it, this is more up to how exactly we want the sponsors to be displayed on the leaderboard and other places.

trevnels-cs (approved):

PR #311, opened 2025-10-28
Challenge attempt history grid

### LINKED ISSUE #195: Challenges
- Ability to view past submissions for each question
- Restart/recycle menu for users
- Permission checks for this and events

trevnels-cs:
Removes unused and placeholder feedback modal in the challenge header and replaces it with the attempt history modal, which contains a grid of past attempts. Could also incorporate hint redemptions here eventually if the necessary route is created.

Also moves to a general-purpose "activity type" badge component which is shared between the admin and user grids.

image

image


frontend/src/components/TeamActivityTypeBadge.tsx
```diff
@@ -0,0 +1,35 @@
...
+  COLOR_HINT,
+  COLOR_NEGATIVE,
+  COLOR_POSITIVE,
+  COLOR_WARNING,
+  type AccentColor,
+} from '@/constants';
+import type { Attempt, HintRedemption, ManualPointAward } from '@/types';
+import { Badge } from '@radix-ui/themes';
+
+export default function TeamActivityTypeBadge({ data }: { data: Attempt | HintRedemption | ManualPointAward | undefined }) {
```
cjones-cs, line 11: Is this also used for individual game? If so, we should remove the Team part of the name to eliminate confusion

cjones-cs (approved):

frontend/src/components/TeamActivityTypeBadge.tsx
trevnels-cs: Yes, however attempts, hints, manual awards, etc. are always associated with the `Team` object in our implementation (even if it's just a team of one,) so I would still consider them to be "team activity." The associated user-facing text does change depending on the context though (i.e. "your activity" vs "your team's activity" in the modal.)

PR #313, opened 2025-10-28
Hint redemption UI, respect event hints_enabled flag for endpoints

### LINKED ISSUE #246: Don't return hints during challenge render if event has hints disabled
Currently, if an event has `hints_enabled` set to false, hints attached to challenges will still be sent to users when rendering a challenge. The user-facing challenge endpoint should instead strip these out, and other hint-related endpoints (i.e. hint redemption) should not be usable in this case.

trevnels-cs:
Can now redeem hints through the hint modal.

image

Also resolves #246 - hints are not sent to clients and cannot be redeemed if their event has `hints_enabled` set to False. This also has the side effect of hiding the hints button, since no hints are present in this situation.

### DIFF
--- a/backend/ng/challenge/models/Challenge.py
+++ b/backend/ng/challenge/models/Challenge.py
@@ -175,7 +175,7 @@ class Challenge(db.Model):
         data = {
             "challenge": self.serialize(),
             "questions": Question.query.filter_by(challenge_id=self.id).all(),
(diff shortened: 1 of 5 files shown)


cjones-cs (changes requested):
lgtm, but the test runner is failing, so sending that back in case there are additional related changes

cjones-cs (approved):

PR #319, opened 2025-10-30
Team Related Bugfixes

### LINKED ISSUE #258: Consider clamping team end times to event end time
If a team starts an event near the end of the event's time window, their end timestamp can end up falling after the event's end, which may result in some weird logic bugs depending on which of these fields are checked. We should either be absolutely sure that the permission checks and UI can handle this case with no weird edge cases, or prevent this situation from happening at all by not letting team end timestamps exceed their event's.

### LINKED ISSUE #277: Users can't re-register for the same team they previously left
If a user is removed from or leaves a team and rejoins again via invite code, the register endpoint produces a 409 error instead of allowing them to rejoin. This is because the system is trying to create a Demographic entry that already exists: `{"timestamp": "2025-10-16T20:28:36.084963Z", "level": "ERROR", "message": "Database Integrity Error: A constraint was violated.", "module": "error_handler.py", "function": "decorated_function", "context": {"error": "(1062, \"Duplicate entry '2-2' for key 'uq_demographic_user_event'\")"}}`

### LINKED ISSUE #281: Can't set team timestamps via update_team
The admin update team route should be able to adjust start/end times for a team as needed, however the validator for Team is missing these fields, so including them in the request body does not result in them being changed.

droge91:
Closes #258, Closes #277, Closes #281

Modifies `Team.set_end_time` to clamp a teams end time to the events end time if it would exceed it
Adds` start_timestamp` and `end_time` to the team validator so admins can modify them if needed
#277 was accidentally fixed in #253 but added tests here to verify functionality so closing with this PR

### DIFF
--- a/backend/ng/team/models/Team.py
+++ b/backend/ng/team/models/Team.py
@@ -152,6 +152,10 @@ class Team(db.Model):
             )
             if res:
                 raise ValidationError(f"Team name '{data['name']}' already exists in event ID {data['event_id']}.")
+        if "start_timestamp" in data:
+            validator.validate_datetime(data, "start_timestamp", friendly_name="Start timestamp")
+        if "end_time" in data:
+            validator.validate_datetime(data, "end_time", friendly_name="End time")
 
         return validator.validate()
 
--- a/backend/ng/team/models/Team.py
+++ b/backend/ng/team/models/Team.py
@@ -152,10 +152,12 @@ class Team(db.Model):
             )
             if res:
                 raise ValidationError(f"Team name '{data['name']}' already exists in event ID {data['event_id']}.")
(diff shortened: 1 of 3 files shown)


backend/ng/team/models/Team.py
```diff
@@ -152,6 +152,10 @@ def validate(cls, data: dict[str, Any]) -> dict[str, Any]:
             )
             if res:
                 raise ValidationError(f"Team name '{data['name']}' already exists in event ID {data['event_id']}.")
+        if "start_timestamp" in data:
```
trevnels-cs: Should we use a datetime range validator here to ensure that end can't be before start?

trevnels-cs (approved):

PR #320, opened 2025-10-31
Docker compose prod update

zoecough:
- updated docker compose to work better for the swarm
- updated ctfd running check to send the "not running" message to stderr so it actually prints to the terminal again
- added pnpm-store to gitignore

### DIFF
--- a/docker-compose.prod.yaml
+++ b/docker-compose.prod.yaml
@@ -31,13 +31,13 @@ services:
       - THREADS=100
     volumes:
       - type: bind
-        source: /mnt/efs/ctf-ng/.data/CTFd/logs
+        source: ./.data/CTFd/logs
         target: /var/log/CTFd
       - type: bind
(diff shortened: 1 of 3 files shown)


docker-compose.prod.yaml
```diff
@@ -31,13 +31,13 @@ services:
...
+        source: ./.data/CTFd/logs
         target: /var/log/CTFd
       - type: bind
-        source: /mnt/efs/ctf-ng/.data/CTFd/uploads
+        source: ./.data/CTFd/uploads
         target: /var/uploads
       - type: bind
-        source: /mnt/efs/ctf-ng/backend/ng
+        source: ./backend/ng
         target: /opt/CTFd/CTFd/plugins/ng/
```
RayofLightz, line 41: Um we want non relative paths. It now matters what dir we run this from

RayofLightz (changes requested):

RayofLightz (approved):

PR #321, opened 2025-10-31
Static addresses

### LINKED ISSUE #315: Container networks should support static ip addresses
Container networks should allow the declaration of IPAM networks. For now just ipv4 is okay.

RayofLightz:
Adds static address support

### DIFF
--- a/frontend/src/routes/events/challenge/ChallengeSidebar.tsx
+++ b/frontend/src/routes/events/challenge/ChallengeSidebar.tsx
@@ -87,7 +87,7 @@ export default function ChallengeSidebar() {
               </Box>
             )}
 
-            {challenge && event && (
+            {challenge && event && granted && (
               <Flex direction="row" gap="2" mt="3" align="center">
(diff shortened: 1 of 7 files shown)


frontend/src/routes/events/challenge/ChallengeSidebar.tsx
```diff
@@ -87,7 +87,7 @@ export default function ChallengeSidebar() {
               </Box>
             )}
 
-            {challenge && event && granted && (
```
zoecough, line 90: what's the check removed for? also this is causing a linter error because granted is still being set

zoecough (changes requested):

frontend/src/routes/events/challenge/ChallengeSidebar.tsx
RayofLightz: My bad, this was before the perm check fix was pushed and I just wanted to be able to quickly do my dev work I can fix this.

zoecough (approved):

PR #324, opened 2025-10-31
Add image pulling to the admin Challenges section, fix image name in container status

trevnels-cs:
Adds a "Pull Images" confirmation modal to the blueprints section of a challenge:

image

image

Clicking pull will have the backend pull associated challenge images, and will show the pull status on each card as sent by the websocket. Pull errors will also be shown above the containers.

image
image
image

Also uses `image.attrs` to get image names for container statuses, since `image.tags` becomes empty for containers running older versions of an image that was previously tagged.

### DIFF
--- a/backend/ng/containers/models/ContainerInstance.py
+++ b/backend/ng/containers/models/ContainerInstance.py
@@ -211,10 +211,18 @@ class ContainerInstance(db.Model):
         client = get_client(config.DOCKER_HOST)
         ctr = client.containers.get(self.dockerid)
 
+        image = "unknown"
+        if ctr.image.attrs["RepoTags"]:
+            # if the image is tagged, show the tag
+            image = ctr.image.attrs["RepoTags"][0]
+        elif ctr.image.attrs["RepoDigests"]:
+            # if this image is no longer tagged, show the digest instead
+            image = ctr.image.attrs["RepoDigests"][0]
+
         data = {
             "id": self.id,
             "name": ctr.name,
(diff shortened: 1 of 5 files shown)


backend/ng/containers/models/ContainerInstance.py
```diff
@@ -211,10 +211,18 @@ def status(self) -> SerializedInstanceStats:
         client = get_client(config.DOCKER_HOST)
         ctr = client.containers.get(self.dockerid)
 
+        image = "unknown"
+        if ctr.image.attrs["RepoTags"]:
+            # if the image is tagged, show the tag
+            image = ctr.image.attrs["RepoTags"][0]
+        elif ctr.image.attrs["RepoDigests"]:
+            # if this image is no longer tagged, show the digest instead
+            image = ctr.image.attrs["RepoDigests"][0]
```
RayofLightz, line 220: Would it make sense to send both the tag and the digest as an object?

RayofLightz (approved):
Just a quick question. Otherwise this should be good to merge

backend/ng/containers/models/ContainerInstance.py
trevnels-cs: The digest is pretty unwieldy to display on the frontend, so I only really want to use it as a fallback for old images that no longer have tags that previously did. If we eventually want something like a copy button to get the image hash, it might make sense to send both though.

cjones-cs (approved):

PR #325, opened 2025-10-31
On failed auth default to no creds

RayofLightz:


backend/ng/containers/utils/Client.py
```diff
@@ -59,4 +59,9 @@ def pull_image(self, image):
                 "password": get_app_config("CONTAINER_REGISTRY_PASSWORD"),
             }
 
-        self.images.pull(image, auth_config=auth)
+        try:
+            self.images.pull(image, auth_config=auth)
```
trevnels-cs: Won't this still be sending our AWS creds to docker hub for every attempted image pull?

RayofLightz: Yeah then on the auth fail it will treat it like a public repo

RayofLightz: My thought was I can either do a bunch of string matching to try and determine if an image is public/private or I can just try using no auth as a fallback

trevnels-cs: I think string matching is the better approach here - not sure how I feel about leaking AWS creds to a registry that shouldn't have them. I think checking if `auth["registry"]` (minus the protocol) is in `image` is a bit safer and will avoid the extra failed request for each image pull

trevnels-cs (approved):
Need to make sure that `CONTAINER_REGISTRY` is not set with a protocol (i.e. `https://`) that the container image strings likely won't have and therefore won't match - looks good apart from that caveat

PR #338, opened 2025-11-04
replace imghdr with filetype library

CarterPerez-dev:
"""
imghdr — Determine the type of an image
Deprecated since version 3.11, removed in version 3.13.

This module is no longer part of the Python standard library. It was [removed in Python 3.13](https://docs.python.org/3/whatsnew/3.13.html#whatsnew313-pep594) after being deprecated in Python 3.11. The removal was decided in [PEP 594](https://peps.python.org/pep-0594/).

Possible replacements are third-party libraries from PyPI: [filetype](https://pypi.org/project/filetype/), [puremagic](https://pypi.org/project/puremagic/), or [python-magic](https://pypi.org/project/python-magic/). These are not supported or maintained by the Python core team.

The last version of Python that provided the imghdr module was [Python 3.12](https://docs.python.org/3.12/library/imghdr.html).
"""
https://docs.python.org/3/library/imghdr.html

### DIFF
--- a/backend/ng/core/utils/file_helpers.py
+++ b/backend/ng/core/utils/file_helpers.py
@@ -2,7 +2,7 @@
 File handling utility functions
 """
 
-import imghdr
+import filetype
 from werkzeug.datastructures import FileStorage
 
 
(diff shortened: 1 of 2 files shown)


droge91 (approved):
Filetype should be fine as a replacement to imghdr, LGTM.

PR #340, opened 2025-11-04
Adds some sane defaults for challenge resource use

RayofLightz:
Adds some defaults for challenge container resource limits


johnmarsden-edu (approved):
LGTM

PR #342, opened 2025-11-05
Don't allow submissions when correct attempt has been made

RayofLightz:
Fixes issue with allowing multiple correct submissions, no more cheese


johnmarsden-edu (commented):
There doesn't appear to be a test that hits the API to trip this error and that should probably be added to ensure that nothing funky happens.

RayofLightz:
`test_validate_attempt_correct_attempt_already_made` tests where the actual check is happening but I can put a test in that calls the api as well

johnmarsden-edu:
I think we should make one for the API just in case something changes at some point, as it also allows us to define exactly what it looks like to interact with API

johnmarsden-edu (approved):

Issue #349, opened 2025-11-10
Registration Fixes


cjones-cs:
Since accepting the terms is required, it should have the same error styling as the name field
<img width="762" height="761" alt="Image" src="https://github.com/user-attachments/assets/2677f445-6aef-446d-86b2-7152e7117532" />

We probably shouldn't display the exact length of an invite code. You
You should be able to input the full url or just the code and it will rip out the necessary part to send to the server
<img width="762" height="761" alt="Image" src="https://github.com/user-attachments/assets/133c081c-0e1d-413a-863f-f612f5c6057a" />

It looks like the perms for the team tab actions don't make sense.
<img width="1473" height="345" alt="Image" src="https://github.com/user-attachments/assets/f27e97e7-49f2-4062-ab83-15a3e578f19a" />
Problems here:
<img width="1473" height="345" alt="Image" src="https://github.com/user-attachments/assets/c17ace02-9c7b-4e04-89e7-5d92d4a2cf6a" />

trevnels-cs:
I believe the reason why the team tab actions look like that is that admins were changed to always have team management permissions, even if they aren't the team captain - I believe that's why the admin still has Remove Player/Assign Captain buttons, since that user is still being granted those permissions

cjones-cs:
Admin perms were the cause of the team tab looking wonky.

PR #351, opened 2025-11-12
Fix/registration modal

cjones-cs:
https://github.com/CyberSkyline/ctf-ng/issues/349


cjones-cs:
Checkbox error message is the same formfield error display. Also allows us to set the label to null to remove the tag
image

trevnels-cs (approved):

PR #354, opened 2025-11-12
User validation

### LINKED ISSUE #336: User validator is missing
The user model does not currently have a validator, so the PUT endpoint used to update users can write whatever data it wants into the ctfd_user object. The validator for this endpoint should restrict updates to ensure they are valid and only touch the expected fields.

droge91:
Adds a proper validator for the User model closes #336

### DIFF
--- a/backend/ng/user/models/User.py
+++ b/backend/ng/user/models/User.py
@@ -82,8 +82,15 @@ class User(db.Model):
 
     @classmethod
     def validate(cls, data: dict[str, Any]) -> dict[str, Any]:
-        # TODO - implement
-        return data
+        """Validate user data before creating or updating a user."""
+        validator = BaseValidator()
+        if "name" in data:
+            validator.validate_string(data, "name", 30, required=True, friendly_name="Username")
+        if "email" in data:
+            validator.validate_email(data, "email", 254, required=True, friendly_name="Email")
+        
+
+        return validator.validate()
 
     @classmethod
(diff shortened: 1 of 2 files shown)


backend/ng/user/models/User.py
```diff
@@ -82,8 +83,15 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
 
     @classmethod
     def validate(cls, data: dict[str, Any]) -> dict[str, Any]:
-        # TODO - implement
-        return data
+        """Validate user data before creating or updating a user."""
+        validator = BaseValidator()
+        if "name" in data:
+            validator.validate_string(data, "name", 30, required=True, friendly_name="Username")
```
trevnels-cs: I think 30 characters is too short to accommodate possible full names - CTFd uses a max of 128 here:
https://github.com/CTFd/CTFd/blob/master/CTFd/models/__init__.py#L399

backend/ng/user/models/User.py
```diff
@@ -82,8 +83,15 @@ def serialize(self, include_admin_fields: bool = False) -> dict[str, Any]:
...
+        if "email" in data:
+            validator.validate_string(data, "email", 254, required=True, friendly_name="Email")
```
trevnels-cs: CTFd's db model has 128 as the max user email length, we should use that in our validator instead of 254.

trevnels-cs (approved):

PR #369, opened 2025-11-13
Team Leave permissions

### LINKED ISSUE #355: Users shouldn't be able to leave teams once they've started an event
Currently, users can theoretically start an event, see the challenges inside it, leave their team, reregister, and start the clock again to effectively reset their time even after having knowledge of the event's challenges. We may need to introduce something like a CAN_LEAVE_TEAM permission that's checked before leaving a team to prevent this, and can be used to hide the appropriate UI elements as well. This permission would be revoked once a team has started or the event's locked flag is set to true

droge91:
Adds a CAN_LEAVE_TEAM Permission for better handling of when users can and cannot leave a team.

closes #355


trevnels-cs:
I think it makes sense to grant this permission to admins along with the others here  - https://github.com/CyberSkyline/ctf-ng/blob/ctfd-teamleave-permissions/backend/ng/permissions/controllers/initial_admin_setup.py#L46

droge91:
> I think it makes sense to grant this permission to admins along with the others here - https://github.com/CyberSkyline/ctf-ng/blob/ctfd-teamleave-permissions/backend/ng/permissions/controllers/initial_admin_setup.py#L46

What would the purpose be for that? Are there likely to be cases where admins will be leaving registered teams? They already have permission to remove people from teams whenever.

trevnels-cs:
This is useful for testing the registration flow repeatedly as an admin. It is possible to perform that removal via the admin panel as well, but I think it makes sense to give admins the permission as well for consistency with how other permissions are handled (i.e. admins already retain the CAN_EDIT_TEAM permission once they've started, which normal users do not)

droge91:
> This is useful for testing the registration flow repeatedly as an admin. It is possible to perform that removal via the admin panel as well, but I think it makes sense to give admins the permission as well for consistency with how other permissions are handled (i.e. admins already retain the CAN_EDIT_TEAM permission once they've started, which normal users do not)

Okay makes sense, added.

trevnels-cs (approved):

PR #370, opened 2025-11-14
Created tooling framework + email send script

fpayer:


tools/operations/config.js
```diff
@@ -0,0 +1,21 @@
+import dotenv from 'dotenv';
+import path from 'path';
+
+const __dirname = import.meta.dirname;
+const rootPath = path.join(__dirname, '../..');
+
+dotenv.config({ path : path.join(rootPath, '.env'), quiet : true });
+
+export const OUTPUT_DIR = path.join(__dirname, './out');
```
zoecough, line 9: is there a reason this is set as a config variable instead of a flag on the sendWithList action?

zoecough (approved):

tools/operations/config.js
fpayer: This is the default output dir for all scripts (See https://github.com/CyberSkyline/ctf-ng/pull/370/files/62773c6d22f57248cde973aeab2b740f599f3715#diff-d9f37945517e029830227e9a3b90ca57ca8e43685666e31cb95653d977ce5b6bR35)

It will get overridden if you specify a different value when running the script.

PR #371, opened 2025-11-14
Change NoVNC port

johnmarsden-edu:


fpayer:
@RayofLightz and I were actually talking about this config. It should be moved outside of `config.py` and into an ENV var instead. Idk what Tristan was thinking about timeline though.

fpayer:
This setting has been moved to .env now so this PR isn't needed anymore.

PR #378, opened 2025-11-14
S3 file handling

zoecough:
- endpoints to upload/download files using S3 pre-signed URL
- shared S3 handling service between public (non-support) and private (support ticket) files
- Updated support ticket file endpoints and removed support ticket file search
- Removed minio


cjones-cs (approved):
lgtm

backend/ng/__init__.py
```diff
@@ -70,13 +70,39 @@ def load(app: Any) -> None:
...
+        s3_service = get_s3_service()
+        s3_service.init_app(app)
+
+        if s3_service.is_configured():
+            logger.info("S3 service initialized successfully")
+        else:
+            logger.warning("S3 service not configured - file operations disabled")
+
+        # Load fileuploads module
+        load_fileuploads(app)
```
RayofLightz, line 96: get_app_config might be a more in platform approach, but its kinda also a pain to get working 100 percent

backend/ng/core/services/s3_service.py
```diff
@@ -0,0 +1,239 @@
...
+            )
+
+            content_type = response.get('ContentType', 'application/octet-stream')
+            content_length = response.get('ContentLength', 0)
+
+            return response['Body'], content_length, content_type
+
+        except ClientError as e:
+            logger.error(f"Error downloading file from S3: {e}")
+            raise
```
RayofLightz: When are we downloading the file stream?

backend/ng/fileuploads/controllers/search.py
```diff
@@ -0,0 +1,106 @@
...
+            return jsonify({
+                "success": True,
+                "data": {
+                    "files": files,
+                    "total_count": len(files),
+                    "query": filename_pattern,
+                    "folder": folder,
+                    "search_type": "folder_specific_search"
+                }
+            })
```
RayofLightz: Use the success_response util function

backend/ng/support/services/s3_upload_service.py
```diff
@@ -1,222 +1,115 @@
...
+    def download_ticket_attachment(self, s3_key: str):
+        """
+        Stream ticket attachment for proxy download
+        Returns (stream, content_length, content_type)
+        """
+        s3_service = self._get_s3_service()
+        if not s3_service or not s3_service.is_configured():
+            raise Exception("S3 service not configured")
+
+        return s3_service.download_file_stream(s3_key)
```
RayofLightz: Dont we want to redirect to the s3 presigned url?

RayofLightz (changes requested):

backend/ng/__init__.py
zoecough: get_app_config pulls from the flask app, this is putting the vars from the env (the os environ in the docker container) into the flask app

backend/ng/core/services/s3_service.py
zoecough: using this for the ticket attachment download, it's wrapped by the upload service def in support and i'm using it to stream content from the TicketAttachment

backend/ng/support/services/s3_upload_service.py
```diff
@@ -77,16 +77,26 @@ def upload_ticket_attachment(self, file: FileStorage, ticket_id: int,
...
-            raise Exception("S3 service not configured")
+            current_app.logger.error("S3 service not configured")
+            return None
 
-        return s3_service.download_file_stream(s3_key)
+        try:
+            # Generate presigned URL with 1 hour expiration
+            presigned_url = s3_service.generate_download_url(
+                s3_key,
+                expires_in=3600  # 1 hour
```
RayofLightz, line 143: Just want to know why the ticket one has a 1 hour expiration?

RayofLightz (approved):

backend/ng/support/services/s3_upload_service.py
zoecough: i wanted to set a short expiry time on everything because the presigned URLs are just supposed to be used once each for whatever operation the user is doing

PR #385, opened 2025-11-17
Event header rework and permission fixes

trevnels-cs:
Reworks team leave/rename controls and places them in the event header, as well as puts the start button back down onto the challenge tab (but with better handling of admin permission special case.) Currently, the Leave Team button is available in all cases, it will be tied to the CAN_LEAVE_TEAM permission once that is merged, similar to how the rename button is currently tied to CAN_EDIT_TEAM.

image

image

image

Since admins have special permission to see challenges before starting, their UI in that state is a little different - they still get the start button, but can also view challenges ahead of time.
image

Since some team controls were moved into the header, the team page has been cleaned up to avoid redundancy. The team name has also been emphasized on that page. Since non-captain actions are all in the page header now, we could potentially explore hiding the actions column completely for non-captain users, since it would otherwise just be whitespace.
image


frontend/src/hooks/events.ts
```diff
@@ -69,23 +69,25 @@ export function useAdminEvent(eventId: number | null) {
 export function registerMyEvent(eventId: number, teamName: string) {
   return apiMutation(`/events/${eventId}/me/register`, { team_name : teamName }, {
     method : 'POST',
-  }).then(() => {
-    mutate('/users/me/events');
-    mutate('/users/me/teams');
-    mutate(`/events/${eventId}/me/team`);
+  }).then(() => Promise.all([
```
cjones-cs, line 72: Why did you need to wrap the mutates in a promise.all?

cjones-cs (approved):

frontend/src/hooks/events.ts
trevnels-cs: I split this `then` chain into two pieces since there was a race condition between eligibility and permissions which led to some error flicker when registering/unregistering. The fix here is to ensure that all of the other endpoints update (via the Promise.all) before rechecking eligibility so that the UI has a chance to update before the split-second error is shown.

PR #386, opened 2025-11-17
Fix CAN_START_TEAM_TIMER Logic

### LINKED ISSUE #383: CAN_START_TEAM_TIMER permission is granted before event start time
Team captains should not have CAN_START_TEAM_TIMER until the event's start timestamp has passed, but it is currently being granted beforehand.

I am able to start the team in this state, but `CAN_VIEW_CHALLENGES` is not granted. This should still not be possible though, since timers shouldn't be able to start until the event does.

droge91:
Fixes a permissions bug where teams could start before an event has started fixes #383


trevnels-cs (approved):
Good to merge once tests pass.

PR #390, opened 2025-11-18
Added markdown render to support tickets

### LINKED ISSUE #388: Support tickets need to be rendered using markdown

fpayer:
Fixes #388 

Also added markdown sanitization + rehype raw to allow images

### DIFF
--- a/frontend/src/components/RadixMarkdown.tsx
+++ b/frontend/src/components/RadixMarkdown.tsx
@@ -11,7 +11,6 @@ import {
 import ReactMarkdown, { type Components } from 'react-markdown';
 import rehypeRaw from 'rehype-raw';
 import rehypeSanitize, { defaultSchema } from 'rehype-sanitize';
-import remarkBreaks from 'remark-breaks';
 import remarkGfm from 'remark-gfm';
 
 const DISALLOWED_TAGS = [ 'input', 'picture', 'source', 'kbd', 'script' ];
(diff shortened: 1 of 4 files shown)


frontend/src/components/RadixMarkdown.tsx
```diff
@@ -46,7 +67,7 @@ export default function RadixMarkdown({ children }: {
   children: string;
 }) {
   return (
-    <ReactMarkdown components={Components} remarkPlugins={[ remarkGfm ]}>
+    <ReactMarkdown components={Components} remarkPlugins={[ remarkBreaks, remarkGfm ]} rehypePlugins={[ rehypeRaw, [ rehypeSanitize, SANITIZE_SCHEMA ] ]}>
```
trevnels-cs: remarkBreaks was removed from here intentionally since it results in nonstandard markdown behavior that broke a few things (i.e. the terms of use shown during event registration, challenge descriptions) since it treats a single \n as a \<br>, when standard markdown expects \n\n to produce a line break in the output and a single \n should render as a space instead.

fpayer: My rationale for adding it in was to have the resulting render match what the user sees in their preview more accurately. 

Instead of using remarkBreaks, what do you think about manipulating the resulting markdown from the markdown editor to insert the <br> instead?

trevnels-cs: image

It looks like milkdown already does this by inserting double newlines for each single line, which will get rendered to a single `<br>` without remarkBreaks

fpayer: ok, sounds good 👍. I'll just remove remakrBreaks

fpayer: Good to go otherwise?

trevnels-cs (approved):

PR #391, opened 2025-11-18
Indvidual container pull restart recycle

RayofLightz:
* Hooks up individual containers restarts + recycles to the frontend
* Moves vnc config to .env file
* Removes nginx caching for vnc auth


conf/nginx/http.conf
```diff
@@ -130,8 +130,8 @@ http {
       proxy_set_header X-Real-IP $remote_addr;
       proxy_set_header Host $host;
       proxy_http_version 1.1;
-      proxy_cache port_forward_auth_cache;
-      proxy_cache_key "$remote_addr";
+      #proxy_cache port_forward_auth_cache;
```
trevnels-cs, line 133: Are there any performance implications in removing this caching? Will the auth request only run once on Websocket init, or for every socket message?

trevnels-cs (approved):

conf/nginx/http.conf
trevnels-cs: Seems like it only does the lookup on initial handshake, so it should be fine.

PR #394, opened 2025-11-19
Working exec service, expose container environment, update deployment sidebar

### LINKED ISSUE #377: Getting to the deployment information is clunky
Because of the links in this column it makes it difficult to just get to the deployment information. I currently have to click on either the small check box or number field to get deployment information up. 

I foresee this UX being an issue when some of the ota devs start using the platform to test challenges.

trevnels-cs:
- Connect xtermjs to exec service - challenge container terminals can now be accessed through the admin UI.
  - Update exec service nginx config to use a URL param instead of passing headers with the socket
  - Update exec service code to clean up streams and handle errors more gracefully to prevent crashes
- Pass container environments to frontend as part of the status endpoint. Easy way to see what dynamic variables are being passed into a container.
  - We may still want a cleaner view of this that's independent of the containers. i.e. to see rendered answers.
- Update deployment sidebar
  - Use cards instead of table for better use of space
  - Add link to associated challenge
  - Removed placeholder data until we decide how/if it should be displayed
  - Removed links from deployment table to prevent misclicks (they're available on the sidebar still.) This resolves #377.

image
image

Known issues:
- bash processes are not stopped when disconnecting the socket. After enough execs, they can eventually accumulate and break the container by hitting limits, requiring a container restart to fix.
- exec sockets time out after about 1 minute of inactivity. I tried adjusting nginx timeout config to resolve this, but it didn't seem to have an effect.


frontend/src/routes/admin/deployments/ServiceCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+          <Box className="text-right">
+            <Skeleton loading={!status}>
+              <Badge color={status?.status === 'running' ? COLOR_POSITIVE : COLOR_NEGATIVE}>
+                {upperCase(status?.status || 'unknown')}
+              </Badge>
+            </Skeleton>
+            <br />
+            <Tooltip content={<Text>{service?.dockerid}</Text>}>
+              <Text color="gray">{service?.dockerid.slice(0, 12)}</Text>
+            </Tooltip>
```
cjones-cs, line 48: Did you test what would happen if there's no service? Seems like it could still produce an empty tooltip, which wouldn't be very helpful.

cjones-cs: Looks like service is guaranteed to be present since it comes in as a prop, but that would be an ! instead of ? right?

frontend/src/routes/admin/deployments/ServiceCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { upperCase } from 'lodash';
+import ContainerExecModal from './ContainerExecModal';
+import ContainerLogsModal from './ContainerLogsModal';
+import RecycleContainerModal from './RecycleContainerModal';
+import RestartContainerModal from './RestartContainerModal';
+
+export default function ServiceCard({ service }: {service: ContainerInstance}) {
+  const { data : status, error } = useContainerStatus(service.id);
+  return (
+    <Card key={service.id}>
```
cjones-cs, line 31: Should probably destructure service since it's used a bunch of times

cjones-cs (changes requested):
Small comment. Also, please handle the linting warning in the server.ts, if possible

frontend/src/routes/admin/deployments/ServiceCard.tsx
trevnels-cs: The ? isn't needed anymore here since ContainerInstance is always present and non-nullable, just removed them

frontend/src/routes/admin/deployments/ServiceCard.tsx
trevnels-cs: Added destructuring for service and status

cjones-cs (approved):

cjones-cs:
approved pending Tristan's review

services/exec-service/.eslintrc.cjs
```diff
@@ -36,6 +36,7 @@ module.exports = {
       'max-len' : 160, // Maximum line length for imports. Must be kept in sync with max-len rule above.
       semi : true, // Imports should end with a semicolon
     } ],
+    'no-console': 'off', // Services can and should perform logging via console
```
RayofLightz, line 39: Consider using disable line in the server.ts

RayofLightz (approved):
This is good for the initial setup I know down the line I will be adding the ability to use a non bash shell for the exec service as well as a pre shared key between the app server and the exec service

services/exec-service/.eslintrc.cjs
trevnels-cs: I think it's cleaner to make this change in the lint config for the service, since we know we want to allow logging there. The `no-console` rule only really makes sense in a web context, which this is not.

trevnels-cs:
@RayofLightz thoughts on defaulting to `sh` since it should always be present? Might be simpler to do that and let the user invoke the shell of their choosing after connecting instead of having it be configurable.

With it being configurable, the user would still have to guess whether bash (or another shell) is present or not before connecting, whereas `sh` will get them connected and they can then see if the other shell of their choosing is available afterwards.

There are already cases where the exec service doesn't work:
image

RayofLightz:
I like that idea. Just makes everything simpler. I will add a commit to this branch changing it to bin sh

PR #404, opened 2025-11-20
Display variables on challenge and deployment admin panels

trevnels-cs:
Challenges show variables with their default/template values. Variable names are used in the question table instead of templates to save space. Also shows callouts instead of empty tables when nothing is present.
Screenshot From 2025-11-20 12-18-09
The Question serializer has been updated to wrap literal string answers with quotes to distinguish them from variables. We can reliably distinguish them this way, since a variable won't have quotes in its name.

Deployments show variables rendered for the corresponding team:
Screenshot From 2025-11-20 12-16-05

### DIFF
--- a/backend/ng/challenge/models/Question.py
+++ b/backend/ng/challenge/models/Question.py
@@ -65,7 +65,7 @@ class Question(db.Model):
             "challenge_id": self.challenge_id,
         }
         if include_admin_fields:
-            data["answer"] = self.answer_variable.name if self.answer_variable else "\"" + self.answer + "\""
+            data["answer"] = self.answer_variable.name if self.answer_variable else f'"{self.answer}"'
 
         return data
 
--- a/backend/ng/challenge/models/Question.py
+++ b/backend/ng/challenge/models/Question.py
@@ -65,7 +65,7 @@ class Question(db.Model):
             "challenge_id": self.challenge_id,
         }
         if include_admin_fields:
-            data["answer"] = self.answer_variable.template if self.answer_variable else self.answer
+            data["answer"] = self.answer_variable.name if self.answer_variable else "\"" + self.answer + "\""
 
         return data
 
--- a/backend/ng/containers/routes/admin_routes.py
+++ b/backend/ng/containers/routes/admin_routes.py
@@ -1,5 +1,7 @@
 from flask_restx import Namespace, Resource
 from flask import request
+from ...challenge.utils import generate_seed
+from ...challenge.models import Challenge
 from ..controllers.get_stats import get_stats
 from ..controllers.admin_exec import admin_exec
 from ..models.ContainerInstance import ContainerInstance
@@ -10,7 +12,9 @@ from ...core.middleware import (
 
 from ...core.middleware.loaders import (
     LoaderType,
+    load_team,
     load_container_instance,
+    load_challenge
 )
 
 from ...core.utils import (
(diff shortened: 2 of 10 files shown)


backend/ng/challenge/models/Question.py
```diff
@@ -65,7 +65,7 @@ def serialize(self, include_admin_fields=False) -> SerializedQuestion:
             "challenge_id": self.challenge_id,
         }
         if include_admin_fields:
-            data["answer"] = self.answer_variable.template if self.answer_variable else self.answer
+            data["answer"] = self.answer_variable.name if self.answer_variable else "\"" + self.answer + "\""
```
cjones-cs: Curious, but why is it "\ "" with two ending quotes instead of one?

frontend/src/routes/admin/deployments/DeploymentSidebar.tsx
```diff
@@ -42,12 +44,36 @@ export default function DeploymentSidebar({ entity }: {entity: Deployment}) {
...
+      && (
+        <Table.Root>
+          <Table.Header>
+            <Table.ColumnHeaderCell>Name</Table.ColumnHeaderCell>
+            <Table.ColumnHeaderCell>Value</Table.ColumnHeaderCell>
+          </Table.Header>
+          <Table.Body>
+            {Object.entries(variables || {}).map(([ key, value ]) => (
+              <Table.Row key={key}>
+                <Table.Cell>
```
cjones-cs, line 67: Should we be naming the cells so screen readers can have context on what the cell contains? Recently saw an accessibility video that talked about that

cjones-cs (approved):

backend/ng/challenge/models/Question.py
trevnels-cs: The latter of the two quotes closes the string

frontend/src/routes/admin/deployments/DeploymentSidebar.tsx
trevnels-cs: Since these tables are using the right semantic table elements, screen readers should be able to pick up on the structure of the table as-is and understand which header corresponds to each cell. This matches the WAI-ARIA docs, though if there are extra properties that should be there we can add them.

image

backend/ng/challenge/models/Question.py
johnmarsden-edu:
```suggestion
            data["answer"] = self.answer_variable.name if self.answer_variable else f'"{self.answer}"'
```

backend/ng/containers/routes/admin_routes.py
```diff
@@ -66,7 +70,28 @@ def get(self, challenge_id, team_id, **kwargs):
...
+        },
+    )
+    @load_team(LoaderType.PARAM)
+    @load_challenge(LoaderType.PARAM)
+    @admin_endpoint()
+    def get(self, challenge: Challenge, team, **kwargs):
+      # create answer variable to question map to ensure seeding is consistent with questions/env
+      qids = {q.answer_variable_id: q.id  for q in challenge.questions}
 
+      values = {v.name: v.as_attr().template.eval(generate_seed(challenge.event_id, challenge.id, qids.get(v.id), team_seed=team.seed)) for v in challenge.variables}
```
johnmarsden-edu, line 93: You should be able to use ContainerBlueprint or Instance (don't remember which off the top of my head) and call the render environment function to get these. I can give you the exact function later if you can't find it.

PR #419, opened 2025-11-21
Improve install script

### LINKED ISSUE #363: Install script does not handle already being in a docker swarm gracefully
The install script does not handle already being in a docker swarm gracefully and you get this error, stopping the install script:

```
Error response from daemon: This node is already part of a swarm. Use "docker swarm leave" to leave this swarm and join another one.
```

johnmarsden-edu:
Try to gracefully handle docker swarm not being initialized and some minor ease of use and grammar edits in the tls certs section

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -207,7 +207,7 @@ if ! compgen -G "$HOME/.docker/*.pem" > /dev/null; then
     cp cert.pem ~/.docker/
     cp key.pem ~/.docker/
     sudo mkdir -p /var/lib/certs/ssl/
-    sudo cp ~/.docker/* /var/lib/certs/ssl/
+    sudo cp ~/.docker/*.json ~/.docker/*.pem /var/lib/certs/ssl/
     cd ../
     sudo mkdir -p /etc/docker/ssl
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -200,7 +207,7 @@ if ! compgen -G "$HOME/.docker/*.pem" > /dev/null; then
     cp cert.pem ~/.docker/
     cp key.pem ~/.docker/
     sudo mkdir -p /var/lib/certs/ssl/
-    sudo cp ~/.docker/* /var/lib/certs/ssl/
+    sudo cp ~/.docker/*.json ~/.docker/*.pem /var/lib/certs/ssl/
```
RayofLightz, line 211: Any particular reason for this change?

RayofLightz (approved):

install.sh
johnmarsden-edu:
```suggestion
    # Turns out that the buildx folder also gets put in that folder and so when you try to copy everything it fails because there's a directory in there that it can't copy without -r
    sudo cp ~/.docker/*.json ~/.docker/*.pem /var/lib/certs/ssl/
```

Issue #420, opened 2025-11-21
Update roles endpoint does not work


trevnels-cs:
<img width="848" height="132" alt="Image" src="https://github.com/user-attachments/assets/a3187603-7f9c-4729-bfdd-e30b5b7889ee" />

trevnels-cs:
Resolved by #422

PR #422, opened 2025-11-23
issue 420

CarterPerez-dev:
Fix AttributeError 
closes - https://github.com/CyberSkyline/ctf-ng/issues/420

### DIFF
--- a/backend/ng/permissions/models/UserRole.py
+++ b/backend/ng/permissions/models/UserRole.py
@@ -84,9 +84,8 @@ class UserRole(db.Model):
         Returns:
             list: List of UserRole instances assigned to the user.
         """
-
-        role_names = data.get("roles", [])
-        if not role_names:
+        role_names = data.get("roles")
+        if role_names is None:
             raise ValidationError("No role names provided")
 
         # Clear existing roles
@@ -128,7 +127,7 @@ class UserRole(db.Model):
     def validate_user_role_update(cls, data):
         """Validate user role updates."""
         validator = BaseValidator()
-        if not data.get("roles"):
+        if data.get("roles") is None:
             validator.errors["roles"] = "No role names provided"
             raise ValidationError("No role names provided")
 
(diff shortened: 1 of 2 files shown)


backend/ng/permissions/models/UserRole.py
```diff
@@ -64,8 +64,12 @@ def assign_role_to_user_by_id(cls, user_id: int, role_id: int):
         return cls.create_user_role(user_id, role_id)
 
     @classmethod
-    def assign_role_to_user_by_name(cls, user_id: int, role_name: RoleEnum):
-        role = Role.query.filter_by(name=role_name.value).first()
+    def assign_role_to_user_by_name(cls, user_id: int, role_name):
```
trevnels-cs: Make sure the type signature for `role_name` is updated to reflect the updates made to the function.

trevnels-cs:
image
Also the underlying role query does not work. It looks like `name` is `a` and not `admin` when it comes into this function.

trevnels-cs:
It looks like the issue here is specifically related to setting a single role. When setting ["admin", "support"], the endpoint works. When passing multiple items, role_names in update_user_roles is `[<RoleEnum.ADMIN: 'admin'>, <RoleEnum.SUPPORT: 'support'>]`. When passing a single item, it's just `RoleEnum.ADMIN` (no list), so the for loop iterates by character over the string.

The issue is that validate_enum sets the value directly at first, then only converts it to a list when multiple values are passed.

trevnels-cs:
Another thing - setting roles to `[]` (to remove all roles) results in this error:
```
{
    "success": false,
    "errors": {
        "validation": "No role names provided: {}"
    }
}
```

The validator needs to accept an empty list as well

fpayer:
Need to get this merged in soon because the behavior in the `development` branch when trying to edit a user's name now causes all of their roles to be removed. So trying to edit the name for an admin user causes them to lose their roles with no way to put them back.

fpayer (approved):

PR #423, opened 2025-11-23
Leaderboard changes

droge91:
Adds tiebreak logic for leaderboard
Adds an admin leaderboard toggle for events
Removes unranked or non started teams from the leaderboard.

### DIFF
--- a/backend/ng/event/routes/admin_routes.py
+++ b/backend/ng/event/routes/admin_routes.py
@@ -232,3 +232,23 @@ class EventChallenges(Resource):
         """
         challenges = event.get_all_challenges()
         return success_response(challenges)
+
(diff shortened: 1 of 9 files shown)


backend/ng/event/routes/admin_routes.py
```diff
@@ -232,3 +232,23 @@ def get(self, event: Event, **kwargs):
...
+            500: "Internal Server Error",
+        },
+    )
+    @admin_endpoint()
+    @load_event(source=LoaderType.PARAM)
+    def post(self, event: Event, **kwargs):
+        """
+        Toggle event leaderboard visibility
+        """
+        event.show_leaderboard = not event.show_leaderboard
```
fpayer: Doesn't this change need to be committed to the db?

fpayer (approved):

Issue #285, opened 2025-10-17
Create admin UI to create/edit sponsors


fpayer:

cjones-cs:
Linked issue to display user information on profile page: https://github.com/CyberSkyline/ctf-ng/issues/207

Issue #284, opened 2025-10-17
Update user profile page to allow the user to configure their sponsor


fpayer:

cjones-cs:
Other user profile issue: https://github.com/CyberSkyline/ctf-ng/issues/207

Issue #207, opened 2025-09-26
Frontend profile page


trevnels-cs:
Need to determine which fields, if any, we should allow the user to edit. Many may be determined by SSO.

Area for badges/achievements?

PR #428, opened 2025-11-24
Sponsors and Profile page

cjones-cs:
https://github.com/CyberSkyline/ctf-ng/issues/285
https://github.com/CyberSkyline/ctf-ng/issues/284
https://github.com/CyberSkyline/ctf-ng/issues/207

Currently, user profile will only allow for editing/display of sponsor data.

### DIFF
--- a/backend/ng/fileuploads/controllers/public_files.py
+++ b/backend/ng/fileuploads/controllers/public_files.py
@@ -4,6 +4,7 @@ Public file operations for sponsor-logos, event-cards, favicons
 import uuid
 import logging
 import requests
+from ...core.utils import success_response, error_response
 from flask import request
 from werkzeug.datastructures import FileStorage
 
@@ -59,35 +60,35 @@ def get_public_file(args):
         filename = args.get('filename')
 
         if not folder or not filename:
-            return {"error": "Folder and filename are required"}, 400
+            return error_response("Folder and filename are required", 400)
 
         if folder not in ALLOWED_FOLDERS:
-            return {"error": "Invalid folder"}, 404
+            return error_response("Invalid folder", 404)
 
         from ...core.services.s3_service import get_s3_service
         s3_service = get_s3_service()
         if not s3_service or not s3_service.is_configured():
-            return {"error": "File storage not configured"}, 503
+            return error_response("File storage not configured", 503)
 
         object_key = f"{folder}/{filename}"
 
         if not s3_service.object_exists(object_key):
-            return {"error": "File not found"}, 404
+            return error_response("File not found", 404)
 
         presigned_url = s3_service.generate_download_url(object_key, expires_in=86400)
 
         logger.info(f"Generated download URL for {object_key}")
 
-        return {
+        return success_response({
             "url": presigned_url,
             "filename": filename,
             "folder": folder,
             "object_key": object_key
-        }
+        })
 
     except Exception as e:
         logger.error(f"Get public file error: {e}")
-        return {"error": "File not found"}, 404
+        return error_response("File not found", 404)
 
 def list_public_files(args):
     """List all files in a specified public folder"""
@@ -95,12 +96,12 @@ def list_public_files(args):
         folder = args.get('folder')
 
         if not folder or folder not in ALLOWED_FOLDERS:
-            return {"error": "Valid folder parameter required"}, 400
+            return error_response("Valid folder parameter required", 400)
 
         from ...core.services.s3_service import get_s3_service
         s3_service = get_s3_service()
         if not s3_service or not s3_service.is_configured():
-            return {"error": "File storage not configured"}, 503
+            return error_response("File storage not configured", 503)
 
         objects = s3_service.list_objects(prefix=folder)
         files = []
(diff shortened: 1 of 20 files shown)


backend/ng/fileuploads/controllers/public_files.py
```diff
@@ -59,48 +60,48 @@ def get_public_file(args):
...
         presigned_url = s3_service.generate_download_url(object_key, expires_in=86400)
 
         logger.info(f"Generated download URL for {object_key}")
 
-        return {
+        return success_response({
             "url": presigned_url,
             "filename": filename,
             "folder": folder,
             "object_key": object_key
```
cjones-cs, line 86: Zoe is going to standardize the output to match what the frontend UploadFile typing is. This will cause only a few minor changes to the frontend to eliminate some extra nesting

frontend/src/routes/admin/sponsors/SponsorDataForm.tsx
```diff
@@ -0,0 +1,74 @@
...
+    register, control, formState : { errors }, watch, setValue,
+  } = rhf;
+  const selectedImage = watch('logo');
+
+  const { data : imagePreview, error : imageError } = useFileUrl('sponsor-logos', selectedImage);
+
+  const [ selectedOption, setSelectedOption ] = useState<'existing' | 'upload'>('existing');
+
+  const { data, error : listError } = useFileList('sponsor-logos');
+  const options = chain(data?.files)
```
cjones-cs, line 27: This is an instance where the extra nesting will be eliminated

frontend/src/types.ts
```diff
@@ -14,12 +14,26 @@ export interface Event {
...
+  url?: string,
+}
+
 export interface User {
   id: number;
   name: string;
   email: string;
   roles: string[];
   registered_at: Date;
+  affiliation: Sponsor | null;
```
cjones-cs, line 36: This hasn't changed in the backend yet, but someone was going to update this to be "sponsor" instead of "affiliation" for consistency

cjones-cs:
File upload backend standardization tracking here: https://github.com/CyberSkyline/ctf-ng/issues/430

backend/ng/fileuploads/controllers/public_files.py
```diff
@@ -59,48 +60,48 @@ def get_public_file(args):
...
-            return {"error": "File storage not configured"}, 503
+            return error_response("File storage not configured", 503)
 
         object_key = f"{folder}/{filename}"
 
         if not s3_service.object_exists(object_key):
-            return {"error": "File not found"}, 404
+            return error_response("File not found", 404)
 
         presigned_url = s3_service.generate_download_url(object_key, expires_in=86400)
```
trevnels-cs, line 78: As far as I understand it, this code currently goes through s3 to create a new presigned url for every request made for a file. Since things like game cards will be requested constantly by many users, I think it makes sense to cache the generated presigned urls inside the s3_service until their expiration time to minimize the number of requests made to s3. @zoecough 

This doesn't need to be part of this PR, but something that I think should get done soon.

backend/ng/sponsors/routes/admin_routes.py
```diff
@@ -22,7 +22,7 @@ class SponsorsAdmin(Resource):
     def get(self, **kwargs):
         """Get all sponsors (admin only)"""
         sponsors = Sponsor.query.all()
-        return {"sponsors": [s.serialize() for s in sponsors]}
+        return success_response(sponsors)
```
trevnels-cs, line 25: Is a separate get for admins needed since it returns the same data as the normal `/sponsors`?

trevnels-cs: Same goes for getting by id

frontend/src/routes/profile/index.tsx
```diff
@@ -1,3 +1,89 @@
...
+              {map(allSponsors, (sponsor) => (
+                <SponsorImageCard
+                  sponsor={sponsor}
+                  selectSponsor={selectSponsor}
+                />
+              ))}
+            </Grid>
+          </>
+        ) : (
+          !isNil(mySponsor) && (
```
trevnels-cs, line 75: Consider adding a message in the case that mySponsor is null instructing users to click edit to select one.

cjones-cs: I'll follow up on that in another pr since this one is already huge. I also need to modify the event reg modal to require a sponsor before registration can be completed.

backend/ng/sponsors/routes/admin_routes.py
cjones-cs: That's a good question. I don't think we would need separate ones, but I defer to @droge91 since he deved that part.

frontend/src/routes/profile/index.tsx
cjones-cs: Issue: https://github.com/CyberSkyline/ctf-ng/issues/444

trevnels-cs (approved):

backend/ng/sponsors/routes/admin_routes.py
droge91: I don't think they are necessary anymore no, I believe there was originally going to be admin only fields for sponsors but that got changed so those routes are just vestigial now.

Issue #373, opened 2025-11-14
Visual Inconsistency In Support Ticket Table


johnmarsden-edu:
<img width="1227" height="385" alt="Image" src="https://github.com/user-attachments/assets/c7866c54-da81-4300-b8fa-fbda37ec5808" />

There's a weird gap at the end of the table.

johnmarsden-edu:
Also looking closer at it I'm not sure that the color's chosen for the status make sense. I feel like close should be less visually engaging than open. But that might just be a personal preference.

trevnels-cs:
The gap is due to ag-grid having a min-height. The table's normal behavior is to not get smaller in height than that

johnmarsden-edu:
When I said end of the table I meant the right end. It seems like it should spread out to take up the available empty space horizontally.

PR #440, opened 2025-11-25
Remove extra space at end of tables

cjones-cs:
https://github.com/CyberSkyline/ctf-ng/issues/373

image

### DIFF
--- a/frontend/src/components/AdminGrid.tsx
+++ b/frontend/src/components/AdminGrid.tsx
@@ -159,7 +159,6 @@ export default function AdminGrid<T>({
           }}
           initialState={initialState}
           className="w-full h-full grow"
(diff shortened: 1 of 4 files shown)


frontend/src/components/AdminGrid.tsx
```diff
@@ -159,6 +159,7 @@ export default function AdminGrid<T>({
           }}
           initialState={initialState}
           className="w-full h-full grow"
+          defaultColDef={{ flex : 1 }}
```
trevnels-cs: Admin grid columns should not be flex - they are intended to scroll horizontally instead of streching to fit the viewport

cjones-cs: removed

trevnels-cs (approved):

PR #449, opened 2025-11-25
Landing page and navbar polish/accessibility

### LINKED ISSUE #439: Unauthed dashboard and events pages
UI routes:

`/`

- links to login
- links to /events
- welcome text

`/events`

- show all the events

trevnels-cs:
image

Landing page on `/` for unauthed users. Resolves #439.

Added missing navbar permission checks (support and profile are not shown when unauthed.)

Added accessibility attributes for notification icon (previously wasn't read as an option by screen readers.) Will now be read as "Notifications" or ("Notifications Unread" if unread indicator is present) in the nav list.

### DIFF
--- a/frontend/src/components/ThemeToggle.tsx
+++ b/frontend/src/components/ThemeToggle.tsx
@@ -1,6 +1,6 @@
+import { Flex, Switch, Text } from '@radix-ui/themes';
 import { useTheme } from 'next-themes';
 import { useEffect, useState } from 'react';
-import { Flex, Switch, Text } from '@radix-ui/themes';
 
 export default function ThemeToggle({ className }: {className : string}) {
   const { theme, setTheme } = useTheme();
@@ -15,6 +15,7 @@ export default function ThemeToggle({ className }: {className : string}) {
       <Switch
         checked={theme === 'dark'}
         onCheckedChange={(checked) => setTheme(checked ? 'dark' : 'light')}
+        aria-label="Dark Theme"
       />
       <Text
         className="dark:text-(--gray-a11)"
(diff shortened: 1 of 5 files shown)


frontend/src/routes/dashboard/Landing.tsx
```diff
@@ -0,0 +1,49 @@
...
+  Link as RadixLink,
+  Text,
+} from '@radix-ui/themes';
+import { TbArrowRight, TbUserCircle } from 'react-icons/tb';
+import { Link } from 'react-router';
+
+export default function Landing() {
+  return (
+    <Flex className="absolute inset-0 overflow-hidden bg-dots-1" align="center" justify="center" direction="column" gap="3" p="3">
+      <title>Login</title>
```
cjones-cs: Title should be Home since login refers to the login form

frontend/src/components/ThemeToggle.tsx
```diff
@@ -15,6 +15,7 @@ export default function ThemeToggle({ className }: {className : string}) {
       <Switch
         checked={theme === 'dark'}
         onCheckedChange={(checked) => setTheme(checked ? 'dark' : 'light')}
+        aria-label="Dark Theme"
```
cjones-cs, line 18: Perhaps "Dark Mode On" / "Light Mode On" would be a better aria label here since this isn't just a dark mode and just labeling it like that doesn't really tell the user what mode it's in.

cjones-cs (changes requested):
changes

frontend/src/components/ThemeToggle.tsx
trevnels-cs: Since this is a switch, the screen reader reads it as "Dark Theme On"/"Dark Theme Off" - the state of the switch is appended to the label.

cjones-cs (approved):

PR #453, opened 2025-11-25
Expose leaderboard controls to the frontend

### LINKED ISSUE #356: Leaderboard Visibility
Tracking for leaderboard visibility related issues:

- Admins should be able to hide the leaderboard for an event
- Leaderboard probably shouldn't be visible before an event starts
- Determine how to handle teams that haven't started yet (I think they should be hidden from the board)

trevnels-cs:
Added missing validator/serialize fields for show_leaderboard, and expose it on event form. (Also brought back missing registration open switch.)
image
The switches look a bit janky now - updating this is on my list for later.

Leaderboard page catches the hidden leaderboard error, hides the grid, and displays this warning:
image

### DIFF
--- a/frontend/src/routes/events/OverviewTabs/LeaderboardTab/index.tsx
+++ b/frontend/src/routes/events/OverviewTabs/LeaderboardTab/index.tsx
@@ -2,7 +2,7 @@ import { radixTheme } from '@/grid';
 import { useLeaderboard } from '@/hooks/events';
 import { Container, Flex, Spinner } from '@radix-ui/themes';
 import { AgGridReact } from 'ag-grid-react';
-import { ErrorCallout } from 'components/Callouts';
+import { ErrorCallout, WarningCallout } from 'components/Callouts';
 import { useParams } from 'react-router';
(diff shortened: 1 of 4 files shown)


frontend/src/routes/events/OverviewTabs/LeaderboardTab/index.tsx
```diff
@@ -15,45 +15,51 @@ export default function LeaderboardTab() {
...
-          paginationPageSize={20}
-          domLayout="autoHeight"
-          loadingOverlayComponent={Spinner}
-          suppressCellFocus
-          suppressHeaderFocus
-        />
+        {!leaderboardError && (
+          <AgGridReact
+            rowData={leaderboard}
+            columnDefs={[
```
cjones-cs, line 27: Should we break these colDefs outside the function like we did in other places?

trevnels-cs: Shouldn't need to be changed here - admin grids needed it due to some nav behavior, this one doesn't

cjones-cs (approved):

Issue #358, opened 2025-11-12
Expose ctfd ban behavior


trevnels-cs:
- Allow admins to ban users
  - This will immediately invalidate all existing sessions for that user
  - This will prevent future logins with a user is banned message
  - Banned users shouldn't be able to use any user_endpoints, in case sessions don't get invalidated properly
- Expose a ban/unban button on the user admin panel. Maybe consider including a "reason" field
- Use the underlying ctfd user banned flag (https://github.com/CTFd/CTFd/blob/master/CTFd/models/__init__.py#L413)

- What happens to teams the banned user is on?
  - Nothing automatically I think - if captain is banned, admin can go assign somebody else manually

```
POST /admin/users/[id]/ban
POST /admin/uesrs/[id]/unban
```

fpayer:
Don't worry about ban reason for now. That'll come later

fpayer:
Nothing to happen to teams automatically

PR #454, opened 2025-11-26
Expose ctfd ban/unban behavior

CarterPerez-dev:
- Add  `ban_controller.py` to to use ctfd's ban feature
- Add `/ng/admin/users/<int:user_id>/unban`  & `/ng/admin/users/<int:user_id>/ban` endpoints
- Add `test_user_ban.py` to test banning functionality
- Edit login endpoint to check for banned users & `auth.py`  decorator as fallback

Closes issue: https://github.com/CyberSkyline/ctf-ng/issues/358


fpayer (approved):

trevnels-cs:
Do we have a way to for admins to tell if a user has been banned? Might be worth passing the flag through the user serialize for admins for UI purposes.

trevnels-cs (approved):

PR #455, opened 2025-12-01
Move image pull to celery worker

RayofLightz:
Migrates the pull thread to a dedicated celery worker because gunicorn doesn't let you just run a background thread.


backend/ng/containers/utils/Client.py
```diff
@@ -69,6 +71,8 @@ def pull_image(self, image):
                     "username": get_app_config("CONTAINER_REGISTRY_USER"),
                     "password": get_app_config("CONTAINER_REGISTRY_PASSWORD"),
                 }
-            self.images.pull(image, auth_config=auth)
+            host = get_app_config("DOCKER_HOST")
+            print(auth)
```
fpayer: Remove debug

fpayer (approved):

zoecough (approved):

PR #459, opened 2025-12-02
String validation / sanitization

### LINKED ISSUE #457: String validation / sanitization
Update the validator to allow for the following optional configurations

- Trim extra whitespace at the beginning / end of the string (default on)
- Validate that the string only contains printable ASCII characters (default off)

droge91:
updates the string validator to optionally

trim whitespace
validate that the string contains only printable ASCII characters

closes  #457

### DIFF
--- a/backend/ng/core/utils/validator.py
+++ b/backend/ng/core/utils/validator.py
@@ -100,6 +100,7 @@ class BaseValidator:
         required: bool = False,
         friendly_name: str | None = None,
         value: Any = None,  # Injected by decorator
+        printable_only: bool = False,
     ) -> None:
         if not isinstance(value, str):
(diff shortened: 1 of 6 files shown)


fpayer:
Can you also add the printable-only validation to support tickets (subject + messages)?

backend/ng/core/utils/validator.py
```diff
@@ -116,6 +118,10 @@ def validate_string(
             )
             return
 
+        if printable_only and (stripped_value.isprintable() is False or stripped_value.isascii() is False):
+            self.errors[field] = f"{friendly_name} contains non-printable characters"
```
trevnels-cs: I think this error message could be better, since this will also block unicode, not just non-printables. Maybe just something like "... contains disallowed characters"

trevnels-cs (approved):
Looks good, though consider enabling `trim_whitespace` for tickets as well

Issue #444, opened 2025-11-25
Require sponsor before event registration


cjones-cs:
Require sponsor selection before event registration. -> message and link to profile page
Put message on profile page about selecting a sponsor before the event registration.

If a sponsor is not selected, redirect from the login form to the profile page.

Backend: require sponsor prior to event registration. Throw an error if not sponsored

PR #460, opened 2025-12-02
Add messages to select sponsor

cjones-cs:
https://github.com/CyberSkyline/ctf-ng/issues/444

image

The Registration Card should block them from getting to the Registration Modal at all, but if they are able to get to it, the backend will prevent registration before they have a sponsor:
image

I added one test case to account for the new sponsor requirement, but a bunch of test cases needed updated to accommodate this change as well

### DIFF
--- a/backend/ng/admin/tests/test_admin_integration.py
+++ b/backend/ng/admin/tests/test_admin_integration.py
@@ -103,9 +103,10 @@ class TestAdminImpersonation:
         data = response.get_json()
         assert data["data"] == []
 
-    def test_admin_team_management_as_user(self, admin_client, user_factory, team_factory, event_factory):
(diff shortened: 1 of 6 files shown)


trevnels-cs:
Adding logic to not open the registration modal automatically when clicking an invite link if the user's sponsor isn't set might be a good idea - that way they will see the message underneath instead of filling in the modal and getting an error.

cjones-cs:
> Adding logic to not open the registration modal automatically when clicking an invite link if the user's sponsor isn't set might be a good idea - that way they will see the message underneath instead of filling in the modal and getting an error.

This won't happen anyway because the registration modal isnt being rendered at all

trevnels-cs (approved):

PR #462, opened 2025-12-02
Misc admin panel accessibility updates

trevnels-cs:
Improvements to roles/aria tags for admin pages

### DIFF
--- a/frontend/src/components/Statistic.tsx
+++ b/frontend/src/components/Statistic.tsx
@@ -35,13 +35,17 @@ export default function Statistic({
       <Flex direction="row" align="center" gap="1">
         <Text color="gray" size="2">{label}</Text>
         {description && (
-        <Tooltip content={description}>
-          <Text color="gray"><TbInfoCircle /></Text>
-        </Tooltip>
+          <Tooltip content={description}>
+            <button type="button">
+              <Text color="gray">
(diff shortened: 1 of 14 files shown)


frontend/src/components/Statistic.tsx
```diff
@@ -35,13 +35,17 @@ export default function Statistic({
...
+              <Text color="gray">
+                <TbInfoCircle aria-label="More info" />
+              </Text>
+            </button>
+          </Tooltip>
         )}
         {delta && <Badge color={deltaColor} variant="soft" radius="full">{delta}</Badge> }
       </Flex>
-      <Heading size={size} className="tabular-nums">{value}</Heading>
+      <Heading size={size} asChild className="tabular-nums"><span>{value}</span></Heading>
```
cjones-cs, line 48: Wouldn't it be better to use a <p> here instead of a span since spans don't have semantic meaning to screen readers? Not really sure why the heading needed changed

trevnels-cs: Statistic values aren't semantically headings, but were being treated as such by screen readers. Applying heading styling to a span keeps the visual emphasis, but without the screen reader using the stat values as semantic headings.

cjones-cs (approved):

PR #463, opened 2025-12-02
Serve NoVNC application from our code

trevnels-cs:
`vnc.html` is at `/static/novnc/vnc.html` or `/dist/novnc/vnc.html` in prod. We then use its path parameter to point at websockify.

Also updates the workspace startup polling in the UI to be compatible with the new approach as well as more reliable.


trevnels-cs:
~~Looks like some of the formatting changes from 
[Updated env vars for aws config](https://github.com/CyberSkyline/ctf-ng/pull/463/commits/178669b53e0ed62b0328dbe3a231461453689546) made it into `vite-static-folder` which this branch is based on~~ Rebased

RayofLightz:
We should consider writing a script using docker cp to be able to generate the no vnc assets.

trevnels-cs:
> We should consider writing a script using docker cp to be able to generate the no vnc assets.

The official source of truth for NoVNC is their GitHub release page - there's no officially maintained container image. My thinking is to have a script that grabs the latest release and deletes the files that aren't needed for their frontend. Though NoVNC changes infrequently enough that I'm not sure if this is a priority (the last stable release was in March of this year)

frontend/public/test.txt
fpayer, line 1: Safe to remove this test.txt file?

fpayer:
Have you been able to verify that the urls don't break in both the vite dev and vite build modes?

trevnels-cs:
> Have you been able to verify that the urls don't break in both the vite dev and vite build modes?

Things are working in dev mode. I can't run the prod environment directly, but the iframe path that's emitted in the production build assets is correct given our nginx config.

fpayer (approved):

cjones-cs (approved):

PR #464, opened 2025-12-02
Update SSO authentication to match on oauth_id instead of email

fpayer:


fpayer:
DO NOT MERGE UNTIL VERIFIED IN STAGING

mblunt (changes requested):
[Line 107:
](https://github.com/CyberSkyline/ctf-ng/pull/464/commits/608014fd3c4c6bba2b38c45588d7b366a1a9a176#diff-90c1bc35acb2a7739c2c4789f4e32f0c74b3ad5bda6ae0dff97d6cf794b7e69eR107)`ctfd_user = User.query.filter_by(id=ng_user.id)` returns a Query object. Need to add a `.first()` to actually get the `User` object back out.

[(old) line 122-123:
](https://github.com/CyberSkyline/ctf-ng/pull/464/commits/608014fd3c4c6bba2b38c45588d7b366a1a9a176#diff-90c1bc35acb2a7739c2c4789f4e32f0c74b3ad5bda6ae0dff97d6cf794b7e69eL122): Seems like we used to also update the name and email if it changed. Now, we just update the `oauth_id`? Do we want OAuth to be able to update the email and name? Its fine if not, but if the user changes their name/email on `login.gov` we won't store the updates.

fpayer:
Good catch with query filter. Just pushed a fix. 

re: name/email - I got updated guidance that there is a desire for users to be able to configure those values on the platform directly (which will be a future feature). For that feature to work, we can't automatically update the name/email from Okta.

mblunt (approved):
lgtm!

fpayer:
Verified in staging

PR #466, opened 2025-12-02
Fix time windows in team validation

### LINKED ISSUE #445: Renaming team produces time constraint validation error
Image

This is an event with no time limit, so start time is present but end time isn't. Also note that it's an admin doing that update, so those fields are being passed through the validator.

trevnels-cs:
Resolves #445. I thought this wasn't significant, but it can prevent admins from doing administrative tasks in certain cases so I think it's worth fixing.

`validate_time_window` enforces that both start and end must be present, which is not guaranteed for teams. Teams can have a start date but no end date when playing an event with no time limit. When doing a self validate, end_time will always be `in` data, but it might not actually be set, which causes validation to stop working entirely.

### DIFF
--- a/backend/ng/team/models/Team.py
+++ b/backend/ng/team/models/Team.py
@@ -152,11 +152,9 @@ class Team(db.Model):
             )
             if res:
                 raise ValidationError(f"Team name '{data['name']}' already exists in event ID {data['event_id']}.")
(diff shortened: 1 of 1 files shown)


droge91 (approved):
Looks good

PR #467, opened 2025-12-02
Use ROUTEPREFIX for auth navigations

trevnels-cs:
Follows ROUTEPREFIX for programmatic navigation that is done during authentication. This ensures users are directed to the correct place in production environments.


fpayer:
Related - can you also remove `ROUTEPREFIX` from the `SSOPATH` in `constants.ts`? `SSOPATH` is getting fed into the react-router `Link`, which is configured to prepend `ROUTEPREFIX` already, so the prefix gets added twice.

fpayer (approved):

cjones-cs (approved):

PR #468, opened 2025-12-02
Remove placeholder/outdated copy text from dashboard

trevnels-cs:
- Removed reference to practice area in not registered message - only directs to event page now.
- Removed placeholder callout for no events running. The header now goes away if nothing is up there.


cjones-cs (approved):
Fix the lint error, but lgtm

PR #473, opened 2025-12-02
Admin user update modal

### LINKED ISSUE #330: Admin - User Updates
- [ ] Allow an admin to edit a users name. Do Not Edit the email bc this will break SSO login. A user would submit a support ticket to get an admin to do this for them.
- [ ] Assign Roles to an individual user. 

Roles have associated permissions, but those will be associated by a config file manually.

trevnels-cs:
Resolves #330.

Modal for updating users and their associated roles. This is a bit weird since it has to hit two endpoints - it's theoretically possible for one to succeed but not the other, but we don't have a way of doing both atomically at the moment.

Eventually I'd like to allow configuring a user's sponsor through here as well, but that's not a priority and the endpoints for doing so don't exist yet. Dynamically populating the roles with a better interface to assign them would also be nice long-term.

While the role update UI is present, its functionality is currently blocked on #420 - the endpoint doesn't currently work. Once #422 merges, it will work.

image

### DIFF
--- a/frontend/src/hooks/users.ts
+++ b/frontend/src/hooks/users.ts
@@ -148,6 +148,15 @@ export function useWorkspaceStatus(userId : number) {
   });
 }
 
+export function adminUpdateUser(userId: number, user: Pick<User, 'name' | 'email'>) {
+  return apiMutation(`/admin/users/${userId}`, user, {
+    method : 'PUT',
+  }).then(() => {
+    mutate(`/admin/users`);
+    mutate(`/admin/users/${userId}`);
+  });
+}
+
 export function restartWorkspace(userId : number) {
   return apiMutation(`/admin/users/${userId}/container/restart`, undefined, {
     method : 'POST',
--- a/frontend/src/routes/admin/users/AdminUpdateUserModal.tsx
+++ b/frontend/src/routes/admin/users/AdminUpdateUserModal.tsx
@@ -1,0 +1,118 @@
+import { COLOR_WARNING } from '@/constants';
+import { adminUpdateUserRoles } from '@/hooks/permissions';
+import { adminUpdateUser } from '@/hooks/users';
+import type { User } from '@/types';
+import {
+  Box,
+  Button,
+  Flex,
+  Switch,
+  TextField,
+} from '@radix-ui/themes';
+import FormField from 'components/FormField';
+import Modal from 'components/Modal';
+import { Controller } from 'react-hook-form';
+import { TbPencil } from 'react-icons/tb';
+
+export default function AdminUpdateUserModal({ user }: {user: User}) {
+  return (
+    <Modal
+      title="Edit User"
+      trigger={(
+        <Button variant="soft" color={COLOR_WARNING}>
+          <TbPencil />
+          Edit
+        </Button>
+      )}
+      onSubmit={async (data) => {
+        const roles = [];
+        if (data.admin) {
+          roles.push('admin');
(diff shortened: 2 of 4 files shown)


frontend/src/hooks/users.ts
```diff
@@ -149,6 +148,15 @@ export function useWorkspaceStatus(userId : number) {
...
 
+export function adminUpdateUser(userId: number, user: Pick<User, 'name' | 'email'>) {
+  return apiMutation(`/admin/users/${userId}`, user, {
+    method : 'PUT',
+  }).then(() => {
+    mutate(`/admin/users`);
+    mutate(`/admin/users/${userId}`);
+  });
+}
+
```
cjones-cs, line 159: Wouldn't it have been better to put these as one endpoint and then disperse the functionality in the controller?

frontend/src/routes/admin/users/AdminUpdateUserModal.tsx
```diff
@@ -0,0 +1,131 @@
...
+            )}
+          </FormField>
+          <FormField label="Email" error={errors.email}>
+            {(injected) => (
+              <TextField.Root
+                {...register(
+                  'email',
+                  { required : 'Email is required', maxLength : { value : 128, message : 'Name cannot be longer than 128 characters' } },
+                )}
+                {...injected}
```
cjones-cs, line 71: I thought the requirements said email had to be immutable because it would break the sso connection?

frontend/src/routes/admin/users/AdminUpdateUserModal.tsx
```diff
@@ -0,0 +1,131 @@
...
+  Switch,
+  TextField,
+} from '@radix-ui/themes';
+import FormField from 'components/FormField';
+import Modal from 'components/Modal';
+import { pick } from 'lodash';
+import { Controller } from 'react-hook-form';
+import { TbPencil } from 'react-icons/tb';
+
+export default function AdminUpdateUserModal({ user }: {user: User}) {
```
cjones-cs, line 18: Alternatively to a model, you could've had inline editing of the name in the admin side bar, then have this modal for just the roles. We have precedence in the Admin Ticket section. This would get rid of the wonkiness with the two endpoints

cjones-cs (changes requested):

frontend/src/routes/admin/users/AdminUpdateUserModal.tsx
trevnels-cs: SSO is getting switched over to an ID and is no longer email-dependent (#464)

frontend/src/hooks/users.ts
trevnels-cs: Since the two endpoints aren't guaranteed to perform an atomic update, I want to keep them separate to ensure that's known. Hiding that possible edge case away inside of a single hook seems risky to me

frontend/src/routes/admin/users/AdminUpdateUserModal.tsx
trevnels-cs: I would eventually like a lot of this to be inline, though I'd want to make reusable components that can be used on other admin panels as well. Things like error handling that the modal handles automatically will need to be reworked for those inline editors. For now, the modal is the simplest way of getting the functionality working, but I do agree that inline editors would be better here long-term

cjones-cs (approved):

Issue #430, opened 2025-11-24
File Upload Standardization


cjones-cs:
https://github.com/CyberSkyline/ctf-ng/pull/428

During sponsor uploads, found that some of the backend services aren't pulling the same data. This makes it hard to define the UploadedFile type on the frontend. Need to standardize the backend to remove extra data in the `list` and `file` endpoints.
This should be mostly okay on the frontend, but a few instances where nesting is removed will need updating.

Ensure all frontend consumed functions are using the `success_response` and `error_response` wrappers.

Remove the randomization of the filenames and allow the user to keep the filename of whatever they uploaded. Perform a dup check and either append (1) to the end or ask for overwriting.
If overwriting, we could add a check box on the ui to toggle this as well.

trevnels-cs:
- Only admins should be able to upload to the favicons, game cards, and sponsor logo folders. (#435)
- Support tickets should be given isolated folders - users may be able to overwrite files on other tickets with the same name if all ticket uploads share a single path.
- File list/search endpoints should be locked down to admins only - users shouldn't have the ability to enumerate all files in a folder (apart from maybe their own support tickets)
- Enforce size/rate limits for file uploads. We don't want people uploading giant files and crashing the system.

Issue #435, opened 2025-11-25
Restrict non-admin file uploads


trevnels-cs:
Currently all users can upload favicons, game cards, etc.

This shouldn't be allowed.

trevnels-cs:
Resolved in #474

PR #474, opened 2025-12-03
S3 fileuploads updates

zoecough:
- moved URL expiration to config
- preserving filenames on upload
- optional file overwriting or number appending
- switched to success_response and error_response wrappers for responses
- better output standardization on endpoints
- optional download url generation on more endpoints
- decorated endpoints (everything except file download)
Closes issues #430 and #435

### DIFF
--- a/backend/ng/fileuploads/controllers/public_files.py
+++ b/backend/ng/fileuploads/controllers/public_files.py
@@ -90,7 +90,8 @@ def generate_upload_url(args):
         # Optionally include download URL
         include_urls = args.get('include_urls', False)
         if include_urls:
-            download_url = s3_service.generate_download_url(folder, final_filename)
+            object_key = f"{folder}/{final_filename}"
+            download_url = s3_service.generate_download_url(object_key, expires_in=config.S3_DOWNLOAD_URL_EXPIRATION)
             uploaded_file["download_url"] = download_url
 
         return success_response(uploaded_file)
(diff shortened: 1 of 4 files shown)


backend/ng/fileuploads/controllers/public_files.py
cjones-cs, line 174: Since each file contains a folder property, do we still need to return folder as a separate item here? I also don't think we need count since frontend can get that other ways

backend/ng/fileuploads/controllers/public_files.py
```diff
@@ -124,38 +178,49 @@ def list_public_files(args):
...
+                    try:
+                        download_url = s3_service.generate_download_url(object_key, expires_in=config.S3_DOWNLOAD_URL_EXPIRATION)
+                        file_info['download_url'] = download_url
+                    except Exception as e:
+                        logger.warning(f"Failed to generate download URL for {object_key}: {e}")
+                        file_info['download_url'] = None
+
+                return success_response({
+                    "files": [file_info],
+                    "count": 1
```
cjones-cs: count?

cjones-cs (changes requested):

cjones-cs:
Just a note for later, but with these changes, we have the option of putting a checkbox on the Add Sponsor Modal. Also we need to verify that the nesting for the sponsors selection on the profile page is working since this should remove a nested item from the return. As such, I'm not sure if we should block this pr until the minor UI work is added to it or just make sure we have that fixed asap this week. All this bc it does technically work in the dev branch rn

trevnels-cs (approved):
Confirmed that these updates are working. Requires some minor frontend updates (mainly `url` -> `download_url` on the type info,) but the security fixes here are important enough that we should merge this and make any required frontend changes.

backend/ng/fileuploads/controllers/public_files.py
```diff
@@ -168,36 +229,32 @@ def search_public_files():
...
+                    except Exception as e:
+                        logger.warning(f"Failed to generate download URL for {obj['key']}: {e}")
+                        file_info['download_url'] = None
+
+                files.append(file_info)
+
+        return success_response({
             "folder": folder,
             "files": files,
             "count": len(files)
```
cjones-cs: Folder/count here?

cjones-cs (changes requested):

cjones-cs (approved):

PR #476, opened 2025-12-03
User banning frontend and fixes

### LINKED ISSUE #358: Expose ctfd ban behavior

trevnels-cs:
- Adds ban/unban modal to users on the frontend
- Adds missing table columns (`banned` and `affiliation`)
- Backend fixes
    - Deregister CTFd's `before_request_func` for ban checking (previously resulted in a 500 for banned users) to instead fall through to our ban logic/message
    - Disallow impersonating banned users (banned user can't do anything and can't exit impersonation, so you end up getting stuck and having to clear cookies to escape)

Closes #358

### DIFF
--- a/frontend/src/routes/admin/users/index.tsx
+++ b/frontend/src/routes/admin/users/index.tsx
@@ -1,12 +1,12 @@
 import { useAllUsers } from '@/hooks/users';
-import type { User } from '@/types';
+import type { AdminUser } from '@/types';
 import type { ColDef } from 'ag-grid-community';
 import AdminGrid from 'components/AdminGrid';
 import { ErrorCallout } from 'components/Callouts';
 import RoleBadge from 'components/RoleBadge';
 import UserSidebar from './UserSidebar';
 
-const colDefs: ColDef<User>[] = [
+const colDefs: ColDef<AdminUser>[] = [
   {
     field : 'id',
     width : 100,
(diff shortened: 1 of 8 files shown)


frontend/src/routes/admin/users/index.tsx
```diff
@@ -46,6 +46,19 @@ const colDefs: ColDef<User>[] = [
     filter : true,
     floatingFilter : true,
   },
+  {
+    field : 'affiliation.name',
+    headerName : 'Affiliation',
```
cjones-cs: Change header name to "Sponsor"

cjones-cs (changes requested):
Sponsor is the ui name. Affiliation is only used on the backend and will be changed in v2 to Sponsor

cjones-cs (approved):

PR #483, opened 2025-12-04
Game cards in S3

### LINKED ISSUE #267: Allow configuring game cards as part of the event
Game cards are currently hardcoded. This should instead be a parameter on the event model to allow for different game cards to be selected for each event.

trevnels-cs:
Resolves #267.

EventDataForm can select/upload files to S3.

Game cards pull their configured image from S3 across the application.

image


frontend/src/routes/admin/events/EventDataForm.tsx
```diff
@@ -6,15 +7,40 @@ import {
...
 export default function EventDataForm({
   rhf,
 }: {
   rhf: UseFormReturn<Omit<Event, 'id'>>,
 }) {
-  const { register, formState : { errors } } = rhf;
+  const {
+    register, control, watch, formState : { errors }, setValue, setError, clearErrors,
+  } = rhf;
+  const { data : gameCards, error : gameCardsError, isLoading : gameCardsLoading } = useFileList('event-cards');
```
cjones-cs, line 25: I wonder if we should put the folder names as const values since we will be using them all over the place, but that can also happen later down the line

frontend/src/routes/admin/events/UploadButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+        <Button
+          {...getRootProps()}
+          color={COLOR_POSITIVE}
+          loading={loading}
+          variant="soft"
+          type="button"
+        >
+          <input {...getInputProps()} />
+          <TbUpload aria-label="Upload" />
+        </Button>
```
cjones-cs, line 29: When I put the dropzone for sponsors, I included messaging to indicate that images would be auto-uploaded on drop instead of on Save. We could probably benefit from some kind of notice here

cjones-cs (approved):

frontend/src/routes/admin/events/UploadButton.tsx
trevnels-cs: Might be something to revisit when this UI gets converted to inline fields - I don't think there's much room left for extra text in the current modal

Issue #353, opened 2025-11-12
Clicking links in the user dropdown menu does not close the menu


trevnels-cs:
Clicking links (i.e. Profile, Admin Portal) in the user dropdown leaves the menu open afterwards, requiring it to be manually closed to get it out of the way.

PR #492, opened 2025-12-04
Fix/clickaway menu

cjones-cs:
https://github.com/CyberSkyline/ctf-ng/issues/353


trevnels-cs (approved):
Working on my end, and no longer able to reproduce #223 (what these handlers were originally added for) with recent dropdown layout changes.

PR #495, opened 2025-12-04
Countdown timer for team end

### LINKED ISSUE #382: Determine how to show remaining time for an event
Currently, teams have no way of knowing when their timer will expire apart from recording the time they started and tracking on their own. We should introduce a component on the event page to show this, likely just the team's end date for now. A live countdown component might be better depending on how complex it would be to implement.

Here's the event page, for reference:

Image

I'm thinking either below the "registration card" element, or above the challenge search bar would be a good spot. Maybe even inside the registration card, in the form of "Your team has until \<time\> to..."

### LINKED ISSUE #471: Add time indicator for how much time is left remaining on the event page
Needs to be relative to the team's time remaining or the event stop time, whichever is sooner.

trevnels-cs:
Show countdown timer for team end time on event/challenge pages, and refresh state when it reaches 0. Can click to toggle hiding the display (shows as `--:--:--` when hidden so it can still be clicked to unhide)

Resolves #382 and #471

image
image

### DIFF
--- a/frontend/src/components/Timer.tsx
+++ b/frontend/src/components/Timer.tsx
@@ -1,0 +1,86 @@
+import {
+  Box,
+  Flex,
+  Heading,
+  Text,
+} from '@radix-ui/themes';
+import { useEffect, useState } from 'react';
+import { twMerge } from 'tailwind-merge';
+
+export default function Timer(
+  {
+    target,
+    size = '6',
+    onEnd,
+  } : {
+    target: Date | null,
+    size?: React.ComponentProps<typeof Heading>['size'],
+    onEnd?: () => void
+  },
+) {
+  const [ timerString, setTimerString ] = useState('');
+  const [ hidden, setHidden ] = useState(false);
+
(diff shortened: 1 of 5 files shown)


frontend/src/components/Timer.tsx
```diff
@@ -0,0 +1,86 @@
...
+      setTimerString(`${hours}:${minutes}:${seconds}`);
+    };
+
+    updateTimer();
+
+    // if there's no target date, do not set interval
+    if (!target) return () => {};
+
+    // if hidden, do not set interval
+    if (hidden) return () => {};
```
cjones-cs, line 56: Could we have combined those two with an OR?

cjones-cs (approved):

frontend/src/components/Timer.tsx
trevnels-cs: Technically yes, though I think it's more readable to have these conditions be explicit

PR #496, opened 2025-12-05
Add Load Testing tooling to repo

johnmarsden-edu:
Using k6 and a docker compose file with configuration options to handle actually running the load test, want to get other eyes on it before we put it on staging and run a load test. I need to 
- [x] revert a change I made and make them share load test user info again 
- [x] make the load test users be randomly generated and save the credentials locally for use by the load tester
- [x] load the credentials in instead of generating them deterministically to avoid having many load testing users with deterministically generated creds.


fpayer:
Should the `admin_user.ts` and `bad_actor.ts` files be empty?

tools/load-test/scenarios/default_browser_user.ts
```diff
@@ -0,0 +1,53 @@
...
+        await this.page.goto(this.url('/'));
+
+        const loginButton = this.page.locator('button', { hasText: 'Log In' });
+        await loginButton.click();
+
+        const loginLink = this.page.locator('a', { hasText: 'Log In' });
+        await loginLink.click();
+        await this.page.waitForNavigation();
+
+        const expoAccountLink = this.page.locator('a', { hasText: 'use an expo account' });
```
fpayer, line 29: We added a mandatory gov notice banner in the login flow. Not sure if you need to account for that still

johnmarsden-edu:
> Should the `admin_user.ts` and `bad_actor.ts` files be empty?

I hadn't had the time to go through and implement those yet.

tools/load-test/scenarios/default_browser_user.ts
johnmarsden-edu: I will but that user browser session doesn't get used yet. I was primarily focused on getting the http one running correctly so that we can have a bunch of them running at the same time.

backend/ng/scripts/populate_sample_data.py
```diff
@@ -14,6 +14,9 @@
 if "SCRIPT" not in os.environ:
     raise OSError("This should only be run from a script. DO NOT run this manually.")
 
+if "LOAD_TEST" not in os.environ:
+    raise OSError("LOAD_TEST environment variable not set. Absorting...")
```
johnmarsden-edu:
```suggestion
```

PR #500, opened 2025-12-05
Set netconfs when updating challenge

trevnels-cs:
Ensure that netconfs are updated when challenge is. Exceptions can be thrown at (`netconf = blueprint_obj.netconfs[network] or None`) due to bracket indexing otherwise if the netconfs array is out of date.

It might make sense to replace the brackets with `.get` for safety as well. It's currently possible to use a network in a container that's not defined in netconfs, which will result in the exception there, resulting in a half-functioning deployment that's a pain to clean up after.


RayofLightz (approved):
Can you send me the Yaml that you had so I can reproduce the netconfs issue?

PR #517, opened 2025-12-08
Container updates batch 2

RayofLightz:
* Add psk to service exec
* Add individual bridges
* Add vnc resource limits


backend/ng/containers/models/IndvidualContainer.py
```diff
@@ -77,22 +77,38 @@ def render_container_name(user_id) -> str:
...
+        if not net:
+            net = client.networks.create(name=container_name, driver="bridge")
+
+
         return client.containers.run(
             NOVNC_CONTAINER,
             name=container_name,
             detach=True,
             publish_all_ports=True,
+            mem_limit="4g",
```
zoecough: could these be set in the config or is this better?

services/exec-service/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "exec-service",
   "private": true,
-  "type": "module",
```
zoecough, line 4: why isn't it a module anymore?

zoecough (approved):
lgtm

services/exec-service/package.json
RayofLightz: This was causing issues because the code is not using ES modules.

backend/ng/containers/models/IndvidualContainer.py
RayofLightz: The mem limit was specifically requested as 4g. It might be worth breaking out into configs just incase someone uses a different image and wants lower/higher limits.

backend/ng/containers/models/IndvidualContainer.py
```diff
@@ -85,18 +96,25 @@ def run_container(client, container_name):
             cap_add=[
                 "NET_ADMIN",
                 "SYS_PTRACE",
-            ]
+            ],
+            mem_limit=NOVNC_RAM,
+            cpu_period=200000,
+            cpu_quota=100000,
+            pids_limit=2000,
```
zoecough, line 103: Should these be in config/the env too?

zoecough (approved):

PR #534, opened 2025-12-10
Fix/reduce logging

CarterPerez-dev:
reduce 4xx errors to info, add request context


backend/ng/core/middleware/error_handler.py
```diff
@@ -17,25 +17,37 @@
 logger = get_logger(__name__)
 
 
+def _get_request_context() -> dict:
+    """
+    Get basic request context for logging
+    """
+    try:
+        return {
```
trevnels-cs: Consider including `session.get("id")` here to log the user ID if they're authed

backend/ng/core/middleware/error_handler.py
```diff
@@ -7,8 +7,7 @@
 from functools import wraps
 
 from CTFd.models import db
-from flask import current_app as app
-from flask_limiter.errors import RateLimitExceeded
```
trevnels-cs, line 11: This import shouldn't be removed since it's being used in this file on the current development branch - you may need to merge development into this branch again

trevnels-cs (approved):

PR #535, opened 2025-12-10
Fix admin ticket sidebar stale state and UI bugs

### LINKED ISSUE #533: BUG: Ticket assignment assigns wrong ticket
When I assign a ticket to another user, it assigns the next ticket in the list rather than the current one. 

Actually, I'm not exactly sure what the logic/effect is - appears to change each time I select a different ticket.  In my recording (sent separately via chat), I only assigned one ticket (#33) to another user.  As you can see when I click around, sometimes it's appears to be assigned to me, sometimes to him....and other tickets also appear to change back and forth each time I click on them.

CarterPerez-dev:
*(Note: Reason I took this issue is becuase I initially assumed this was a backend issue, but after investigating I found
  the bug was in the frontend, so I thought might as well take care of it)*

- Reset sidebar dropdown states when switching between tickets using useEffect to sync state with selectedRow.id changes
- Fix stopCellSelection using wrong field names (author_id -> author_name, etc.) which caused navigation flicker when clicking entity links in grid rows
- Handle null values properly for assigned_to, event_id, and challenge_id to fix Unassign button incorrectly being enabled on unassigned tickets

Closes issue: https://github.com/CyberSkyline/ctf-ng/issues/533

https://github.com/user-attachments/assets/02cd81a5-1211-498b-b788-66c7f377c8d8

### DIFF
--- a/frontend/src/routes/admin/tickets/MessagesSidebar.tsx
+++ b/frontend/src/routes/admin/tickets/MessagesSidebar.tsx
@@ -45,17 +45,24 @@ import {
   map,
   without,
 } from 'lodash';
-import { useId, useState } from 'react';
+import { useEffect, useId, useState } from 'react';
 import { TbMessage, TbX } from 'react-icons/tb';
 
 export default function MessagesSidebar({ entity : selectedRow }: { entity: AdminTicket }) {
   // Dropdowns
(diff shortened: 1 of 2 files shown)


trevnels-cs (approved):
Confirmed to resolve #533 on my end

Issue #510, opened 2025-12-08
Support ticket email notifications contain incorrect link to the ticket


fpayer:
The "View Ticket" link in the email sent to the admin support inbox take you to the user-facing support ticket page, not the admin-facing support ticket page.

PR #536, opened 2025-12-10
Fix email ticket links to use correct URL for recipient type

CarterPerez-dev:
- Add ADMIN_TICKET_URL_PATH config for admin ticket routes
- New ticket emails (to admin inbox) now link to /admin/tickets?id=X
- Reply emails use recipient-appropriate URLs:
  - Admin reply → user gets /support/{id}
  - User reply → admin gets /admin/tickets?id={id}

closes issue: https://github.com/CyberSkyline/ctf-ng/issues/510

### DIFF
--- a/backend/ng/emails/services/email_templates.py
+++ b/backend/ng/emails/services/email_templates.py
@@ -104,12 +104,13 @@ class TicketEmailTemplates:
         return f"{server_domain}{route_prefix}"
 
     @staticmethod
-    def new_ticket(ticket_data: TicketData) -> tuple[str, str, str]:
+    def new_ticket(ticket_data: TicketData, is_admin_recipient: bool = True) -> tuple[str, str, str]:
         """
         Generate new ticket notification email
 
         Args:
             ticket_data: Serialized ticket data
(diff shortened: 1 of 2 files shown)


backend/ng/emails/services/email_templates.py
```diff
@@ -104,12 +104,13 @@ def _get_base_url() -> str | None:
         return f"{server_domain}{route_prefix}"
 
     @staticmethod
-    def new_ticket(ticket_data: TicketData) -> tuple[str, str, str]:
+    def new_ticket(ticket_data: TicketData, is_admin_recipient: bool = True) -> tuple[str, str, str]:
```
fpayer: Don't any calls to `new_ticket` need to be updated to specify `is_admin_recipient`?

CarterPerez-dev: it defaults to true, I added it for consistency with if is_admin_reply to reduce confusion but looking at it now it probably just *adds* confusion/complexity from the implied logic (new tickets are always admin). I removed it

fpayer (approved):

Issue #504, opened 2025-12-05
Feedback data has no validation


trevnels-cs:
While we do want a fairly dynamic schema for feedback data, individual fields still need to have validation based on a common set of rules.

For example, strings need a character limit. I can currently submit the entire Bee Movie script as feedback and the system will accept it.

PR #537, opened 2025-12-10
Add validation limits for feedback_data JSON fields

CarterPerez-dev:
- Add config constants for feedback validation limits:
  - FEEDBACK_MAX_STRING_LENGTH (5000 chars)
  - FEEDBACK_MAX_KEYS (50 fields)
  - FEEDBACK_MAX_KEY_LENGTH (100 chars)
- Validate feedback_data on create and update
- Add tests for string length validation

Closes issue: https://github.com/CyberSkyline/ctf-ng/issues/504

- [x] Must merge before: https://github.com/CyberSkyline/ctf-ng/pull/515
(actually I dont think it matters...)


trevnels-cs:
We should probably also configure MAX_CONTENT_LENGTH at the global flask level to put a hard limit on POST body size that's enforced before any parsing occurs. The validation introduced here only occurs after the data is parsed, so a sufficiently large payload could still cause performance problems or DoS due to parsing overhead. This would also cover all other endpoints which suffer from the same potential issue

backend/ng/feedback/models/Feedback.py
```diff
@@ -92,6 +93,34 @@ def serialize(self, include_admin_fields: bool = False) -> SerializedFeedback:
...
+                errors={"feedback_data": f"Too many fields (max {config.FEEDBACK_MAX_KEYS})"}
+            )
+
+        errors = {}
+        for key, value in feedback_data.items():
+            if len(key) > config.FEEDBACK_MAX_KEY_LENGTH:
+                errors[key] = f"Field name cannot exceed {config.FEEDBACK_MAX_KEY_LENGTH} characters"
+                continue
+
+            if isinstance(value, str) and len(value) > config.FEEDBACK_MAX_STRING_LENGTH:
```
trevnels-cs: Consider adding a case for numbers as well, since those could also come via JSON and bypass the string length check here. Nested dicts are also not length checked here. A simpler approach might be to `str` the entire feedback object and check the length of that instead of working with individual keys

trevnels-cs:
It looks like tests need to be updated to account for the changes to this logic

trevnels-cs (approved):

PR #541, opened 2025-12-10
Challenge feedback submission

### LINKED ISSUE #424: Challenge feedback interface

trevnels-cs:
image

image


frontend/src/components/FormField.tsx
```diff
@@ -22,7 +24,10 @@ export default function FormField({
 
   return (
     <Flex direction="column" gap="1">
-      {label && <label htmlFor={id} data-invalid={error ? 'true' : 'false'}>{label}</label>}
+      <Flex direction="row" justify="between" align="end" className="empty:hidden">
+        {label && <label htmlFor={id} data-invalid={error ? 'true' : 'false'}>{label}</label>}
+        {rightLabel && <Text size="2" color={rightLabel.startsWith('-') ? 'red' : 'gray'}>{rightLabel}</Text>}
```
cjones-cs: This is cool. While a label is our current use case, I think changing this to accept a component would be useful down the road. That way you could add things like helpful tooltips or that sort of thing.

It would also allow the logic of controlling the color here to be inside the parent component instead of in here since this is geared specifically toward the number counting case. Starts with a dash is very specific for a generic component

frontend/src/routes/admin/challenges/SidebarTabs/ChallengeFeedbackTab.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    // Filter only entries with a valid number for quality
+    const qualityEntries = feedback.filter(
+      (entry) => typeof entry.feedback_data?.quality === 'number',
+    );
+    const totalQuality = qualityEntries.reduce(
+      (sum, entry) => sum + (entry.feedback_data.quality as number),
+      0,
+    );
+    const averageQuality = qualityEntries.length > 0 ? totalQuality / qualityEntries.length : 0;
```
cjones-cs, line 81: While doing these calculations on the frontend is acceptable, wouldn't it normally be better to do it on the backend and send it alongside the feedback rows?

frontend/src/routes/events/challenge/FeedbackModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+const items = Array(10).fill(null).map((_, i) => (
+  // eslint-disable-next-line react/no-array-index-key
+  <SegmentedControl.Item key={String(i)} value={String(i + 1)}>{i + 1}</SegmentedControl.Item>
+));
+
+/**
+ * Character limit for textarea fields.
+ * Should be safely below content length restriction assuming all text fields are at max.
+ */
+const CHAR_LIMIT = 2000;
```
cjones-cs: While we can support this much, I wonder if this is too high to get meaningful content. Most of the feedback fields I've used are generally closer to 200-500 char.

cjones-cs (changes requested):

frontend/src/routes/events/challenge/FeedbackModal.tsx
trevnels-cs: I think 500 is reasonable, I just want to make sure there's enough room there that people don't feel limited

frontend/src/routes/admin/challenges/SidebarTabs/ChallengeFeedbackTab.tsx
trevnels-cs: Since feedback doesn't have a specific schema on the backend for flexibility, the backend doesn't know what fields are there so can't really perform those calculations. The fields that are present here are defined solely by the frontend.

frontend/src/routes/events/challenge/FeedbackModal.tsx
```diff
@@ -0,0 +1,148 @@
...
+                  </SegmentedControl.Root>
+                )}
+              />
+            )}
+          </FormField>
+
+          <FormField
+            label="What did you like about this challenge?"
+            rightComponent={(
+              <Text size="2" color="gray" className="[label[data-invalid=true]+&]:!text-(--red-11)">
```
cjones-cs, line 107: Should we use COLOR_NEGATIVE instead of this red, so it exactly matches the theming of the error messages?

trevnels-cs: The tailwind compiler won't know about the values of those variables, so it can't be used in a class string like this. Invalid labels are also defined specifically as `--red-11` in css, so this is consistent with that styling

cjones-cs (approved):

PR #542, opened 2025-12-10
Display error for non-modal challenge provisioning

trevnels-cs:
Previously, errors during challenge start were only displayed on the challenge switch modal, not when starting from a fresh disconnected workspace:

image

### DIFF
--- a/frontend/src/routes/events/challenge/ChallengeSidebar.tsx
+++ b/frontend/src/routes/events/challenge/ChallengeSidebar.tsx
@@ -16,6 +16,7 @@ import ChallengeIcon from 'components/ChallengeIcon';
 import RadixMarkdown from 'components/RadixMarkdown';
 import Timer from 'components/Timer';
 import { groupBy } from 'lodash';
+import { useState } from 'react';
 import { TbArrowLeft } from 'react-icons/tb';
 import { Link, useParams } from 'react-router';
 import { mutate } from 'swr';
@@ -34,6 +35,8 @@ export default function ChallengeSidebar() {
     Number(idEvent),
     Number(idChallenge),
   );
+
+  const [ provisioningError, setProvisioningError ] = useState<Error | undefined>();
 
   const {
     challenge, questions, hints, attempts,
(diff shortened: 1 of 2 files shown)


frontend/src/routes/events/challenge/ConnectModal.tsx
```diff
@@ -57,7 +64,16 @@ export default function ConnectModal({ eventId, challengeId, isTeam }: {
   if (currentChallenge === null) {
     // If the workspace isn't connected to anything, don't require confirmation
     return (
-      <Button onClick={handleConnect} loading={loading || isValidating} color={COLOR_POSITIVE} className="pulsate">
+      <Button
+        // For starting without modal, pass errors to the parent for display
+        onClick={
+            () => handleConnect()
+              .catch(onError)
```
cjones-cs, line 71: Why did you decide to put the catch down here instead of inside the handleConnect function? In other examples, we put them together so all the logic is in one place

frontend/src/routes/events/challenge/ChallengeSidebar.tsx
```diff
@@ -99,13 +102,21 @@ export default function ChallengeSidebar() {
...
+                {provisioningError && <ErrorCallout className="mt-3">{provisioningError.message}</ErrorCallout>}
+                <Flex direction="row" gap="2" mt="3" align="center" justify="between">
+                  <ConnectModal
+                    eventId={event.id}
+                    challengeId={challenge.id}
+                    isTeam={event.max_team_size > 1}
+                    onError={setProvisioningError}
+                  />
+                  <Box flexShrink="0">
+                    {hints && hints.length > 0 && <HintsModal eventId={Number(idEvent)} challengeId={Number(idChallenge)} />}
```
cjones-cs: Should we be using this here or doing event.id and challenge.id like you pass to the ConnectModal?

frontend/src/routes/events/challenge/ConnectModal.tsx
```diff
@@ -18,16 +21,20 @@ export default function ConnectModal({ eventId, challengeId, isTeam }: {
...
+    onError?.(undefined);
     return connectWorkspace(eventId, challengeId).finally(() => {
       setLoading(false);
     });
   };
 
-  const handleReset = async () => recycleChallengeContainers(eventId, challengeId);
+  const handleReset = async () => {
+    onError?.(undefined);
+    return recycleChallengeContainers(eventId, challengeId);
```
cjones-cs, line 32: Is there the possibility that the reset fails and needs to call onError? We clear onError here, but only allow setting it in the handleConnect function

cjones-cs (changes requested):

frontend/src/routes/events/challenge/ConnectModal.tsx
trevnels-cs: When using the modal, it does the error handling - the non-modal case is the only time the error should be sent to the underlying page instead. If this was done in handleConnect, errors in the modal case would show up twice. onError is only for this one special case, and is replicating what the modal logic does.

frontend/src/routes/events/challenge/ConnectModal.tsx
trevnels-cs: Reset is handled through a modal, which does the error handling. onError is only for inline operations that don't have a modal to display their error.

cjones-cs (approved):

PR #543, opened 2025-12-10
Make challenge env creation atomic

RayofLightz:
Makes challenge env creation atomic

### DIFF
--- a/backend/ng/containers/controllers/start_containers.py
+++ b/backend/ng/containers/controllers/start_containers.py
@@ -1,7 +1,10 @@
+from CTFd.models import db
+from CTFd.utils import get_app_config
 from ...team.models.Team import Team
 from ...user.models.User import User
 from ..models.ContainerInstance import ContainerInstance
 from ..models.IndvidualContainer import IndvidualContainer
+from ..utils.get_client import get_client
 from ...challenge.models.ContainerBlueprint import ContainerBlueprint
 from ..constants import CHALLENGER_NET_NAME
 from ...core import BusinessLogicError
@@ -12,13 +15,31 @@ def start_containers(challenge_id: int, team_id: int, current_user: User) -> boo
 
     ctrs = []
     networks = []
-    for blueprint in blueprints:
-        ctrs.append(ContainerInstance.create_container_instance(blueprint.id, team))
-        if blueprint.networks:
-            networks.extend(blueprint.networks)
+    try:
+        for blueprint in blueprints:
+            ctrs.append(ContainerInstance.create_container_instance(blueprint.id, team, commit=False))
+            if blueprint.networks:
+                networks.extend(blueprint.networks)
+
+    except Exception as err:
+        for ctr in ctrs:
+            ctr.remove()
+        print(networks)
+        for net in networks:
+            DOCKER_HOST = get_app_config("DOCKER_HOST")
+            client = get_client(DOCKER_HOST)
+            net_obj = client.get_network_by_name(ContainerInstance.render_network_name(team_id, net, challenge_id))
+            if net_obj:
+                net_obj.remove()
+
+        db.session.rollback()
+        raise BusinessLogicError(f"Challenge failed to start, please contact support {str(err)}") from err
+
+    db.session.commit()
 
     indvidual_ctr = IndvidualContainer.create_indvidual_container(current_user.id)
     indvidual_ctr.disconnect_from_networks()
+
 
     if CHALLENGER_NET_NAME not in set(networks):
         raise BusinessLogicError("Challenge has no challenger network")
--- a/backend/ng/containers/controllers/start_containers.py
+++ b/backend/ng/containers/controllers/start_containers.py
@@ -17,23 +17,24 @@ def start_containers(challenge_id: int, team_id: int, current_user: User) -> boo
     networks = []
     try:
         for blueprint in blueprints:
-            ctrs.append(ContainerInstance.create_container_instance(blueprint.id, team, commit=False))
+            ## Networks should get appened first to ensure they get properly
+            ## cleaned up
             if blueprint.networks:
                 networks.extend(blueprint.networks)
+            ctrs.append(ContainerInstance.create_container_instance(blueprint.id, team, commit=False))
 
     except Exception as err:
+        DOCKER_HOST = get_app_config("DOCKER_HOST")
+        client = get_client(DOCKER_HOST)
         for ctr in ctrs:
             ctr.remove()
-        print(networks)
-        for net in networks:
-            DOCKER_HOST = get_app_config("DOCKER_HOST")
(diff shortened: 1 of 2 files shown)


backend/ng/containers/controllers/start_containers.py
```diff
@@ -12,14 +15,32 @@ def start_containers(challenge_id: int, team_id: int, current_user: User) -> boo
...
+            if blueprint.networks:
+                networks.extend(blueprint.networks)
+
+    except Exception as err:
+        for ctr in ctrs:
+            ctr.remove()
+        print(networks)
+        for net in networks:
+            DOCKER_HOST = get_app_config("DOCKER_HOST")
+            client = get_client(DOCKER_HOST)
```
trevnels-cs: This could be pulled out of the loop to use one client for cleanup

backend/ng/containers/controllers/start_containers.py
```diff
@@ -12,14 +15,32 @@ def start_containers(challenge_id: int, team_id: int, current_user: User) -> boo
...
+    try:
+        for blueprint in blueprints:
+            ctrs.append(ContainerInstance.create_container_instance(blueprint.id, team, commit=False))
+            if blueprint.networks:
+                networks.extend(blueprint.networks)
+
+    except Exception as err:
+        for ctr in ctrs:
+            ctr.remove()
+        print(networks)
```
trevnels-cs: I think this print can be removed

backend/ng/containers/controllers/start_containers.py
```diff
@@ -12,14 +15,32 @@ def start_containers(challenge_id: int, team_id: int, current_user: User) -> boo
...
     networks = []
-    for blueprint in blueprints:
-        ctrs.append(ContainerInstance.create_container_instance(blueprint.id, team))
-        if blueprint.networks:
-            networks.extend(blueprint.networks)
+    try:
+        for blueprint in blueprints:
+            ctrs.append(ContainerInstance.create_container_instance(blueprint.id, team, commit=False))
+            if blueprint.networks:
+                networks.extend(blueprint.networks)
```
trevnels-cs, line 23: If the container is started but an exception is raised afterwards while still inside create_container_instance (i.e. when connecting it to a network), then it won't be added to `ctrs` and won't be cleaned up.

This may need to be adjusted so that create_container_instance just creates the container and returns the instance, then connect networks is called afterwards on the returned instance once it's in the ctrs list

When I manually raise an exception in `connect_networks` for a specific container, the container being created at the time the exception is thrown won't be removed.

backend/ng/containers/controllers/start_containers.py
```diff
@@ -12,14 +15,32 @@ def start_containers(challenge_id: int, team_id: int, current_user: User) -> boo
...
+        print(networks)
+        for net in networks:
+            DOCKER_HOST = get_app_config("DOCKER_HOST")
+            client = get_client(DOCKER_HOST)
+            net_obj = client.get_network_by_name(ContainerInstance.render_network_name(team_id, net, challenge_id))
+            if net_obj:
+                net_obj.remove()
+
+        db.session.rollback()
+        raise BusinessLogicError(f"Challenge failed to start, please contact support {str(err)}") from err
```
trevnels-cs: I would recommend adding a : after support or wrap the `str(err)` message in parens in the output string for separation

backend/ng/containers/controllers/start_containers.py
```diff
@@ -12,14 +15,32 @@ def start_containers(challenge_id: int, team_id: int, current_user: User) -> boo
...
+    except Exception as err:
+        for ctr in ctrs:
+            ctr.remove()
+        print(networks)
+        for net in networks:
+            DOCKER_HOST = get_app_config("DOCKER_HOST")
+            client = get_client(DOCKER_HOST)
+            net_obj = client.get_network_by_name(ContainerInstance.render_network_name(team_id, net, challenge_id))
+            if net_obj:
+                net_obj.remove()
```
trevnels-cs: Getting this during cleanup after running the start endpoint multiple times with failures
```
File "/opt/CTFd/CTFd/plugins/ng/containers/controllers/start_containers.py", line 33, in start_containers
    net_obj.remove()
  File "/opt/venv/lib/python3.12/site-packages/docker/models/networks.py", line 91, in remove
    return self.client.api.remove_network(self.id)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/docker/api/network.py", line 186, in remove_network
    self._raise_for_status(res)
  File "/opt/venv/lib/python3.12/site-packages/docker/api/client.py", line 277, in _raise_for_status
    raise create_api_error_from_http_exception(e) from e
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/docker/errors.py", line 39, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation) from e
docker.errors.NotFound: 404 Client Error for https://172.17.0.1:2376/v1.52/networks/jg8nbyyh28w2oomprc3fc0ypv: Not Found ("network jg8nbyyh28w2oomprc3fc0ypv not found")
```
remove may need a try/catch

backend/ng/containers/controllers/start_containers.py
RayofLightz: This is a good point and I am glad you caught it as it showed I need to catch a few other errors in create_container_instance method as well. This should be fixed now

backend/ng/containers/controllers/start_containers.py
RayofLightz: Good catch

backend/ng/containers/controllers/start_containers.py
RayofLightz: How are you raising the error in the code that you get this error? The get_network_by_name should return None if there is no network

RayofLightz: TBH this sounds like a race condition in the docker daemon. If you are intentionally throwing an error it might be happening to fast and send back stale data

RayofLightz: Yeah this is an issue in the daemon not the sdk

trevnels-cs: I think this is a race condition - the network does get removed, but the exception is still thrown. Might be wise to just catch it and continue iterating instead of bailing out of the loop early and potentially not removing other networks.

trevnels-cs (approved):

PR #557, opened 2025-12-11
Connect YAML editor to challenge lint functionality

trevnels-cs:
image

image

Challenge upload/update YAML editors now lint (throttled to at most once per second) as the user types or if a file is uploaded. Form submission is blocked if the lint has errors.

Note that valid challenges will lint as invalid currently - @johnmarsden-edu is working on a fix for that.

### DIFF
--- a/backend/ng/challenge/controllers/admin/lint_challenge.py
+++ b/backend/ng/challenge/controllers/admin/lint_challenge.py
@@ -139,7 +139,7 @@ def lint_challenge(yaml_content: str) -> SerializedLint | None:
         compose_file = parse_compose_string(yaml_content)
         warnings = compose_file.warnings()
 
-        if warnings is not None and (warnings.self_warnings or warnings.field_warnings):
+        if warnings is not None:
             return {
                 "warnings": serialize_warnings(warnings),
             }
--- a/backend/ng/challenge/controllers/admin/lint_challenge.py
+++ b/backend/ng/challenge/controllers/admin/lint_challenge.py
@@ -139,7 +139,7 @@ def lint_challenge(yaml_content: str) -> SerializedLint | None:
         compose_file = parse_compose_string(yaml_content)
         warnings = compose_file.warnings()
 
-        if warnings.self_warnings or warnings.field_warnings:
+        if warnings is not None and (warnings.self_warnings or warnings.field_warnings):
             return {
                 "warnings": serialize_warnings(warnings),
             }
--- a/frontend/src/components/ChallengeForm.tsx
+++ b/frontend/src/components/ChallengeForm.tsx
@@ -1,11 +1,6 @@
 import { lintChallenge } from '@/hooks/challenge';
 import type { LintResult } from '@/types';
-import {
-  Box,
-  Flex,
-  Strong,
-  Text,
-} from '@radix-ui/themes';
+import { Flex, Strong, Text } from '@radix-ui/themes';
 import { throttle } from 'lodash';
 import { useEffect, useRef, useState } from 'react';
 import { Controller, type UseFormReturn } from 'react-hook-form';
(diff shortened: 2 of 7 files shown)


cjones-cs (approved):
This looks good, but I would wait until John gets ready with his change to merge these together

trevnels-cs:
John gave the go-ahead for me to add that fix directly to this branch, fix is now in place. Also adjusted the formatting for the lint messages, since it was previously creating divs inside of the callout `<p>`, which isn't valid HTML.

frontend/src/components/ChallengeForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useEffect(() => {
+    trigger('yaml');
+  }, [ lintResult, trigger ]);
+
+  return (
+    <>
+      <Controller
+        control={control}
+        rules={{
+          validate : () => !(lintResult && 'errors' in lintResult),
```
johnmarsden-edu: You use this a lot of times it might be worth breaking it it out into a variable.

backend/ng/challenge/controllers/admin/lint_challenge.py
```diff
@@ -139,11 +139,11 @@ def lint_challenge(yaml_content: str) -> SerializedLint | None:
         compose_file = parse_compose_string(yaml_content)
         warnings = compose_file.warnings()
 
-        if warnings.self_warnings or warnings.field_warnings:
+        if warnings is not None and (warnings.self_warnings or warnings.field_warnings):
```
johnmarsden-edu: I need to double check but you might actually be able to get rid of the check after the `and` because I think `warnings()` should only return none if neither of them are true

frontend/src/components/ChallengeForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <YamlEditor
+          value={field.value}
+          onChange={(v) => { field.onChange(v); lintRef.current(v); }}
+          ref={field.ref}
+        />
+      )
+    }
+      />
+      {lintResult && ('errors' in lintResult || 'warnings' in lintResult) && (
+        <Flex direction="column" gap="3">
```
johnmarsden-edu, line 54: Did I need to make any changes to also have these show up as little squiggles?

johnmarsden-edu (approved):

frontend/src/components/ChallengeForm.tsx
trevnels-cs: Squiggles will take some integration work with CodeMirror (namely an editor plugin that walks the AST and maps lint message paths to character ranges.) For now, displaying the messages as-is is the simplest approach, but having the paths relative to the root of the document is sufficient information to be able to implement it in the future.

The main thing I see that would cause issues is that warning paths don't correspond to a path in the document. i.e. for external networks the path is `.ComposeFile.network warnings.competitor_net` when it should be `.networks.competitor_net`

johnmarsden-edu: Okay I will get to that when I can as it was on my list of things to do anyway

backend/ng/challenge/controllers/admin/lint_challenge.py
trevnels-cs: That's right, the `warnings.self_warnings or warnings.field_warnings` check also happens inside `warnings()`

frontend/src/components/ChallengeForm.tsx
trevnels-cs: When I have the cycles, I can handle the CodeMirror parts since that's pretty frontend-heavy and I've done the exact same thing before (though with JSON instead of YAML.) The main thing on the parser side is just making sure the returned paths always correspond to the document tree

PR #560, opened 2025-12-12
Adds blocked domains to email validation

### LINKED ISSUE #553: Add a "blocked_domains" field for event registration
There are subdomains of "allowed_domains" that need to be disallowed from registering for events. We need to update the domain verification logic to reject any domains that match anything in `blocked_domains`. 

e.g.
Allow:
`.gov`

Block:
`associates.cisa.dhs.gov`
`la.gov`
`fresnocountyca.gov`

**Note -** 
Make sure you use `blocked_domains` as the field name. The DB tables have already been updated in staging and prod to include this new column.

droge91:
Adds a blocked domain check to email validation + makes the check a bit cleaner
closes #553

### DIFF
--- a/backend/ng/event/models/Event.py
+++ b/backend/ng/event/models/Event.py
@@ -35,6 +35,7 @@ class Event(db.Model):
     hints_enabled = db.Column(db.Boolean, default=False, nullable=False)
     time_limit_minutes = db.Column(db.Integer, nullable=True)
     allowed_domains: Mapped[list[str]] = db.Column(JSON, nullable=False, default=[])
+    blocked_domains: Mapped[list[str]] = db.Column(JSON, nullable=False, default=[])
     show_leaderboard = db.Column(db.Boolean, default=False, nullable=False)
 
     __table_args__ = (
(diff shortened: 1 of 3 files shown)


backend/ng/event/models/Event.py
```diff
@@ -424,10 +434,12 @@ def check_eligibility(self, user):
         if self.allowed_domains:
             user_email_domain = user.ctfd_user.email.split("@")[-1].lower()
             allowed_domains_lower = [domain.lower() for domain in self.allowed_domains]
-            for domain in allowed_domains_lower:
-                if domain in user_email_domain:
-                    break
-            else:
+            blocked_domains_lower = [domain.lower() for domain in self.blocked_domains]
+
+            if any(user_email_domain == domain or user_email_domain.endswith("." + domain) for domain in blocked_domains_lower):
```
fpayer, line 439: For the two domain lists, can you verify that there isn't already a preceeding `.` before you prepend one?

fpayer (approved):

PR #563, opened 2025-12-12
Container recycle updates

RayofLightz:
* Adds full recycle for admins
* Better accounts for multiple team members being on comp net
* Reorg some of the start code to be more reusable
* Add a redis lock so people can't start/recycle a challenge when it is already being started/recycled

### DIFF
--- a/backend/ng/containers/controllers/recycle_containers.py
+++ b/backend/ng/containers/controllers/recycle_containers.py
@@ -4,7 +4,7 @@ from ..constants import CHALLENGER_NET_NAME
 from ...core import BusinessLogicError
 
 def recycle_containers(challenge_id: int, team_id: int) -> bool:
-    disconnected_indvidual_containers = ContainerInstance.recycle_instance_group(challenge_id, team_id)
+    disconnected_indvidual_containers = ContainerInstance.remove_instance_group(challenge_id, team_id)
 
     networks = ContainerInstance.start_instance_group(challenge_id, team_id)
 
--- a/backend/ng/containers/controllers/recycle_containers.py
+++ b/backend/ng/containers/controllers/recycle_containers.py
@@ -3,24 +3,21 @@ from ..models.IndvidualContainer import IndvidualContainer
 from ..constants import CHALLENGER_NET_NAME
 from ...core import BusinessLogicError
 
-def recycle_containers(challenge_id: int, team_id: int, current_user: int) -> bool:
-    indv_ctr = IndvidualContainer.get_user_indvidual_container(current_user.id)
-    indv_ctr.disconnect_from_networks()
-
-    indv_ctrs = ContainerInstance.recycle_instance_group(challenge_id, team_id)
+def recycle_containers(challenge_id: int, team_id: int) -> bool:
+    disconnected_indvidual_containers = ContainerInstance.recycle_instance_group(challenge_id, team_id)
 
     networks = ContainerInstance.start_instance_group(challenge_id, team_id)
 
     if CHALLENGER_NET_NAME not in set(networks):
-        raise BusinessLogicError("Challenge has no challenger network")
+        raise BusinessLogicError(f"Challenge has no {CHALLENGER_NET_NAME} network")
 
 
-    indv_ctr.connect_to_network(ContainerInstance.render_network_name(team_id, CHALLENGER_NET_NAME, challenge_id))
+    challenger_net = ContainerInstance.render_network_name(team_id, CHALLENGER_NET_NAME, challenge_id)
 
-    for ctr in indv_ctrs:
+    for ctr in disconnected_indvidual_containers:
         indv_ctr_obj = IndvidualContainer.get_indvidual_container_by_dockerid(ctr)
         if indv_ctr_obj:
             indv_ctr_obj.disconnect_from_networks()
-            indv_ctr_obj.connect_to_network(ContainerInstance.render_network_name(team_id, CHALLENGER_NET_NAME, challenge_id))
+            indv_ctr_obj.connect_to_network(challenger_net)
 
     return True
--- a/backend/ng/containers/controllers/recycle_containers.py
+++ b/backend/ng/containers/controllers/recycle_containers.py
@@ -1,29 +1,26 @@
-from CTFd.utils import get_app_config
-from .start_containers import start_containers
-from ..utils.get_client import get_client
 from ..models.ContainerInstance import ContainerInstance
 from ..models.IndvidualContainer import IndvidualContainer
-from ...challenge.models.ContainerBlueprint import ContainerBlueprint
+from ..constants import CHALLENGER_NET_NAME
+from ...core import BusinessLogicError
 
 def recycle_containers(challenge_id: int, team_id: int, current_user: int) -> bool:
     indv_ctr = IndvidualContainer.get_user_indvidual_container(current_user.id)
     indv_ctr.disconnect_from_networks()
 
-    instances = ContainerInstance.get_instance_group(challenge_id, team_id)
+    indv_ctrs = ContainerInstance.recycle_instance_group(challenge_id, team_id)
 
-    DOCKER_HOST = get_app_config("DOCKER_HOST")
-    client = get_client(DOCKER_HOST)
-    for instance in instances:
-        instance.remove()
+    networks = ContainerInstance.start_instance_group(challenge_id, team_id)
 
-    blueprints = ContainerBlueprint.get_for_challenge(challenge_id)
-    for blueprint in blueprints:
-        for net in blueprint.networks:
-            net_obj = client.get_network_by_name(ContainerInstance.render_network_name(team_id, net, challenge_id))
-            if net_obj:
-                try:
-                    net_obj.remove()
-                except Exception:
-                    pass
+    if CHALLENGER_NET_NAME not in set(networks):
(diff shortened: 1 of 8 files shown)


backend/ng/containers/controllers/recycle_containers.py
```diff
@@ -1,29 +1,26 @@
...
-from ...challenge.models.ContainerBlueprint import ContainerBlueprint
+from ..constants import CHALLENGER_NET_NAME
+from ...core import BusinessLogicError
 
 def recycle_containers(challenge_id: int, team_id: int, current_user: int) -> bool:
     indv_ctr = IndvidualContainer.get_user_indvidual_container(current_user.id)
     indv_ctr.disconnect_from_networks()
 
-    instances = ContainerInstance.get_instance_group(challenge_id, team_id)
+    indv_ctrs = ContainerInstance.recycle_instance_group(challenge_id, team_id)
```
johnmarsden-edu: Nit: name is not descriptive. Maybe deployment containers, challenge containers, or something of that nature.

backend/ng/containers/controllers/recycle_containers.py
```diff
@@ -1,29 +1,26 @@
...
-    for blueprint in blueprints:
-        for net in blueprint.networks:
-            net_obj = client.get_network_by_name(ContainerInstance.render_network_name(team_id, net, challenge_id))
-            if net_obj:
-                try:
-                    net_obj.remove()
-                except Exception:
-                    pass
+    if CHALLENGER_NET_NAME not in set(networks):
+        raise BusinessLogicError("Challenge has no challenger network")
```
johnmarsden-edu: This should potentially expose the expected name of the network

backend/ng/containers/controllers/recycle_containers.py
```diff
@@ -1,29 +1,26 @@
...
 
-    return start_containers(challenge_id, team_id, current_user)
+
+    indv_ctr.connect_to_network(ContainerInstance.render_network_name(team_id, CHALLENGER_NET_NAME, challenge_id))
+
+    for ctr in indv_ctrs:
+        indv_ctr_obj = IndvidualContainer.get_indvidual_container_by_dockerid(ctr)
```
johnmarsden-edu, line 18: Nit: misspelled individual. After looking it seems like we have consistently cut out an i, is there a reason for that?

backend/ng/containers/controllers/recycle_containers.py
```diff
@@ -1,29 +1,26 @@
...
+        if indv_ctr_obj:
+            indv_ctr_obj.disconnect_from_networks()
+            indv_ctr_obj.connect_to_network(ContainerInstance.render_network_name(team_id, CHALLENGER_NET_NAME, challenge_id))
```
johnmarsden-edu: This should probably be rendered ahead of time and used repeatedly rather than rendered repeatedly

backend/ng/containers/models/ContainerInstance.py
```diff
@@ -273,6 +280,90 @@ def get_instance_group(cls, challenge_id: int, team_id: int):
...
+            DOCKER_HOST = get_app_config("DOCKER_HOST")
+            client = get_client(DOCKER_HOST)
+            for instance in instances:
+                instance.remove()
+
+            blueprints = ContainerBlueprint.get_for_challenge(challenge_id)
+            indv_ctrs = []
+            for blueprint in blueprints:
+                for net in blueprint.networks:
+                    net_obj = client.get_network_by_name(cls.render_network_name(team_id, net, challenge_id))
```
johnmarsden-edu, line 305: Same note here, don't need to render this repeatedly

backend/ng/containers/models/ContainerInstance.py
```diff
@@ -273,6 +280,90 @@ def get_instance_group(cls, challenge_id: int, team_id: int):
     def get_instance_by_id(cls, instance_id: int):
         return cls.query.filter_by(id=instance_id).first()
 
+    @classmethod
+    def recycle_instance_group(cls, challenge_id: int, team_id: int):
```
johnmarsden-edu: Should specify the type being returned

backend/ng/containers/models/ContainerInstance.py
```diff
@@ -273,6 +280,90 @@ def get_instance_group(cls, challenge_id: int, team_id: int):
...
+                            net_obj.remove()
+                        except Exception:
+                            pass
+            lock.release()
+            return indv_ctrs
+        else:
+            raise BusinessLogicError("Challenge is already being started/reset")
+
+    @classmethod
+    def start_instance_group(cls, challenge_id: int, team_id: int):
```
johnmarsden-edu: Specify return type

backend/ng/containers/models/ContainerInstance.py
```diff
@@ -273,6 +280,90 @@ def get_instance_group(cls, challenge_id: int, team_id: int):
...
+                        try:
+                            net_obj.remove()
+                        except Exception:
+                            pass
+                db.session.rollback()
+                lock.release()
+                raise BusinessLogicError(f"Challenge failed to start, please contact support: {str(err)}") from err
+            db.session.commit()
+            lock.release()
+            return networks
```
johnmarsden-edu, line 368: It feels odd to me that starting the instance group returns the networks. I could see returning the container group and the networks, but just the networks feels odd.

backend/ng/containers/models/IndvidualContainer.py
```diff
@@ -26,6 +26,10 @@ def __repr__(self):
     def get_user_indvidual_container(cls, user_id: int):
         return cls.query.filter_by(user=user_id).first()
 
+    @classmethod
+    def get_indvidual_container_by_dockerid(cls, docker_id: str):
```
johnmarsden-edu, line 30: Just a thought but at some point we might want to use a more builder oriented pattern for these because the names are getting to be a bit much 🤣

backend/ng/containers/routes/admin_routes.py
```diff
@@ -149,6 +152,35 @@ def post(self, container_instance, **kwargs):
...
+            "challenge_id": "Id of the challenge",
+            "team_id": "Id of the team to recycle",
+        },
+        responses={
+            200: "Success",
+            400: "Bad request"
+        },
+    )
+    @admin_endpoint()
+    def post(self, challenge_id, team_id, **kwargs):
```
johnmarsden-edu, line 167: This logic looks almost the same as what's in the recycle code above, is there a significant difference that justifies the duplication?

backend/ng/containers/controllers/recycle_containers.py
RayofLightz: The variable name or the method name?

backend/ng/containers/controllers/recycle_containers.py
RayofLightz: The reason is it just now got caught

backend/ng/containers/models/ContainerInstance.py
RayofLightz: Right now its because we really only care about the networks. As we need to connect in the individual vnc instances in

backend/ng/containers/controllers/recycle_containers.py
johnmarsden-edu: The variable name

backend/ng/containers/controllers/recycle_containers.py
RayofLightz: Oh shit. I can't properly fix this as it would cause a db change right now

backend/ng/containers/models/ContainerInstance.py
RayofLightz: It does need to be rendered each time here.

backend/ng/containers/routes/admin_routes.py
RayofLightz: Good catch. I think as I was deving on this at one point they were much more different, but now they can be merged.

backend/ng/containers/controllers/recycle_containers.py
johnmarsden-edu: That is really funny

johnmarsden-edu (approved):
Looks good to me

backend/ng/containers/models/ContainerInstance.py
```diff
@@ -273,6 +280,90 @@ def get_instance_group(cls, challenge_id: int, team_id: int):
...
+                    if net_obj:
+                        ## There is a potential race condition where the daemon has deleted the network
+                        ## But the network is not fully removed from NETDB
+                        try:
+                            net_obj.remove()
+                        except Exception:
+                            pass
+                db.session.rollback()
+                lock.release()
+                raise BusinessLogicError(f"Challenge failed to start, please contact support: {str(err)}") from err
```
trevnels-cs: Logging the `err` here could potentially leak information about the challenge environment to the user that would give them an unfair advantage (i.e. the existence of services on internal networks that should be hidden from the user as part of the challenge.)

It may be wise to give the user a clean error message and log the full stack trace only on the server.

backend/ng/containers/models/ContainerInstance.py
trevnels-cs: `remove_instance_group`/`cleanup_instance_group` may be a better name for this, since it removes the containers but doesn't recreate them

PR #564, opened 2025-12-12
Adds user-based rate limiting w/ Redis

fpayer:
- Support ticket creation/ new messages / attachment uploads limited at 1 per 5 seconds
- Container recycling limited at 1 per 5 minutes
- Answer submission limited at 1 per second

### DIFF
--- a/backend/ng/__init__.py
+++ b/backend/ng/__init__.py
@@ -1,41 +1,46 @@
 #!/usr/bin/env python3
 
 from typing import Any
-from flask_socketio import SocketIO
 
 from CTFd.models import db
-from .core.utils.logger import get_logger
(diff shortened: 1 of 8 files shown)


backend/ng/__init__.py
```diff
@@ -69,6 +74,10 @@ def load(app: Any) -> None:
         else:
             logger.warning("Redis notification system failed to initialize")
 
+        app.config["RATELIMIT_STORAGE_URI"] = app.config.get("REDIS_URL")
+
+        limiter.init_app(app)
```
RayofLightz, line 79: We might want to consider putting rate limit on its own redis db.

RayofLightz (approved):
Looks good just a point to consider

backend/ng/__init__.py
fpayer: I think we should be fine with a single redis db right now. If we run into issues, we can expand.

PR #565, opened 2025-12-12
Permission updates for feedback system

trevnels-cs:
**`CAN_VIEW_CHALLENGES` is now granted even after team end time**
Continue showing challenges after time is up to allow users to review their submission history and share feedback. Also makes time expiring while looking at a challenge more graceful - you no longer get a screen full of permission errors when time is up, since you can continue to get challenge data at that point (but can no longer submit anything.) Users will just see submit buttons disappear and answer fields lock. Also changes history/feedback UI to be available without CAN_PLAY_CHALLENGES permission, since they aren't tied to playing challenges. Hints may still be viewed, but cannot be redeemed in this state.

**Permission changes for feedback**
1. Users must now be registered for an event to get/set feedback. Users shouldn't be allowed to add feedback to things that they didn't participate in.
2. `CAN_VIEW_CHALLENGES` permission is required to submit feedback (user must have at least started the event.) Feedback shouldn't be allowed without having seen the challenges.
3. Event feedback can only be submitted after the team's end time as per requirements. (We do permit challenge feedback being submitted during the game though, but don't explicitly direct users towards it.) For events without a time bound, this restriction does not apply.

### DIFF
--- a/backend/ng/feedback/routes/user_routes.py
+++ b/backend/ng/feedback/routes/user_routes.py
@@ -2,7 +2,6 @@
 User Routes for Feedback Operations
 """
 
-from datetime import datetime
 from flask_restx import Namespace, Resource
 
 
--- a/backend/ng/feedback/routes/user_routes.py
+++ b/backend/ng/feedback/routes/user_routes.py
@@ -2,17 +2,24 @@
 User Routes for Feedback Operations
 """
 
+from datetime import datetime
 from flask_restx import Namespace, Resource
+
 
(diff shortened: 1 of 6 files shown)


droge91 (commented):
The Permission changes look fine. But it looks like a bunch of the feedback tests are broken now, I believe the load_team_by_user_and_event being present now means the tests need to register the user before creating a feedback object

fpayer (approved):

PR #570, opened 2025-12-15
Presigned url caching

zoecough:


RayofLightz (approved):

zoecough:
Going to include redis util here as well instead of merging it in using the notification service

backend/ng/core/services/s3_service.py
```diff
@@ -89,6 +142,11 @@ def generate_download_url(self, object_key: str, expires_in: int = 3600) -> str:
         if not self.is_configured():
             raise Exception("S3 service not configured")
 
+        cached_url = self._get_cached_url('download', object_key, expires_in)
```
fpayer, line 145: Might be good to have the default value defined as a const at the top of the file.

fpayer (approved):

PR #571, opened 2025-12-15
add admin vnc

RayofLightz:
@trevnels-cs if you could take a look and let me know how to make the sizing for the admin VNC modal not super small I would really appreciate that.


trevnels-cs:
It seems like setting the Modal className to `!max-w-[75vw]` and the iframe className to `w-full h-[75vh]` produces a pretty good result on my end. I would also recommend adding `&resize=scale` to the NoVNC params to fit the screen to the iframe by default. This will scale on the client side, as opposed to the resize being used on the challenge page, which adjusts the actual resolution of the machine.

zoecough (approved):
lgtm

Issue #200, opened 2025-09-26
Display the team name on the join team invite modal when someone is joining a team


cjones-cs:
Display the team name on the join team invite modal when someone is joining a team. This will need a GET route to map an invite code to the corresponding team data.

PR #572, opened 2025-12-15
Feature/team name code

cjones-cs:
https://github.com/CyberSkyline/ctf-ng/issues/200

Watches the join code field in the form for changes, then fires a network call for the team name.

Watches the invite code from the url and gets the team name that way,

Changes the redirection on close from the /events to the /events/eventId page to match the current location of the modal button.

### DIFF
--- a/frontend/src/hooks/events.ts
+++ b/frontend/src/hooks/events.ts
@@ -92,6 +92,12 @@ export function registerMyEventTeamJoin(eventId: number, inviteCode: string) {
   });
 }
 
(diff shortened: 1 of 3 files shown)


trevnels-cs (approved):
Using `throttle` from lodash to set debouncedCode may result in cleaner rate-limiting behavior than managing setTimeout, but the current implementation is good as-is.

I did discover #576 while testing, but that's an unrelated backend issue that doesn't need to be solved here.

PR #575, opened 2025-12-15
Event feedback functionality

trevnels-cs:
- Feedback tab on event pages containing feedback form
- Prompt for feedback in event header if not submitted

### DIFF
--- a/frontend/src/components/Callouts.tsx
+++ b/frontend/src/components/Callouts.tsx
@@ -1,7 +1,17 @@
+import {
+  COLOR_INFO,
+  COLOR_NEGATIVE,
+  COLOR_POSITIVE,
+  COLOR_WARNING,
+} from '@/constants';
 import { Callout } from '@radix-ui/themes';
 import type { ReactNode } from 'react';
-import { TbAlertTriangle, TbCancel, TbInfoCircle } from 'react-icons/tb';
-import { COLOR_INFO, COLOR_NEGATIVE, COLOR_WARNING } from '@/constants';
+import {
+  TbAlertTriangle,
+  TbCancel,
+  TbCheck,
+  TbInfoCircle,
+} from 'react-icons/tb';
 
 export function ErrorCallout({ children, className }: {children: ReactNode, className?: string}) {
   return (
@@ -28,11 +38,25 @@ export function WarningCallout({ children, className }: {children: ReactNode, cl
     </Callout.Root>
   );
 }
+
 export function InfoCallout({ children, className }: {children: ReactNode, className?: string}) {
   return (
     <Callout.Root variant="surface" color={COLOR_INFO} className={className}>
       <Callout.Icon>
         <TbInfoCircle aria-label="Info" />
+      </Callout.Icon>
+      <Callout.Text className="whitespace-pre-wrap">
+        {children}
+      </Callout.Text>
+    </Callout.Root>
+  );
+}
+
+export function SuccessCallout({ children, className }: {children: ReactNode, className?: string}) {
+  return (
+    <Callout.Root variant="surface" color={COLOR_POSITIVE} className={className}>
+      <Callout.Icon>
+        <TbCheck aria-label="Success" />
       </Callout.Icon>
       <Callout.Text className="whitespace-pre-wrap">
         {children}
--- a/frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
+++ b/frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
@@ -6,6 +6,7 @@ import {
   Flex,
   RadioCards,
   Strong,
+  Text,
   TextArea,
 } from '@radix-ui/themes';
 import { ErrorCallout, InfoCallout, SuccessCallout } from 'components/Callouts';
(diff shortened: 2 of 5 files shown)


frontend/src/components/Callouts.tsx
```diff
@@ -40,3 +51,16 @@ export function InfoCallout({ children, className }: {children: ReactNode, class
     </Callout.Root>
   );
 }
+
+export function SuccessCallout({ children, className }: {children: ReactNode, className?: string}) {
```
cjones-cs, line 55: I think it would be a good future feature for this to have a timeout and disappear. That way you wouldn't get confused if this ends up occurring on button press 1 or 2 in an inline form

frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+    try {
+      await submitEventFeedback(eventId, data);
+      setButtonState('success');
+    } catch (err) {
+      setError((err as Error).message);
+    } finally {
+      setTimeout(() => {
+        setButtonState(null);
+      }, 2000);
+    }
```
cjones-cs, line 86: Doesn't swr already handle doing the await internally?

frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+            {(injected) => (
+              <Controller
+                name="role"
+                control={control}
+                render={({ field }) => (
+                  <RadioCards.Root
+                    value={field.value || null}
+                    onValueChange={(e) => {
+                      field.onChange(e);
+                    }}
```
cjones-cs: You could remove the extra {} here

frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+                        {role}
+                      </RadioCards.Item>
+                    ))}
+
+                    <TextArea
+                      value={field.value && ROLES.includes(field.value) ? '' : field.value || ''}
+                      onChange={field.onChange}
+                      placeholder="Other..."
+                      className="!min-h-min"
+                      rows={1}
```
cjones-cs, line 141: Since you're only allowing a single row, wouldn't using a TextField component be better?

frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+                    </RadioCards.Item>
+
+                    <RadioCards.Item value="Master's Degree">
+                      Master&apos;s Degree
+                    </RadioCards.Item>
+
+                    <RadioCards.Item value="PhD">
+                      PhD
+                    </RadioCards.Item>
+                  </RadioCards.Root>
```
cjones-cs, line 171: I would use the same technique as the ROLES and put education levels in an array. I'm also wondering if we need to include below a high school diploma? Unlikely, but might still be possible.

frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+              />
+            )}
+          </FormField>
+
+          <FormField
+            label="How can we improve the next President’s Cup? Please provide any other feedback you would like to share."
+            error={errors.additionalFeedback}
+          >
+            {(injected) => (
+              <TextArea
```
cjones-cs, line 299: For these TextAreas are we going to include the char counter?

cjones-cs (changes requested):

frontend/src/components/Callouts.tsx
trevnels-cs: The message used for this is intended to not time out, since it includes info about updating previous responses that should be shown if any submission has occurred previously

frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
trevnels-cs: This happens after SWR is done, it's an extra timeout to ensure the success message is shown, then goes away after 2s. Also serves as a debounce to prevent spamming the button

frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
trevnels-cs: These options were specified by CISA, so that would be up to them

frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
trevnels-cs: `rows={1}` is just needed to ensure the textarea fits with the rest of the grid, since the rows prop governs max height. It can accept more than one row and will also wrap as necessary to ensure the entire response stays visible

frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx
trevnels-cs: For the long form text area, yes. The "other" fields now have a char limit too, but the counter isn't shown since it doesn't fit into that UI, and those fields aren't expected to be long responses - the size of those text areas signifies shorter responses that won't risk going close to the limit

cjones-cs (approved):

PR #599, opened 2025-12-17
Stop exhausted teams ctrs

RayofLightz:
Adds admin UI for deleting + recycling a whole deployment. Adds a script to stop exhausted team's instances

### DIFF
--- a/backend/ng/scripts/stop-exhausted-teams.py
+++ b/backend/ng/scripts/stop-exhausted-teams.py
@@ -5,6 +5,7 @@ from CTFd import create_app, CTFdFlask
 from CTFd.plugins.ng.team.models.Team import Team
 from CTFd.plugins.ng.challenge.models.Challenge import Challenge
 from CTFd.plugins.ng.containers.models.ContainerInstance import ContainerInstance
+from CTFd.plugins.ng.containers.models.IndvidualContainer import IndvidualContainer
 from CTFd.plugins.ng.core.utils import utc_now
 
 parser = argparse.ArgumentParser(
@@ -29,3 +30,8 @@ with app.app_context():
         challenges = Challenge.query.filter(Challenge.event == team.event).all()
         for challenge in challenges:
             ContainerInstance.stop_instance_group(challenge.id, team.id)
+
+        for member in team.members:
+            workspace_ctr = IndvidualContainer.get_user_indvidual_container(member.user_id)
+            workspace_ctr.disconnect_from_networks()
+            workspace_ctr.stop()
--- a/backend/ng/scripts/stop-exhausted-teams.py
+++ b/backend/ng/scripts/stop-exhausted-teams.py
@@ -2,7 +2,7 @@
 from CTFd import create_app, CTFdFlask
 from CTFd.plugins.ng.team.models.Team import Team
 from CTFd.plugins.ng.challenge.models.Challenge import Challenge
-#from CTFd.plugins.ng.container.models.ContainerInstance import ContainerInstance
+from CTFd.plugins.ng.containers.models.ContainerInstance import ContainerInstance
 from CTFd.plugins.ng.core.utils import utc_now
 from datetime import timedelta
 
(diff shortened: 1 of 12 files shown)


frontend/src/hooks/container.ts
```diff
@@ -71,3 +71,19 @@ export function recycleContainer(containerId: number) {
...
+  }).then(() => {
+    mutate(`/admin/container/challenge/${challengeId}/team/${teamId}/services`);
+  });
+}
+
+export function deleteDeployment(challengeId: number, teamId: number) {
+  return apiMutation(`/admin/container/challenge/${challengeId}/team/${teamId}/delete`, undefined, {
+    method : 'POST',
+  }).then(() => {
+    mutate(`/admin/container/challenge/${challengeId}/team/${teamId}/services`);
```
trevnels-cs, line 87: Since a deletion is changing the overall container list, make sure to mutate `/admin/container` too

frontend/src/routes/admin/deployments/DeleteDeploymentModal.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Modal
+      title="Delete Deployment"
+      description="Are you sure you want to delete this deployment?"
+      submitVerb="Delete"
+      submitColor={COLOR_NEGATIVE}
+      onSubmit={async () => deleteDeployment(challengeId, teamId)}
+      trigger={(
+        <Button
+          variant="ghost"
```
trevnels-cs: Use `variant="soft"` for buttons in these headers

frontend/src/routes/admin/deployments/RecycleDeploymentModal.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Modal
+      title="Recycle Deployment"
+      description="Are you sure you want to recycle this deployment?"
+      submitVerb="Recycle"
+      submitColor={COLOR_NEGATIVE}
+      onSubmit={async () => recycleDeployment(challengeId, teamId)}
+      trigger={(
+        <Button
+          variant="ghost"
```
trevnels-cs: Use `variant="soft"` for buttons in these headers

frontend/src/routes/admin/deployments/DeleteDeploymentModal.tsx
```diff
@@ -0,0 +1,30 @@
...
+          color={COLOR_NEGATIVE}
+        >
+          <TbTrash />
+          Delete
+        </Button>
+      )}
+    >
+      <Text color="gray">
```
trevnels-cs, line 29: I would recommend using WarningCallout for warnings about data loss to make sure they stand out in the modal. Same for recycle

backend/ng/scripts/stop-exhausted-teams.py
```diff
@@ -0,0 +1,37 @@
...
+
+app: CTFdFlask = create_app()
+
+
+now = utc_now()
+delta = timedelta(days=args.days)
+query_end_date_start = now - delta
+
+with app.app_context():
+    res = Team.query.filter(Team.end_time >= query_end_date_start).all()
```
trevnels-cs: I believe this comparison stops teams with end times in the future instead of teams that have ended in the past, not sure if that's what we want

RayofLightz: Good catch I need a less than in there with the current date as well.

backend/ng/scripts/stop-exhausted-teams.py
zoecough, line 1: Is this going to be automatic or run manually? I don't think it makes a difference usage-wise in this case but we have a script system under `tools/operations` for single operation scripts like this to put everything in one place

zoecough (approved):

backend/ng/scripts/stop-exhausted-teams.py
RayofLightz: I saw that but it looks like its mainly for the emailer or other tools developed for Node? This is going to be manually run right now but eventually we want to look into running it as a cron job or something of that nature.

PR #604, opened 2025-12-19
Leaderboard caching

zoecough:


zoecough:
cache-related testing is excluded for now in the interest of time, working on it on another branch

backend/ng/core/middleware/cache_decorators.py
```diff
@@ -0,0 +1,49 @@
...
+    return decorator
+
+
+def cache_with_args(key_template: str, ttl: int = 300):
+    """Cache function result with dynamic key from arguments"""
+    def decorator(func: Callable) -> Callable:
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            try:
+                cache_key = key_template.format(*args, **kwargs)
```
fpayer: Is there a reason to allow the caller to specify the key_template vs just the key? I'm thinking that the logic would be simplier if the caller only provide the key.

zoecough: It's so that different calls get their own cache entry if the args are different which I thought would make it simpler on the caller's end but I can switch to using the key instead

backend/ng/core/utils/cache_decorators.py
```diff
@@ -0,0 +1,49 @@
+"""Generic Redis caching decorators"""
```
RayofLightz, line 1: My only nit is we probably don't want this in middleware as we probably don't want to decorate a flask end point with these. Maybe move it under util

backend/ng/core/middleware/cache_decorators.py
fpayer: I agree with having different cache entries for different args. My question is specifically regarding the benefit of the caller specifying the key_template vs using the default in the "except" block

fpayer (approved):

RayofLightz (approved):

PR #611, opened 2025-12-29
Add the ability to download the whole log off a container

RayofLightz:
Adds the ability to download the whole log off a container as a file.


fpayer:
What do you think about bumping up the default preview size? Maybe to 1,000?

That might reduce the number of times the full log needs to be downloaded

fpayer (approved):

PR #621, opened 2026-01-07
Check role instead of type for manual award validator

trevnels-cs:
The code here was checking ctfd user type instead of our role field for determining whether a user is admin. Since our role update endpoints and other auth code don't touch this field, we should be using our role field here for consistency with the rest of the system and to ensure role changes in the UI are properly applied to everything. (This is the only spot where user.type is checked like this)

### DIFF
--- a/backend/ng/core/utils/validator.py
+++ b/backend/ng/core/utils/validator.py
@@ -264,6 +264,10 @@ class BaseValidator:
         """
         Validate that a user ID belongs to an admin user
         """
+        # local imports needed here to avoid module dependency cycle
+        from ...permissions.controllers.get_user_roles import get_user_roles
+        from ...permissions.models.enums import RoleEnum
+
         try:
             user_id = int(value)
         except (ValueError, TypeError):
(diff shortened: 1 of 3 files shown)


backend/ng/core/utils/validator.py
```diff
@@ -276,7 +280,7 @@ def validate_admin_id(
             self.errors[field] = f"{friendly_name} with ID {user_id} does not exist"
             return
 
-        if user.type != "admin":
+        if RoleEnum.ADMIN not in get_user_roles(user_id):
             self.errors[field] = "User must be an admin to perform this action"
             return
```
RayofLightz, line 286: Would we want this to be an or incase for whatever reason we want to use ctfd's admin accounts as well

trevnels-cs: Since all of our other code is built with the roles being the sole source of truth for permissions, I'm not sure if having the `or` makes sense. If we did want to use the `type` field to determine admins, it would already take significant refactoring elsewhere in the codebase

fpayer: I'd suggest not doing both cause that wasn't accounted for when we designed the permission system and I'd be concerned about future weird bugs that might occur if we also include the ctfd admin account. 

If we did want to support native ctfd users to have the same admin permissions, it might be safer to come up with a way to make ctfd admin users also ng admin users rather than changing the permission check logic.

fpayer (approved):

PR #623, opened 2026-01-07
Set up embedded grafana for reports

trevnels-cs:
Adds a grafana OSS instance to the environment which is proxied through nginx via auth request and embedded on the admin panel under the Reports tab.

image

Remaining to-do items:
-  [x] Set up grafana config/provisioning to automatically add the db as a datasource and create dashboards based on files in the repo.
- [x] Look at some light caching for the auth request. (Grafana can be very noisy over HTTP and our appserver gets hit with every request. Caching auth responses for a few seconds keyed on the session cookie is probably enough to significantly decrease load on our app server while still handling session expiration well)
- [x] Translate these changes to the production compose/nginx config


conf/nginx/http.conf
```diff
@@ -76,6 +80,23 @@ http {
       proxy_set_header X-Forwarded-Host $server_name;
     }
 
+    location /reports {
```
RayofLightz: We will probably want this as `/admin/reports`

trevnels-cs: /admin/reports is already being used as a frontend route for the corresponding admin panel page, which is why this is just /reports. I could do something like `/admin/reports/grafana`, but it seems a little weird running a full application on that deep of a subpath

fpayer: `/admin/grafana`?

frontend/src/routes/admin/reports/index.tsx
```diff
@@ -0,0 +1,26 @@
+import { Card, Inset } from '@radix-ui/themes';
+import { useTheme } from 'next-themes';
+import { useEffect } from 'react';
+
+export default function AdminReports() {
+  useEffect(() => {
+    localStorage.setItem('grafana.navigation.docked', 'false');
+  }, []);
```
cjones-cs, line 10: I think you should check whether the item is created/set. As this is, wouldn't it reset it to false on each time a person reloads the page? While we should have it false by default, I think we should respect if the user wants to set it to true and persist that after their first visit.

frontend/src/routes/admin/reports/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  const { resolvedTheme } = useTheme();
```
cjones-cs, line 12: Just verifying, but I assume you also checked that the light mode of this looks as good as the dark mode? I only saw the dark mode screenshot

cjones-cs (changes requested):

fpayer (approved):

trevnels-cs:
Configs have been updated - would appreciate extra eyes on them, especially the production side which I'm less familiar with. Generally following Grafana's docs for reverse proxy config: https://grafana.com/tutorials/run-grafana-behind-a-proxy/

For prod setup, `conf/grafana/.env` will need to be updated with correct paths (rerun setup script or copy `conf/grafana/.env.default` to make it)

fpayer (approved):

conf/grafana/grafana.ini
```diff
@@ -1,6 +1,6 @@
 [server]
-root_url = /admin/grafana/
-serve_from_sub_path = true
+# nginx is doing TLS termination, use HTTP internally
+protocol = http
 
 [auth.anonymous]
 enabled = true
```
RayofLightz, line 6: If we want we can setup garfana with a self signed cert and have nginx accept the self signed cert

RayofLightz: We are doing this with the novnc connection

RayofLightz (commented):
Just a note about us being able to not have to terminate ssl.

conf/grafana/grafana.ini
trevnels-cs: I'm not sure if this is needed - Grafana should be using the same certs as the application itself since they'll be under the same domain.

This config just tells Grafana to expect HTTP traffic (which is what nginx will be sending it), even if the `GF_SERVER_ROOT_URL` (its external path) is configured to be https, like it will be in prod.

Issue #581, opened 2025-12-16
Admin - Refactor Event Creation/Editing


cjones-cs:
The event add/edit modal is bloated, convert it to use inline editing. 
Use tabs to separate different information blocks.

cjones-cs:
Details
- name 
 - description (markdown editor with preview tab)
 - image
 
Registration
 - public/private
 - registration open? yes/no
 - registration start/end dates
 - allowed/denied email domain lists
 - allow team management yes/no

Gameplay
 - team size limit
 - start/end dates
 - time limit mins (if not specified, team start/end = event start/end)
 - hints enabled
 - leaderboard visibility
 
Challenges
 - list of challenges (the current sidebar content)

Feedback
 - grid of feedback

PR #625, opened 2026-01-08
Refactor admin event

cjones-cs:
https://github.com/CyberSkyline/ctf-ng/issues/581
image
image


trevnels-cs:
I found a state management issue between the form data and the selected event - submitting an edit to event 1, selecting event 2, then clicking edit again will cause the fields to revert to event 1's data. Trying to save in this state will fail due to a name collision (trying to save event 1's data onto event 2.)

Since RHF only loads defaultValues initially and doesn't respond to the event prop changing, a useEffect that calls RHF's `reset` function with the new event data when the event prop changes should fix it. This is how the feedback form does it: https://github.com/CyberSkyline/ctf-ng/blob/190f832559a639bca56a28a05331b827d03f2bb4/frontend/src/routes/events/OverviewTabs/FeedbackTab/index.tsx#L89-L92

frontend/src/routes/admin/events/SidebarTabs/EventGameplayTab.tsx
```diff
@@ -1,16 +1,207 @@
...
+                      ref={field.ref}
+                      size="3"
+                      {...injected}
+                    />
+                  </Box>
+                )}
+              />
+            )}
+          </FormField>
+          <FormField label="Show Leaderboard" error={errors.show_leaderboard}>
```
trevnels-cs, line 158: This field is labeled Show Leaderboard, but the non-editing version is labeled Leaderboard Visible. I think we should prefer to use one consistent label for both states instead. There's a few other fields like this as well:
- Logo/Image
- Visibility/Public
- Registration Ends/Registration Closes
- Time Limit/Time Limit (minutes)

trevnels-cs (approved):

PR #628, opened 2026-01-09
Add redis lock expire + pid increase

RayofLightz:
Adds an expiration to redis lock + increases the pid limit for workspaces containers


backend/ng/containers/models/ContainerInstance.py
```diff
@@ -20,6 +20,7 @@
 logger = logging.getLogger(__name__)
 
 NET_ERR_REGEX = re.compile("(?:endpoint.*already exists in)|(?:already attached to network)")
+LOCK_EXPIRE = 5*60
```
fpayer: Is this value in seconds? Might be good to leave a comment to indicate the unit.

RayofLightz: I decided to make the variable name more descript that way its obvious its in seconds when passed to the expire arg.

fpayer (approved):

Issue #520, opened 2025-12-08
Ticket Tags removed with click anywhere


thatsatechnique:
Functionally: 
Tags applied to a ticket are removed from the ticket association when a user clicks anywhere in the label (rather than on the 'x').  

Expected behavior - only clicking on the 'x' removes a tag. 

-- 
Visually: 
Tags applied to a ticket include a visible black box.  When selecting tags for a ticket, this appears to be a checkbox (as if you could include several tags at once).  

<img width="600" height="235" alt="Image" src="https://github.com/user-attachments/assets/52695e7f-2214-4162-b87b-68173599f5d1" />

cjones-cs:
The black box is the assigned color for the tag. When creating a tag in the admin panel, you can select the color associated with it, but the default is black.

<img width="947" height="477" alt="Image" src="https://github.com/user-attachments/assets/005d6007-5e5a-4b69-8fb9-470aea56a35b" />

You can actually include several tags at once:

<img width="677" height="81" alt="Image" src="https://github.com/user-attachments/assets/c3f63004-df00-4bbd-bb09-984dfb88e487" />

Being able to click the entire tag to remove was by design, but I can change it to just the X icon on the button.

PR #636, opened 2026-01-12
Make X button on side of badge

cjones-cs:
https://github.com/CyberSkyline/ctf-ng/issues/520

image

Keyboard Focus:
image


RayofLightz (approved):
In the future we might want to yoink this pattern out into its own component. But looks good to merge

PR #637, opened 2026-01-12
Improve start confirmation and include time limit

trevnels-cs:
image

Updates the event start confirmation modal to ensure users understand that they have limited time and what that limited time is. Also clarifies how starting works for teams.

If the timer will be truncated, the truncated time is calculated/displayed and an extra info box explains this to the user.

This calculation is updated every 5 sec to ensure the contents of the modal are accurate, even if the user idles on the page for a prolonged period.

### DIFF
--- a/frontend/src/routes/events/StartModal.tsx
+++ b/frontend/src/routes/events/StartModal.tsx
@@ -2,17 +2,64 @@ import { COLOR_POSITIVE } from '@/constants';
 import { startMyTeam, useEventStatus } from '@/hooks/events';
 import { useRegistration } from '@/hooks/users';
 import type { Event } from '@/types';
-import { Box, Button, Text } from '@radix-ui/themes';
+import {
+  Box,
+  Button,
+  Checkbox,
+  Flex,
+  Strong,
+  Text,
+} from '@radix-ui/themes';
+import { InfoCallout, WarningCallout } from 'components/Callouts';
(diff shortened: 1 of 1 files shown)


frontend/src/routes/events/StartModal.tsx
```diff
@@ -45,11 +92,93 @@ export default function StartModal({ event }: {event: Event}) {
...
+                  {isTeam ? 'your team\'s' : 'your'}
+                  {' '}
+                  allotted time can elapse.
+                </Strong>
+                <br />
+                Your timer will be truncated.
+                {' '}
+                {isTeam ? 'Your team' : 'You'}
+                {' '}
+                will only have until the event&apos;s scheduled end time to work.
```
cjones-cs, line 123: Little nit, but I find it weird to mix escape chars. You use apos here but a few lines up you use \

cjones-cs (approved):

frontend/src/routes/events/StartModal.tsx
trevnels-cs: I'm also not a huge fan of the mixing here, but it's done this way because the different contexts require different types escaping to be formatted correctly. `\'` is used in the JS strings, `&apos;` is used in HTML text. Making them all the same either results in a lint error or `&apos;` ending up in the actual text instead of `'`

PR #642, opened 2026-01-14
Use Name instead of Team for individual leaderboard headers

trevnels-cs:
image


cjones-cs (approved):
Approving this since it works and we should get this out asap. We should follow up on putting isTeam as a virtual field in swr since we use it in so many places

PR #646, opened 2026-01-15
Add self restarts for workstations

RayofLightz:
Does two main things of importance
1. Gives users the ability to restart their own workstation
2. Adds a redis lock for starting/restarting/deleting/recycling workspaces

### DIFF
--- a/backend/ng/containers/models/IndvidualContainer.py
+++ b/backend/ng/containers/models/IndvidualContainer.py
@@ -39,39 +39,42 @@ class IndvidualContainer(db.Model):
     def create_indvidual_container(cls, user_id: int, commit: bool = True):
         redis_client = get_redis_client(3)
 
-        lock = redis_lock.Lock(redis_client, cls.render_lock_key(user_id), expire=LOCK_EXPIRE_SECONDS)
-        if lock.acquire(blocking=False):
-            db_exists = cls.query.filter_by(user=user_id).first()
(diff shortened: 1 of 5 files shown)


RayofLightz:
I will say I am thinking about moving the lock acquisition from the db object creator into the actual run_container function. Just because that is more true to where the lock is actually needed and will reduce the need for constantly calling to redis for a lock. As create_indvidual_container gets called each time the vnc forwarder is hit.

backend/ng/containers/models/IndvidualContainer.py
```diff
@@ -181,55 +203,80 @@ def get_current_challenge(self) -> int | None:
...
 
     def stop(self):
-        client = get_client(self.hostip)
-        try:
-            ctr = client.containers.get(self.dockerid)
-            if ctr.status == DOCKER_RUNNING:
-                ctr.stop(timeout=5)
-        except docker.errors.NotFound:
-            pass
+        redis_client = get_redis_client(3)
```
zoecough, line 246: why is it 3

zoecough (approved):

backend/ng/containers/models/IndvidualContainer.py
RayofLightz: The idea is it should be connecting to its own redis "table"

RayofLightz: Idk if its actually doing in practice though

PR #649, opened 2026-01-22
Feature/practice area

cjones-cs:
Add button on existing practice area page to the PC7 event. Updates disclosure language.
**Link is to the admin portal's Practice Area event. Id = 7**

Fix admin event description direction. Add Save button to top button control for edits

### DIFF
--- a/frontend/src/routes/admin/events/SidebarTabs/EventDetailsTab.tsx
+++ b/frontend/src/routes/admin/events/SidebarTabs/EventDetailsTab.tsx
@@ -196,7 +196,7 @@ export default function EventDetailsTab({ event }: { event: Event }) {
 
           <DataList.Item>
             <DataList.Label>Description</DataList.Label>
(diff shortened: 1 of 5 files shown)


frontend/src/routes/admin/events/SidebarTabs/EventDetailsTab.tsx
```diff
@@ -198,7 +196,7 @@ export default function EventDetailsTab({ event }: { event: Event }) {
 
           <DataList.Item>
             <DataList.Label>Description</DataList.Label>
-            <DataList.Value>
+            <DataList.Value className="flex-col">
```
trevnels-cs: It looks like radix adds a blank ::before element to this that throws off the flex alignment:

Image

Consider adding something like `before:hidden` on the className here to get rid of the extra line

trevnels-cs (approved):

PR #656, opened 2026-01-29
Event practice mode flag

trevnels-cs:
Adds a `practice` boolean for events, which:
- Automatically registers & starts users when they visit a practice event page, with team names as `practice-{uuid}`
- Hides registration/name/leave controls from the event.
- Hides event status badge.
- Hides the leaderboard and feedback tabs from the event overview.
- Hides the event from the Events page listing.

The Practice page's "PC7 Practice Area" link also now points to the first event with the practice flag set. In the future, this page could be updated to handle multiple practice areas as well.

Also includes a fix for admin panel layout to ensure the long auto-generated team names don't overflow action buttons off the screen.

When deploying to staging/production, the events db table will need the practice boolean column added, and existing practice areas will need the practice flag set to true.

### DIFF
--- a/backend/ng/event/controllers/user/join_event_controller.py
+++ b/backend/ng/event/controllers/user/join_event_controller.py
@@ -1,5 +1,7 @@
+from uuid import uuid4
 from CTFd.models import db
 
+from ....core.utils import utc_now
 from ....core import NotFoundError, ValidationError
 from ....team.models.Team import Team
 from ....event.models.Demographic import Demographic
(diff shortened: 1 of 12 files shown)


trevnels-cs:
Another thought - do we want to hide practice event(s) from the main dashboard, have a separate section on the page for it, or just have a link to /practice?

fpayer (approved):

fpayer:
We should check with CISA on that

cjones-cs (approved):

trevnels-cs:
Will go ahead and merge this, I can make the dashboard update in a subsequent PR once we have info about the expected dashboard behavior

PR #659, opened 2026-02-04
Added a build stage to the docker image for faster builds

fpayer:


dockerfiles/ctfd.Dockerfile
```diff
@@ -1,71 +1,91 @@
...
+# Install only runtime dependencies
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends \
+  libffi8 \
+  libssl3 \
   procps \
   net-tools \
   curl \
   supervisor
+RUN rm -rf /var/lib/apt/lists/*
```
RayofLightz: Is there a reason you are rm-ing the apt lists?

RayofLightz (approved):
The only thing I can think of is we might need the build deps in the main thing as well since we install deps when the auto restart is triggered

dockerfiles/ctfd.Dockerfile
fpayer: To reduce image size, but it feels like overkill. I'm gonna remove

fpayer:
Added

PR #660, opened 2026-02-04
Optimize query for support ticket listing

fpayer:


backend/ng/support/models/Ticket.py
```diff
@@ -421,18 +423,18 @@ def find_filtered_tickets(
...
+        # Optimized eager loading with selectinload for collections
+        query = query.options(
+            joinedload(cls.author),
+            joinedload(cls.assigned_user),
+            joinedload(cls.event),
+            joinedload(cls.team),
+            joinedload(cls.challenge),
+            selectinload(cls.messages),
+            selectinload(cls.tags)
+        ).order_by(cls.last_updated.desc())
```
RayofLightz, line 435: Just want to know what the query time difference is? Since a lot of the joinedloads are One to One select in load could also be faster for those.

RayofLightz (approved):
Looks good

backend/ng/support/models/Ticket.py
RayofLightz: Wait no thats Many to one now that I am looking at the schema more.

fpayer: I loaded the existing prod data into my local workspace. The performance is 2.3 seconds before the change vs 400ms after.

PR #663, opened 2026-03-10
Increase challenge description length to 32k

fpayer:


RayofLightz (approved):
Approved assuming unit tests pass

backend/ng/challenge/models/Challenge.py
```diff
@@ -33,9 +33,9 @@ class Challenge(db.Model):
 
     id: Mapped[int] = db.Column(db.Integer, primary_key=True)
     name: Mapped[str] = db.Column(db.String(MAX_CHALLENGE_NAME_LENGTH), nullable=False)
-    description: Mapped[str] = db.Column(db.String(MAX_CHALLENGE_DESCRIPTION_LENGTH), nullable=True)
+    description: Mapped[str] = db.Column(db.Text, nullable=True)
     icon: Mapped[str] = db.Column(db.String(MAX_CHALLENGE_ICON_LENGTH), nullable=True)
-    summary: Mapped[str] = db.Column(db.String(MAX_CHALLENGE_SUMMARY_LENGTH), nullable=True)
+    summary: Mapped[str] = db.Column(db.Text, nullable=True)
```
RayofLightz: Does summary need updated since its length isn't changing? Its probably best to keep it a VARCHAR if possible. This is mostly a micro optimization as it depends on the db but it looks like depending on the DB text will get stored out of row while VARCHAR gets stored in row.

RayofLightz (approved):
Real quick question that probably doesn't matter all that much

backend/ng/challenge/models/Challenge.py
fpayer: I guess not. Will revert that one

PR #664, opened 2026-03-19
Setup improvments

RayofLightz:
Does a few thing for improving setup stuff
* Adds better docs around the setup script
* Adds docs in the README about changing the systemd config for docker
* Makes certain vendor specific things configurable
* Removes unused config.py values


conf/ctfd/.config.json.default
fpayer, line 1: We should do something similar to what we do with the .env vars. Have a default that gets copied over during `pnpm install`. Otherwise any local modifications are going to be tracked by git.

fpayer (changes requested):

conf/ctfd/.config.json.default
RayofLightz: Good point

conf/nginx/http.conf
fpayer, line 1: Why comment out grafana?

RayofLightz: Good catch that was just for dev for me as at the time my box was running a little low on storage so I wasn't running grafana. I will revert it

fpayer (approved):

PR #666, opened 2026-05-07
Improve challenge search, add tag filtering

trevnels-cs:
Update sorting/filtering on the event page:
- Search bar now uses `minisearch` to index across name, summary, description, and tags for fast searching. I've verified with a few hundred challenges with large metadata and observed no slowdown after applying some optimizations.
- Tags as defined in YAML are used to construct the filter UI. Tags can be clicked to toggle them on and off.
  - Only tags that are present within the current event are listed.
  - The filter panel is collapsed by default on page load to put focus on the challenges - users that want to narrow things down can click the "Filter" button to access it.
  - Tag groups are determined by taxonomy, in YAML this looks like `Taxonomy:Tag`. YAML is the source of truth for capitalization and spacing, the platform itself doesn't normalize or apply semantic meaning to tags to give as much flexibility as possible.
  - Tags/taxonomies are sorted deterministically. This is alphabetical in most cases (including the taxonomy as part of the string), though tags are sorted by frequency in the filter panel for easy access to the most common ones. No special cases exist for custom ordering (again, YAML is the source of truth.)
- Tags are also now listed on the challenge admin panel adjacent to the description. These are listed in full without truncation.

image
image

### DIFF
--- a/frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
+++ b/frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
@@ -1,6 +1,6 @@
 import { COLOR_POSITIVE, COLOR_WARNING, type AccentColor } from '@/constants';
 import type { Challenge, MeChallenge } from '@/types';
-import { prettyPrintTag } from '@/util';
+import { prettyPrintTag, tagComparator } from '@/util';
 import {
   Badge,
   Box,
@@ -50,18 +50,7 @@ function ChallengeCard({
   const inProgress = progress && progress.num_attempts_made > 0 && !complete;
 
   const sortedTags = useMemo(
-    () => [ ...challenge.tags ].sort((a, b) => {
-      // sort tags alphabetically. tags without a taxonomy (:) should go last
-      const aHasTaxonomy = a.includes(':');
-      const bHasTaxonomy = b.includes(':');
-      if (aHasTaxonomy && !bHasTaxonomy) {
-        return -1;
-      }
-      if (!aHasTaxonomy && bHasTaxonomy) {
-        return 1;
-      }
-      return a.localeCompare(b);
-    }),
+    () => [ ...challenge.tags ].sort(tagComparator),
     [ challenge.tags ],
   );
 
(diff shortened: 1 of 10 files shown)


frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
```diff
@@ -57,32 +65,53 @@ export default function ChallengeCard({
...
-              <tb.TbCircleCheckFilled />
-            </Text>
+        <Flex direction="column" gap="3" justify="between" className="h-full">
+          <Box>
+            <Flex direction="row" align="center" gap="1">
+              <Heading size="4" color={color}>
+                <ChallengeIcon icon={challenge?.icon} />
+              </Heading>
+              <Heading size="4" color={color}>{challenge.name}</Heading>
+              <Box flexGrow="1" />
```
cjones-cs, line 76: Does this box do anything? It's self-closing, so I'd imagine it creates vertical space, but wouldn't that also be accomplished with padding/margin classes?

frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
```diff
@@ -57,32 +65,53 @@ export default function ChallengeCard({
...
+              {complete && (
+                <Text size="2" color={color}>
+                  <tb.TbCircleCheckFilled />
```
cjones-cs: I think our standard was to import the specific icon instead of using dot notation

cjones-cs (changes requested):

frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
trevnels-cs: That box fills the horizontal space between the left side of the card header (icon, name), and right side (progress icon). The alternative would be to use `justify='between'` on the flexbox, then wrap the left and right groups in additional Flex elements, but I think I avoided that when this was originally written since it would result in messier markup

frontend/src/routes/events/OverviewTabs/ChallengesTab/ChallengeCard.tsx
trevnels-cs: This file does a full import of `tb` to dynamically pick the progress icon. I added an explicit import for the specific icons that are used in the JSX alongside that

cjones-cs (approved):

PR #667, opened 2026-05-11
Prod secrets manager

zoecough:
Added two secrets in AWS secrets manager - one for the app secrets (pulled from .env.production) that load into the app container side-by-side with the non-secret values in .env.production, and one for the ssl certs that get copied into tmpfs on the nginx container to avoid storing them on disk.

Not sure if `development` is the right target here because there are other changes on it that haven't been promoted to `staging` or `production` yet.

### DIFF
--- a/conf/ctfd/init_secrets.py
+++ b/conf/ctfd/init_secrets.py
@@ -105,8 +105,6 @@ class SecretsManager:
 
 
 if __name__ == "__main__":
-    # Print shell export statements for eval in run_production.sh:
-    #   eval "$(python3 init_secrets.py)"
     import shlex
     sm = SecretsManager()
     secrets = sm.load_secrets()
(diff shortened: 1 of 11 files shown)


fpayer:
I think the sensor will still catch the ssl keys because they are writing to a file on disk. We might need to ask for an exemption for those.

conf/ctfd/init_secrets.py
fpayer, line 1: Where does init_secrets get invoked?

zoecough: oops that was still getting invoked in the submodule i'm moving it out

zoecough:
> I think the sensor will still catch the ssl keys because they are writing to a file on disk. We might need to ask for an exemption for those.

That's fine. I'm writing them to tmpfs so the storage is at least ephemeral.

fpayer:
We also need to put the Okta API key in the secrets manager

fpayer (approved):

PR #676, opened 2026-06-17
user creation script

zoecough:
added nonadmin and bulk user creation functionality and renamed create_admin_user scripts to create_user


fpayer:
Just want to confirm that the password won't end up in cmdline history. It looks like the .sh script  does an interactive prompt for the password and the .py script takes it via the commandline. However, since the .py script is invoked via a Docker exec, that should not leave the password in a history file. Correct?

zoecough:
correct

fpayer (approved):

PR #679, opened 2026-07-06
Change color of notif indicator to red

cjones-cs:
Use standard red color (color-negative) for notifi indicator. It also pulsates to make it more noticeable.


RayofLightz (approved):
You are good to merge this, but let me know if you want to just wait for the notification sound.

PR #680, opened 2026-07-06
Add ability to pull vnc image from admin panel

RayofLightz:
Hooks up vnc image pull to the frontend

### DIFF
--- a/frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
+++ b/frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
@@ -11,7 +11,7 @@ type Status = 'pulling' | 'success' | 'fail' | null;
 
 export default function AdminPullVncModal() {
   const [ pullState, setPullState ] = useState<Status>(null);
-  const [ pullError, setPullError ] = useState<string>();
+  const [ pullError, setPullError ] = useState<string | undefined>(undefined);
 
   const handlePull = async () => {
     setPullState('pulling');
@@ -27,9 +27,12 @@ export default function AdminPullVncModal() {
       setPullState('fail');
       setPullError(error);
     });
-  }, []);
 
-  const buttonMessage = pullState ? 'Image Pull Succes' : 'Pull Vnc Image';
+    return () => {
+      socket.off('pull-sccess');
+      socket.off('pull-fail');
+    };
+  }, []);
 
   if (pullError) {
     return (
@@ -52,7 +55,7 @@ export default function AdminPullVncModal() {
           loading={pullState === 'pulling'}
         >
           <TbDownload />
-          {buttonMessage}
+          {pullState ? 'Image Pull Succes' : 'Pull Vnc Image'}
         </Button>
       )}
     >
--- a/frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
+++ b/frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
@@ -29,7 +29,7 @@ export default function AdminPullVncModal() {
     });
 
     return () => {
-      socket.off('pull-sccess');
+      socket.off('pull-success');
       socket.off('pull-fail');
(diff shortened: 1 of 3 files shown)


frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+import Modal from 'components/Modal';
+import socket from '@/socket';
+import { TbDownload } from 'react-icons/tb';
+import { useEffect, useState } from 'react';
+
+type Status = 'pulling' | 'success' | 'fail' | null;
+
+export default function AdminPullVncModal() {
+  const [ pullState, setPullState ] = useState<Status>(null);
+  const [ pullError, setPullError ] = useState<string>();
```
cjones-cs: This is set to string, but then you don't define a default value. I think this is fine since it's inferred, but I think the proper type would end up being `useState<string | undefined>(undefined)`

frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+      setPullState('success');
+    });
+
+    socket.on('pull-fail', ({ error } : { error: string }) => {
+      setPullState('fail');
+      setPullError(error);
+    });
+  }, []);
+
+  const buttonMessage = pullState ? 'Image Pull Succes' : 'Pull Vnc Image';
```
cjones-cs: Since this is only used in one place, consider doing it inline in the trigger instead of making an extra variable

frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+  useEffect(() => {
+    socket.on('pull-success', () => {
+      setPullState('success');
+    });
+
+    socket.on('pull-fail', ({ error } : { error: string }) => {
+      setPullState('fail');
+      setPullError(error);
+    });
+  }, []);
```
cjones-cs: This will only run on initial mount since the dependency array is empty, but still included. The issue is you're initializing event listeners, but not removing them. Consider adding some removal logic at the end: 

> return () => {
    socket.off('pull-success', ...);
    socket.off('pull-fail', ...);
  };

frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+type Status = 'pulling' | 'success' | 'fail' | null;
+
+export default function AdminPullVncModal() {
+  const [ pullState, setPullState ] = useState<Status>(null);
+  const [ pullError, setPullError ] = useState<string>();
+
+  const handlePull = async () => {
+    setPullState('pulling');
+    pullVNCImage();
```
cjones-cs, line 18: Since you're setting a pull state to 'pulling', do you need to tag into the catch/error states from the hook to set it to success/fail states? I guess the sockets do that below, but will that automatically throw a meaningful error into the Modal component's catch?

cjones-cs (changes requested):

frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
RayofLightz: We get success/fail information from the websocket not swr. Since all the prescup stuff if REST/http fetch based we have to background the image pull and send the response back over the websockets. Otherwise we end up with an http timeout

frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
RayofLightz: Good catch on this

cjones-cs (approved):

PR #681, opened 2026-07-07
Add date time stamps for container instances

RayofLightz:
Chris had asked for us to add timestamps for when challenge containers get created. Including a screenshot of the admin panel image as well.
Screenshot 2026-07-07 at 10 21 47 AM

### DIFF
--- a/backend/ng/containers/models/ContainerInstance.py
+++ b/backend/ng/containers/models/ContainerInstance.py
@@ -15,6 +15,7 @@ from .. constants import DOCKER_RUNNING, DOCKER_BRIDGE, DOCKER_MEM_REGEX, DOCKER
 from ...challenge.models.ContainerBlueprint import ContainerBlueprint
 from ...team.models.Team import Team
 from ...core import BusinessLogicError
+from ...core.utils import utc_now
 from ..utils.redis import get_redis_client
 
 logger = logging.getLogger(__name__)
@@ -35,6 +36,7 @@ class SerializedContainerInstance(TypedDict):
     team_id: int
     hostip: str
     dockerid: str
+    created_at: str | None
 
 class ContainerInstance(db.Model):
     __tablename__ = "ng_container_instances"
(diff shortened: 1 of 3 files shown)


frontend/src/routes/admin/deployments/ServiceCard.tsx
```diff
@@ -48,6 +48,12 @@ export default function ServiceCard({ service }: { service: ContainerInstance })
             </Tooltip>
             <br />
             <Text color="gray">{hostip}</Text>
+            {createdAt ? (
+              <>
+                <br />
+                <Text color="gray">{createdAt?.toLocaleString()}</Text>
+              </>
+            ) : null}
```
cjones-cs: `createdAt && (...)` instead of having a null branch

backend/ng/containers/models/ContainerInstance.py
```diff
@@ -43,7 +45,8 @@ class ContainerInstance(db.Model):
     team_id = db.Column(db.Integer, db.ForeignKey("ng_teams.id"), nullable=False)
     hostip = db.Column(db.String(255), nullable=False)
     dockerid: Mapped[str] = db.Column(db.String(255), nullable=False)
-
+    # Allow nullabe for old containers that don't have a start time
+    created_at = db.Column(db.DateTime, nullable=True, default=utc_now)
```
cjones-cs, line 49: What's the harm in updating the older containers to have a time value? I think long term it may be easier to maintain if all of them were enforced to have a time rather than maintain 2 separate styles due to historical data

cjones-cs (changes requested):

backend/ng/containers/models/ContainerInstance.py
RayofLightz: My view on this is that I would rather be true to the fact that the old deployments don't have a `created_at` time. Now its not that big of a deal going forward but when auditing there is a difference between knowing we don't have the info vs having it set to an arbitrary time when this gets deployed.

RayofLightz: I had thought about just setting a value originally, but given that this is specifically being asked about for audit stuff I didn't want to do that.

cjones-cs (approved):

PR #682, opened 2026-07-09
Notif ding when receiving new message

cjones-cs:
Play a ding when a new notif is received.
Don't play a ding when clearing notifs (unread count decreases or stays the same)

Allow user to turn ding on/off in the profile settings.
This sets a local storage value that can be read from anywhere in the site, thus we can tag into the same kind of functionality to set other sounds and alert types.

There's a sample sound button next to the setting so you can know what the notif sound is so you can associate meaning to it. No mystery sounds. I added an aria-label so screen readers would know what the play button does

image
image


frontend/src/hooks/notifications.ts
```diff
@@ -39,3 +40,37 @@ export function markAllNotificationsRead() {
     mutate('/notifications/me/unread-count');
   });
 }
+
+/**
+ * Manages the local storage and state of the notification sound setting
+ */
+const STORAGE_KEY = 'notificationSound';
+
+export function useNotificationSound() {
```
RayofLightz: Nit: `useNotificationSoundEnabled` or something like `useNotificationSoundMuted`

frontend/src/components/NotificationsPopover.tsx
```diff
@@ -16,20 +17,43 @@ import { ErrorCallout } from 'components/Callouts';
...
+  const audioRef = useRef<HTMLAudioElement>(new Audio(ding));
+
+  useEffect(() => {
+    const count = unreadCount?.count;
+
+    if (isNil(count)) return;
+
+    // Play the notification sound if the unread count has increased
+    // Only play the sound if the user localstorage setting for notification sound is enabled
+    if (!isUndefined(previousUnreadCount.current) && count > previousUnreadCount.current && soundEnabled === true) {
```
RayofLightz, line 48: destruct previousUnreadCount to something like previousCount could make this line a little cleaner

frontend/src/routes/profile/index.tsx
```diff
@@ -86,8 +94,36 @@ export default function Profile() {
...
+          <Flex align="center" gap="3">
+            <Switch
+              checked={soundEnabled}
+              onCheckedChange={(checked) => {
+                setSoundEnabled(checked);
+              }}
+              name="Notification Sound"
+              size="3"
+            />
+            <Text as="label">Sound</Text>
```
RayofLightz: Make the text Play Notification Sound. Its a little more descriptive.

frontend/src/routes/profile/index.tsx
```diff
@@ -86,8 +94,36 @@ export default function Profile() {
...
+              aria-label="Play Notification Sound"
+              radius="full"
+              variant="surface"
+              onClick={() => {
+                profileAudioRef.current.currentTime = 0;
+                profileAudioRef.current.play().catch(() => {
+                  // throw away the error intentionally
+                });
+              }}
+            >
```
RayofLightz, line 124: This might look better if its bellow the setting not right next to it imho. Honestly looking at the attached screenshot I think it just needs some spacing away from the setting toggle.

frontend/src/hooks/notifications.ts
```diff
@@ -39,3 +40,37 @@ export function markAllNotificationsRead() {
...
+      window.removeEventListener('notificationSoundChange', handleStorageChange);
+    };
+  }, []);
+
+  const setNotificationSound = (value: boolean) => {
+    localStorage.setItem(STORAGE_KEY, String(value));
+    setEnabled(value);
+
+    // Needed because storage event does not fire in the same tab that made the change
+    window.dispatchEvent(new Event('notificationSoundChange'));
```
RayofLightz, line 75: Do you need this because you are calling setEnabled in this function which is effectively what the event listener is doing

RayofLightz (changes requested):

frontend/src/routes/profile/index.tsx
cjones-cs: I also think some more visual spacing would be good, but I wasn't sure if we wanted to deviate from the gap=2/3 layouts we had in other places. I think having it all inline is still better imho because if we ever added additional settings for notifications, like enable/disable emails or other sounds, then we would just list them like this. So one row makes the most sense for me

frontend/src/routes/profile/index.tsx
cjones-cs: We shouldn't use that specific text because it's the aria-label for the play button and denotes an action button. I've changed it to "Notification Sound" instead.

frontend/src/components/NotificationsPopover.tsx
cjones-cs: While I could do that in this limited block scope without too much worry, it's generally a bad idea to destructure refs, so I'd prefer to keep it like this since it doesn't cause too much issue since it's only one extra word.

frontend/src/hooks/notifications.ts
cjones-cs: Yes, because it won't update the value in the current tab without it. It will update in other tabs though

RayofLightz: Okay. Do you know what its not handled by the set on line 69?

RayofLightz (approved):

PR #685, opened 2026-07-13
Accordion component on past dashboard events

cjones-cs:


trevnels-cs (approved):
Looks good!

One potential styling suggestion, keeping the panel background on the trigger only and letting the page background show through the content (removing `bg-[var(--color-panel-solid)]` from the root component styles) helps improve contrast on the cards inside and makes the accordion triggers more distinct from their content:

Before:
Image

After:
Image

trevnels-cs (approved):

PR #688, opened 2026-07-14
Check types in CI alongside frontend lint

trevnels-cs:
- Adds a `type-check` script which runs as part of the frontend-linter action to break the build on TS errors. (The script does a `cd ..` first so paths emitted by tsc are relative to the repo root, and GitHub can link diagnostics from the action to files in the repo.) 
- Resolves existing TS errors (including a bump to TS6 which revealed a few more issues TS5 didn't catch.)
  - The majority of frontend changes here are fixes to type signatures with no runtime effects, the couple of changes which do affect the application itself are bugs that tsc correctly identified
- Removed the overarching `typescript-eslint` package since we're only using the namespaced subpackages.

### DIFF
--- a/.github/workflows/frontend-linter.yml
+++ b/.github/workflows/frontend-linter.yml
@@ -35,7 +35,14 @@ jobs:
         cd frontend
         pnpm install --ignore-scripts
 
-    - name: Test
+    - name: Type Check
+      run: |
+        cd frontend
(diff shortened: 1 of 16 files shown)


zoecough (approved):
LGTM, might be a good idea to have Caitlin look over your linting changes since I'm less knowledgeable on ts and frontend

PR #689, opened 2026-07-14
Make date parsing explicit at the component level

### LINKED ISSUE #203: Fix grid date filters
Currently date filters don't work for a case of == because the unix dates are datetimes. Create a custom filter so that we only do checks on the date part of the value instead of the whole datetime.

ag-grid uses js equality checks underneath by default.

Filtering on before/after/between currently work

### LINKED ISSUE #687: Date parsing during frontend deserialization applies to user provided strings
If I pass a valid ISO timestamp in a user-provided text field, it would be parsed as a Date when being reflected back to the frontend. Frontend code wouldn't expect this for string fields, so this may result in weird rendering as a result of date -> string coercion.

Image 

Parsing dates at their usage sites instead of implicitly is the recommended way to proceed here.

trevnels-cs:
Fixes #687.

Do date parsing at the component layer which is safer than the implicit parsing done before. Also automatically resolves #203 by letting the grid parse string dates and manage its filters based on those.

Introduces a `formatDate` helper to make date formatting within components cleaner.

### DIFF
--- a/frontend/src/routes/events/OverviewTabs/ChallengesTab/NotAvailable.tsx
+++ b/frontend/src/routes/events/OverviewTabs/ChallengesTab/NotAvailable.tsx
@@ -27,9 +27,7 @@ export default function NotAvailable() {
             <Text size="3" color="gray">
               The event will start on
               {' '}
-              {formatDate(event.start_time, { year : 'numeric', month : 'numeric', day : 'numeric' })}
-              {' at '}
-              {formatDate(event.start_time, { hour : 'numeric', minute : 'numeric' })}
+              {formatDate(event.start_time, { dateStyle : 'long', timeStyle : 'short' })}
               .
             </Text>
           )}
--- a/frontend/src/routes/events/OverviewTabs/ChallengesTab/NotAvailable.tsx
+++ b/frontend/src/routes/events/OverviewTabs/ChallengesTab/NotAvailable.tsx
@@ -1,4 +1,5 @@
 import { useEvent, useEventStatus } from '@/hooks/events';
+import { formatDate } from '@/util';
 import {
   Box,
   Container,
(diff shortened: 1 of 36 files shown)


frontend/src/routes/events/OverviewTabs/ChallengesTab/NotAvailable.tsx
```diff
@@ -26,9 +27,9 @@ export default function NotAvailable() {
             <Text size="3" color="gray">
               The event will start on
               {' '}
-              {event.start_time.toLocaleDateString()}
+              {formatDate(event.start_time, { year : 'numeric', month : 'numeric', day : 'numeric' })}
               {' at '}
-              {event.start_time.toLocaleTimeString()}
+              {formatDate(event.start_time, { hour : 'numeric', minute : 'numeric' })}
```
cjones-cs: Why is this one different than the other displayed dates? Wouldn't it be better to use the same display format for all of them?

frontend/src/types.ts
```diff
@@ -26,7 +26,7 @@ export interface Sponsor {
 export interface UploadedFile {
   filename: string,
   folder: string,
-  last_modified?: Date,
+  last_modified?: string,
```
cjones-cs, line 29: wouldn't this also be string | null like the types above?

trevnels-cs: This one is still `?: string` since some of the S3-related endpoints do not include last_modified on the object at all

frontend/src/routes/events/OverviewTabs/ChallengesTab/NotAvailable.tsx
trevnels-cs: Collapsed this down into one `formatDate` call. Using long format for the date so that it reads more naturally as a prose message:

> The event will start on January 1, 2026 at 12:00 PM.

cjones-cs (approved):

PR #690, opened 2026-07-15
Upgrade vendored novnc to 1.7.0

### LINKED ISSUE #652: Special Keys in NoVNC hidden by style
On the NoVNC panel, the special keys width is too small, resulting in the following odd interface. 

Image

Playing with the CSS, the following fixed this issue for me in `base.css` from the NoVNC css files.

```css
/* Control bar buttons */
#noVNC_control_bar .noVNC_button {
    /*min-width: unset;*/
   min-width: -webkit-fill-available;
...
```

trevnels-cs:
NoVNC 1.7.0 resolves #652, as well as includes some performance improvements.


RayofLightz (approved):
Just want to confirm these files are just pulled from the novnc upstream repo correct?

trevnels-cs:
Yep, these are from their 1.7.0 release artifact.

PR #693, opened 2026-07-16
Progress bars for image pull status

trevnels-cs:
- Adds `pull-progress` messages to the image pull task, which periodically send an aggregate progress value to the client.
- `ImagePullStatus` component & hook, with global socket listeners writing to SWR to keep pull events flowing when on other pages.
- `ChallengeBlueprintTab` and `AdminProvisionerCard` updated to use the new pull status component, including progress bars to display the progress value.

Screenshot From 2026-07-16 15-38-42

### DIFF
--- a/frontend/src/components/ImagePullStatus.tsx
+++ b/frontend/src/components/ImagePullStatus.tsx
@@ -1,0 +1,59 @@
+import { COLOR_NEGATIVE, COLOR_POSITIVE } from '@/constants';
+import { useImagePullStatus } from '@/hooks/container';
+import {
+  Code,
+  Flex,
+  HoverCard,
+  Progress,
+  Skeleton,
+  Text,
+} from '@radix-ui/themes';
+import { TbCheck, TbInfoCircle } from 'react-icons/tb';
+
+export default function ImagePullStatus({ id } : { id: string | number }) {
+  const { data : pull } = useImagePullStatus(id);
+
+  if (pull?.status === 'pulling') {
+    return (
+      <Flex direction="row" align="center" gap="2">
+        {!!pull.percent && (
+          <Text size="2" color="gray">Pulling</Text>
+        )}
+        <Skeleton loading={!pull.percent}>
+          <Progress value={pull.percent} className="w-24" />
+        </Skeleton>
+      </Flex>
+    );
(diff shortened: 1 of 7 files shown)


frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
```diff
@@ -1,67 +1,35 @@
...
+  const { data : pull } = useImagePullStatus('VNC');
 
-  if (pullError) {
-    return (
-      <ErrorCallout className="ml-4">
-        {pullError}
-      </ErrorCallout>
-    );
-  }
+  const handlePull = () => pullVNCImage();
```
RayofLightz: Any reason for not having the anon function inside the onSubmit since its only used there?

frontend/src/components/ImagePullStatus.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  if (pull?.status === 'fail') {
+    return (
+      <Flex direction="row" align="center" gap="1">
+        <Text color={COLOR_NEGATIVE}>Error</Text>
+        <HoverCard.Root>
+          <HoverCard.Trigger>
+            <button type="button">
+              <Text color={COLOR_NEGATIVE}>
+                <TbInfoCircle aria-label="More info" />
```
RayofLightz: Consider using the Cancel icon since that is what we used before

RayofLightz (changes requested):
A few small nits, but otherwise this looks good

frontend/src/components/ImagePullStatus.tsx
trevnels-cs: I will look at making that change, the main reason for (i) was to signify that there's additional content on hover, which text and a different icon doesn't convey as well. Some underline styling on the text would probably convey the same well enough

frontend/src/routes/admin/dashboard/AdminPullVncModal.tsx
trevnels-cs: Good point, there used to be more in that function but it's no longer necessary now.

RayofLightz (approved):

PR #695, opened 2026-07-21
Allow expo user creation and password changes via admin panel

trevnels-cs:
New create expo account button on the Users page opens this modal:
Screenshot From 2026-07-21 10-07-29

Password changes are available for expo users through the existing edit user form. Leaving the fields blank means that the password will not be changed. These fields are disabled for SSO/Okta users to ensure those accounts always go through OAuth and 2FA for authentication.
Screenshot From 2026-07-21 10-08-03

SSO grid column shows whether users are authenticated via SSO (none are on my dev instance)
Screenshot From 2026-07-21 10-08-21

### DIFF
--- a/backend/ng/user/tests/test_user_api.py
+++ b/backend/ng/user/tests/test_user_api.py
@@ -474,3 +474,57 @@ def test_non_admin_create_user(logged_in_client):
     )
 
     assert response.status_code == 403
+
+
+def test_user_put_password(admin_client, user, db_session):
+    """
+    Test that an admin can change an expo user's password
+    """
+    from CTFd.utils.crypto import verify_password
+
+    response = admin_client.put(f"/ng/admin/users/{user.id}", json={"password": "newpassword123"})
+
+    assert response.status_code == 200
+    assert response.get_json()["success"] is True
+
+    db_session.refresh(user)
+    assert verify_password("newpassword123", user.password)
+
+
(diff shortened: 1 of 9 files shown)


trevnels-cs:
Terminology switched from guest to expo

cjones-cs (approved):

backend/ng/user/tests/test_user_api.py
```diff
@@ -388,3 +388,143 @@ def test_update_user_sponsor_no_data(client_factory,user_factory):
...
+    """
+    from CTFd.utils.crypto import verify_password
+
+    response = admin_client.put(f"/ng/admin/users/{user.id}", json={"password": "newpassword123"})
+
+    assert response.status_code == 200
+    assert response.get_json()["success"] is True
+
+    db_session.refresh(user)
+    assert verify_password("newpassword123", user.password)
```
RayofLightz, line 491: Is this a controller that already existed and just had no TCC?

trevnels-cs: Yeah, the PUT endpoint had already existed, but the validator previously did not accept a password. This test makes sure the password clears the updated validator and the correct hash gets written by ctfd

RayofLightz: Okay everything looks good then

PR #696, opened 2026-07-21
Update deps for frontend & exec-service

trevnels-cs:
Bump pnpm dependencies to latest minor versions across frontend and exec-service. `dockerode` is bumped by a major version  to bring in security patches for its transitive deps (no breaking changes included with that, just a change to compatible node versions.) `pnpm audit` reports 0 vulnerabilities after these updates, down from 95 (though many of those are from build-time or internal packages.)

Also brought the exec-service packageManager field into alignment with the frontend's to avoid pnpm errors due to conflicting versions.

I've been click testing on this branch for a few days with no visible breakage or behavior changes.

Going to evaluate new major versions of things like ag-grid and xterm separately, since those are a bit riskier.


RayofLightz (approved):
We will have to remember to redeploy the exec service. Might be worth having @zoecough or myself make a playbook for that

PR #700, opened 2026-07-31
Automated certificate generation

trevnels-cs:
- On-demand certificate generation using Typst python bindings
  - See https://typst.app/blog/2025/automated-generation for the general approach
  - Produces Section 508 compliant PDF via PDF/UA-1 spec
  - Document structure and assets are heavily memoized, average render time is ~3ms after initial compiler load
  - 1 request/sec rate limit for certificates. Users shouldn't hit this interactively, but will prevent spam.
  - Fonts are not included in the repo due to licensing - a fallback serif font will be used by default until fonts are dropped into `backend/ng/certificates/templates/fonts` on your local machine.
- Certificates can be configured per-event via a dropdown of available .typ files under the compiler root
  - For now, there is just the base template and an evergreen CISA-branded variant.
  - Path traversal is blocked both at the python and typst level - everything must be relative to `backend/ng/certificates/templates`, the compiler root.
- Base template can handle both per-event and per-challenge certificates depending on the data passed.
  - For non-practice events, certificates become available for teams when the event time window closes (or if the event has no stop time, it's based on the team's.)
    - Download link is in the event page header
  - For practice events, certificates are available for each challenge upon completion.
    - Download link appears below the challenge questions, where users will see it after making their last submission.

Deployment notes:
- `certificate_template` column will need to be added to the Event table.
- Fonts will need to be manually dropped into their directory. I can send files when needed.


backend/ng/certificate/controllers/list_certificates.py
```diff
@@ -0,0 +1,24 @@
...
+        list[str]: Relative paths of available .typ template files
+    """
+    templates = []
+    for root, _dirs, files in os.walk(CERTIFICATES_DIR):
+        for filename in files:
+            if filename.endswith(".typ") and not filename.startswith("_"):
+                relative_path = os.path.relpath(os.path.join(root, filename), CERTIFICATES_DIR)
+                templates.append(relative_path.replace(os.sep, "/"))
+
+    return sorted(templates)
```
RayofLightz, line 24: Any particular reason for sorting this?

backend/ng/event/models/Event.py
```diff
@@ -260,6 +275,7 @@ def create_event(
             blocked_domains=blocked_domains,
             show_leaderboard=show_leaderboard,
             practice=practice,
+            certificate_file=certificate_file,
```
RayofLightz: Nit should be certificate_template

backend/ng/event/routes/user_routes.py
```diff
@@ -576,3 +579,51 @@ class EventChallengeRecycleContainers(Resource):
...
+        responses={
+            200: "Success - Returns the certificate PDF",
+            400: "Certificate is not yet available",
+            403: "Forbidden - Authentication required",
+            404: "Event has no certificate, or does not issue this certificate type",
+        },
+    )
+    def get(self, event: Event, team: Team, current_user: User, **kwargs):
+        pdf = CertificateService.render_certificate(current_user, team, event, tz=request.args.get("tz"))
+        return send_file(io.BytesIO(pdf), mimetype="application/pdf", as_attachment=True, download_name=f"Certificate - {event.name}.pdf")
```
RayofLightz: We might want to set a few cache-control headers on these responses.
Namely private and decently high max age. I would also consider a higher rate limit than what is currently on it.

frontend/src/components/CertificateCallout.tsx
```diff
@@ -0,0 +1,71 @@
...
+    href = `/ng/events/${eventId}/certificate`;
+  }
+
+  if (!href) {
+    return null;
+  }
+
+  href += `?tz=${encodeURIComponent(Intl.DateTimeFormat().resolvedOptions().timeZone)}`;
+
+  return (
```
RayofLightz, line 54: If they haven't completed the event and thus can't download the cert we should probably not render the callout.

RayofLightz (changes requested):

frontend/src/components/CertificateCallout.tsx
trevnels-cs: There is a check further above in this file that will return `null` for this component in cases that the certificate is not yet available to a user

backend/ng/event/routes/user_routes.py
trevnels-cs: My main consideration with the rate limit was to make sure that a user that accidentally closes out of a certificate and wants to redownload it again right away won't face an error, though with more aggressive caching this may pose less of an issue. There's also the case of someone wanting to download a bunch of practice certificates at once after completing a few different challenges, so the limit should probably leave headroom for that.

Thoughts on something in the 5-10 sec range for the ratelimit? Keep in mind that PDF generation is extremely cheap here, the few DB queries to load the user's data are more time-consuming than producing the file itself.

backend/ng/certificate/controllers/list_certificates.py
trevnels-cs: The order in which data comes back from walk is filesystem-dependent, so the sort normalizes it to always be consistent when listing files for the admin dashboard

backend/ng/event/routes/user_routes.py
RayofLightz: I think cache control with a 5 second rate limit should be good

frontend/src/components/CertificateCallout.tsx
RayofLightz: Oh I see it my bad

RayofLightz (approved):
Looks good

PR #702, opened 2026-08-06
Feature/admin override badge

cjones-cs:
Add ability to make any existing hook pass an additional value `isAdminOverride`. The value will be consumed in the frontend component in the same way data or isLoading is. Default value of isAdminOverride is false.

In order to use isAdminOverride, simply change the useSWR to useAdminOverrideSWR. This will provide the value (false) and you can programatically set it in your backend function on a per-case basis.

It is intended to work with the generic frontend component `AdminOverride` to display either a radix badge or a callout (themed similarly to the WarningCallout (orange).

Admin Badge
image

Admin Callout
image

Normal users will continue not seeing the leaderboard based on the event.show_leaderboard value:
image

If an admin is capable of seeing the leaderboard because event.show_leaderboard = True, then you won't see the badge. This is due to the backend logic in the leaderboard function and not a generic condition.
image


frontend/src/routes/events/OverviewTabs/LeaderboardTab/index.tsx
```diff
@@ -58,6 +59,8 @@ export default function LeaderboardTab() {
   return (
     <Container size="4">
       <Flex direction="column" gap="3">
+        {leaderboardAdmin
+          && <AdminOverride type="badge">ADMIN VIEW</AdminOverride>}
```
RayofLightz: Small thing you can have ADMIN VIEW be the default content and then you just need to have it be `<AdminOverride />`

RayofLightz (approved):

